PR #6, opened 2018-04-18
- move to onnx-1.1.1

guschmue:
- fix split_op, splitv_op
- add mobilenet_v2


guschmue:
moved to onnx-1.1.1 because of:
https://github.com/onnx/onnx/issues/762

Issue #54, opened 2018-06-26
pip install onnx==1.2 failed, No matching distribution found for onnx==1.2


pengwa:
Could not find a version that satisfies the requirement onnx==1.2 (from versions: 0.1, 0.2, 0.2.1, 1.0.0, 1.0.1, 1.1.0, 1.1.1, 1.1.2, 1.2.1, 1.2.2)
No matching distribution found for onnx==1.2

I think we might need update the README for a version, which exists in https://pypi.org/project/onnx/#history? :)

guschmue:
yes, my bad. I'll change the readme. setup.py will actually install it anyway.

PR #57, opened 2018-06-26
fix onnx version in readme

guschmue:


guschmue:
Fix for https://github.com/onnx/tensorflow-onnx/issues/54

PR #63, opened 2018-07-05
support, test for fusedbatchnorm

guschmue:


guschmue:
still missing: opset1, 6 support and graph rewrite in case there is a training path (which is pretty common)

PR #71, opened 2018-07-11
document the python api

guschmue:


guschmue:
@pengwa ... you might want to take look.

pengwa:
@guschmue seems I don't permission to merge, so you would have to do that. :)

guschmue:
cool, thanks!

Issue #79, opened 2018-07-18
tensorflow op Fill is not supported


pengwa:
When I tried to convert dcgan inference model, TensorFlow Fill operator can not be converted well. 

```
node {
  name: "generator/ones_1/shape_as_tensor"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_INT32
        tensor_shape {
          dim {
            size: 4
          }
        }
        tensor_content: "@\000\000\000\016\000\000\000\016\000\000\000\n\000\000\000"
      }
    }
  }
}
node {
  name: "generator/ones_1/Const"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_FLOAT
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_FLOAT
        tensor_shape {
        }
        float_val: 1.0
      }
    }
  }
}
node {
  name: "generator/ones_1"
  op: "Fill"
  input: "generator/ones_1/shape_as_tensor"
  input: "generator/ones_1/Const"
  attr {
    key: "T"
    value {
      type: DT_FLOAT
    }
  }
  attr {
    key: "index_type"
    value {
      type: DT_INT32
    }
  }
}
```

While on the other hand, the corresponding operator in ONNX ConstantFill is in experimental stage. 

Since the mapping is pretty straight forward, I will create a pull request for fixing this.

PR #80, opened 2018-07-18
add constant fill support

pengwa:
fix https://github.com/onnx/tensorflow-onnx/issues/79


guschmue:
The ConstantFill is experimental and as far I know only works for caffe2.
Can we maybe use the Constant op or an Initializer until onnx has some official Fill op ? We badly want this to work on winml :)

pengwa:
@guschmue  yes, sure, I feel okay for not merging this at this moment. :) Never mind, I also feel *Fill operators are not that necessary especially for inference scenarios. But for supporting training in the future, that might be different. :)  I'll close this for now.

Issue #81, opened 2018-07-19
Conv2DBackpropInput conversion failure: new_shape[i] = shape[perm[i]] IndexError: list index out of range


pengwa:
```

-------------------- start handling dcgan_2.pb ----------------------------
change working directory to /home/pengwang/community/tensorflow
------ summarize the frozen graph, to get the inputs and outputs name
bazel-bin/tensorflow/tools/graph_transforms/summarize_graph --in_graph=/tmp/frozen/dcgan_2.pb
Found 2 possible inputs: (name=y, type=float(1), shape=[64,10]) (name=z, type=float(1), shape=[?,100])
No variables spotted.
Found 1 possible outputs: (name=generator/Sigmoid, op=Sigmoid)
Found 7080115 (7.08M) const parameters, 0 (0) variable parameters, and 6 control_edges
Op types used: 50 Const, 23 Identity, 12 Sub, 8 Mul, 8 Reshape, 4 ConcatV2, 3 FusedBatchNorm, 3 Relu, 2 Add, 2 BiasAdd, 2 Conv2DBackpropInput, 2 Fill, 2 MatMul, 2 Placeholder, 1 Sigmoid
To use with tensorflow/tools/benchmark:benchmark_model try these arguments:
bazel run tensorflow/tools/benchmark:benchmark_model -- --graph=/tmp/frozen/dcgan_2.pb --show_flops --input_layer=y,z --input_layer_type=float,float --input_layer_shape=64,10:-1,100 --output_layer=generator/Sigmoid
------ update the inputs and outputs name to format like input_name:index
python3 /home/pengwang/community/learning/onnx/update_name_with_index.py y,z
updated input names is y:0,z:1, output names is generator/Sigmoid:0
------ start convertion, tensorflow usage require caller program must not in tensorflow root folder, so switch to current user directory with cd
using tensorflow=1.9.0-rc0, onnx=1.2.1
2018-07-19 10:59:39.363171: I tensorflow/tools/graph_transforms/transform_graph.cc:318] Applying fold_batch_norms
2018-07-19 10:59:39.390275: I tensorflow/tools/graph_transforms/transform_graph.cc:318] Applying fold_old_batch_norms
2018-07-19 10:59:39.502133: I tensorflow/core/platform/cpu_feature_guard.cc:141] Your CPU supports instructions that this TensorFlow binary was not compiled to use: SSE4.1 SSE4.2 AVX AVX2 FMA
[...]

PR #82, opened 2018-07-19
fix Conv2DBackpropInput conversion failure

pengwa:
to fix https://github.com/onnx/tensorflow-onnx/issues/81


guschmue:
cool, thanks, I must have screwed this up recently. Need to check why the unit tests did not catch it.

PR #88, opened 2018-07-26
fix the graph output node dtype issue

pengwa:
send for an early review, need solve the int32 issue before merge.

### DIFF
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -390,13 +390,11 @@ class Graph(object):
         # create output_tensor_values
         output_tensor_values = []
         for name in output_names:
-            op = self.get_node_by_name(name)
-            if op:
-                dtype = op.dtype
-                if not dtype:
-                    continue
-                v = helper.make_tensor_value_info(name, dtype, self.get_shape(name))
-                output_tensor_values.append(v)
+            dtype = self.get_dtype(name);
+            if not dtype:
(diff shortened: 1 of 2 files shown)


pengwa:
I check some handling, one of them is trying to insert a cast op, so do you think we should do that here? or there is better way? 

cast_op = ctx.insert_new_node_on_input(node, "Cast", name)
        cast_op.set_attr("to", onnx_pb.TensorProto.INT64)

guschmue:
yes, that is one of those. Don't know any other way of doing it on opset 7 (I know, its not helping perf).
Some is fixed in opset 8 (like reshape) but not all of it and opset 8 we don't want to depend on until winml can use it (which might be some time).

pengwa:
Thanks @guschmue , I tried another approach. since Graph (e.g. ctx)'s _dtypes are prepared  before func() are executed, while I think limiting the int32->int64 is operator level stuff/workaround. So I update the global ctx's _dtype when TopK operator is being handled, then when make_graph is called, ouput node is generated with the updated topk:1 dtype. 

Do you have any concerns since I am modifying the dtypes of topk:1 in the middle of graph handling?

guschmue:
The main worry I'd have is that the consumer might not expect int64. But in this specific case its always a index which should be ok.

pengwa:
@guschmue yes, please help merge it. Thanks!! 

there will be something that can be improved, I will follow up with this.

PR #108, opened 2018-08-22
add transpose optimizer, and integrate it after tf graph conversion

pengwa:


tf2onnx/convert.py
```diff
@@ -94,10 +95,14 @@ def main():
         "converted from {}".format(args.input), args.inputs, args.outputs,
         optimize=not args.continue_on_error)
 
+    onnx_graph = OnnxGraph(model_proto.graph)
+    optimizer = TransposeOptimizer(onnx_graph)
```
guschmue: can you make this optional via command line, don't think this tested well enough

tf2onnx/optimizer/onnx_graph.py
```diff
@@ -0,0 +1,379 @@
+from __future__ import division
+from __future__ import print_function
+
```
guschmue: why not just use graph.py ... most of the code in this file looks the same.

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -0,0 +1,382 @@
...
+            # make sure keepdims is 1, then we can do the swap, otherwise, plese don't, because
+            # once keepdims is not set, original dims are lost, so transpose back won't work well.
+            # by default, if keepdims is not specified, it is 1
+            if axes == [1, 2] and ((keepdims and keepdims.i == 1) or (not keepdims)):
+                p.set_attr("axes", [2, 3])
+            else:
+                return
+            switch_transpose_and_node(g, p, trans)
+            return True
+
```
guschmue: how about ops like stridedslice ? I think you'd need to make a pass through all ops if any of the attributes is impacted.

tf2onnx/tfonnx.py
```diff
@@ -1369,7 +1369,7 @@ def tensorflow_onnx_mapping(g, continue_on_error, custom_op_handlers):
 def tf_optimize(sess, inputs, outputs, graph_def):
     """Optimize tensorflow graph for inference."""
     transforms = [
-        # "fold_constants(ignore_errors=true)",
```
guschmue: fold_constants on some graphs runs into a tensorflow issue ... that is the reason why it is off.

tests/run_pretrained_models.py
```diff
@@ -206,11 +206,10 @@ def run_onnxmsrt(self, name, onnx_graph, inputs):
             self.onnx_runtime = time.time() - start
         return results
 
-    def run_onnxmsrtnext(self, name, onnx_graph, inputs):
+    def run_onnxmsrtnext(self, name, model_proto, inputs):
         """Run test against msrt-next backend."""
```
guschmue: when running this with your changes:
mobilenet_v1_75_192
        downloaded /tmp/pre-trained\mobilenet_v1_0.75_192_frozen\mobilenet_v1_0.75_192/frozen_graph.pb
2018-08-22 14:28:57.717831: I T:\src\github\tensorflow\tensorflow\tools\graph_transforms\transform_graph.cc:318] Applying fold_constants
2018-08-22 14:28:57.870910: I T:\src\github\tensorflow\tensorflow\tools\graph_transforms\transform_graph.cc:318] Applying fold_batch_norms
2018-08-22 14:28:58.012064: I T:\src\github\tensorflow\tensorflow\tools\graph_transforms\transform_graph.cc:318] Applying fold_old_batch_norms
        tensorflow OK
before optimization: ops statistics: Counter({'Transpose': 58, 'Conv': 28, 'Add': 28, 'Max': 27, 'Min': 27, 'Mul': 13, 'Identity': 1, 'Softmax': 1, 'Reshape': 1, 'AveragePool': 1, 'Squeeze': 1})
        to_onnx FAIL name 'TensorShapeProto' is not defined
        run_onnx FAIL 'NoneType' object has no attribute 'SerializeToString'

mobilenet_v1_100_224
        downloaded /tmp/pre-trained\mobilenet_v1_1.0_224_frozen\mobilenet_v1_1.0_224/frozen_graph.pb
2018-08-22 14:28:15.857182: I T:\src\github\tensorflow\tensorflow\tools\graph_transforms\transform_graph.cc:318] Applying fold_constants
2018-08-22 14:28:16.066649: I T:\src\github\tensorflow\tensorflow\tools\graph_transforms\transform_graph.cc:318] Applying fold_batch_norms
2018-08-22 14:28:16.295635: I T:\src\github\tensorflow\tensorflow\tools\graph_transforms\transform_graph.cc:318] Applying fold_old_batch_norms
        tensorflow OK
before optimization: ops statistics: Counter({'Transpose': 58, 'Conv': 28, 'Add': 28, 'Max': 27, 'Min': 27, 'Mul': 13, 'Identity': 1, 'Softmax': 1, 'Reshape': 1, 'AveragePool': 1, 'Squeeze': 1})
        to_onnx FAIL name 'TensorShapeProto' is not defined
        run_onnx FAIL 'NoneType' object has no attribute 'SerializeToString'

tf2onnx/tfonnx.py
pengwa: maybe I can enable it once use decides to enable the optimization? because current optimization depends on this, otherwise, lots of transpose remains. but eventually I will have a try when it is disabled.

tests/run_pretrained_models.py
pengwa: oh, my bad, it's a last minute change. I'll fix it. sorry.

tf2onnx/convert.py
pengwa: okay, let me do that.

tf2onnx/optimizer/onnx_graph.py
pengwa: originally I am trying to re-use graph.py, while I found more onnx graph specific things are needed, so gradually changed a lot, so I would prefer to make it a new simpler classes. I'm actually trying to make the class even more simpler.

tf2onnx/optimizer/transpose_optimizer.py
pengwa: for ops that are not listed explicitly in the if branch, the transposes before it will remain there, so the correctness of the graph can be guaranteed. 

I'll add stridedslice support later.

tf2onnx/optimizer/onnx_graph.py
pengwa: OnnxGraph almost have most of its member function rewritten, to 1) update output nodes/output node number, 2). get those unused node/initilizers removed automatically once set_nodes() is called.

guschmue: We want exactly one internal onnx graph / node representation. If there are changes you need in graph.py I'm sure that can be done.

tf2onnx/tfonnx.py
guschmue: I think that would be ok. 
The error message you'd get is 'expects to be colocated with unknown node ' and there are some reports of this (I thought I had seen some detailed tensorflow issue on this but can't find it anymore).
We do have models that runs into this but I have not looked closer ... maybe time to understand what the root cause is.

tf2onnx/optimizer/onnx_graph.py
pengwa: let me try

pengwa:
thanks @guschmue !!

PR #113, opened 2018-08-26
fix unit tests

guschmue:


pengwa:
sorry for the regression...

guschmue:
no worry, ci should have prevented it but we disabled the backend tests in travis because caffe2 crashes in the 14.04 travis image. Need look at it.

PR #115, opened 2018-08-28
_convert_shapenode_to_int64 did not use passe-in-ed input_number

pengwa:


tf2onnx/graph.py
```diff
@@ -343,9 +350,9 @@ def get_dtype(self, name):
         """Get dtype for node."""
         return self._dtypes.get(name)
 
-    def set_dtype(self, name, val):
-        """Set dtype for node."""
-        self._dtypes[name] = val
+    def override_dtype(self, name, val):
```
guschmue: not sure what is wrong with keeping the final types in one dict

pengwa: @guschmue it's actually to fix your old comment :)  https://github.com/onnx/tensorflow-onnx/pull/88#issuecomment-408092157

guschmue: got it.

PR #119, opened 2018-09-04
support for onnx-1.3

guschmue:
Support for onnx-1.3 which was already released.
- onnx-1.3 uses opset 8 which is supported by few runtimes so we default to opset 7 for now. The default can be overwritten with ```--opset 8```.
- print the used opset at startup
- set default target to nothing, aka no old caffe2 workarounds are going to be used by default

@pengwa, can you review ?

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -15,7 +15,8 @@ import traceback
 
 import numpy as np
 import tf2onnx
-from onnx import helper, onnx_pb, numpy_helper
+from onnx import helper, onnx_pb, numpy_helper, defs
+
 from tensorflow.python.framework import graph_util
 from tensorflow.tools.graph_transforms import TransformGraph
 from tf2onnx import utils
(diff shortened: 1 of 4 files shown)


tf2onnx/tfonnx.py
```diff
@@ -31,7 +32,17 @@
...
 POSSIBLE_TARGETS = [TARGET_RS4, TARGET_CAFFE2]
-DEFAULT_TARGET = [TARGET_RS4, TARGET_CAFFE2]
+DEFAULT_TARGET = []
+PREFERRED_OPSET = 7
+
+
+def find_opset(opset):
+    if opset is None or opset == 0:
+        opset = defs.onnx_opset_version()
+        if opset > PREFERRED_OPSET:
```
pengwa: does this mean, though user passed --opset 8, 7 is still be used during conversion? is this expected?

pengwa: sorry just mis-understood. by default, 7 is used.

PR #122, opened 2018-09-07
fix issues with unknown dims that prevented some models to run

guschmue:
- fix RandomNormal, RandomUniform in case the shape comes from a tensor
- change the way we implement min/max broadcast in opset <8 to avoid knowing shapes


guschmue:
@pengwa ... can you take a look?

guschmue:
gracias!

PR #124, opened 2018-09-10
use tf shape inference to determin output_shape attribute

guschmue:


guschmue:
Find output_shape attribute from tensorflow shape inference. This is more likely to succeed that looking at the input since the input would need to be const.

Issue #7, opened 2018-04-19
When to convert tensorflow mobilenet frozen graph, occurred error


FrozenGene:
The mobilenet model: https://storage.googleapis.com/download.tensorflow.org/models/mobilenet_v1_1.0_224_frozen.tgz

when to convert it, prompt:
  File "/usr/local/lib/python3.6/site-packages/tf2onnx-0.0.0.1-py3.6.egg/tf2onnx/convert.py", line 67, in <module>
  File "/usr/local/lib/python3.6/site-packages/tf2onnx-0.0.0.1-py3.6.egg/tf2onnx/convert.py", line 56, in main
  File "/usr/local/lib/python3.6/site-packages/tf2onnx-0.0.0.1-py3.6.egg/tf2onnx/tfonnx.py", line 1000, in process_tf_graph
  File "/usr/local/lib/python3.6/site-packages/tf2onnx-0.0.0.1-py3.6.egg/tf2onnx/tfonnx.py", line 939, in tensorflow_onnx_mapping
  File "/usr/local/lib/python3.6/site-packages/tf2onnx-0.0.0.1-py3.6.egg/tf2onnx/tfonnx.py", line 521, in relu6_op
ValueError: negative dimensions are not allowed

guschmue:
I changed the default for unknown dimension from -1 to 1 in case the runtime does not support -1 (onnx-caffe2 ie. does not support -1)

baoruxiao:
@guschmue, still not work, could you point a direction where I can override? Thanks

leeqiaogithub:
I have the same problem when convert mobilenet_v2_1.0_224.pb to onnx.how can i do to resolve it?

guschmue:
different issue related to relu6. 
Let me see if I can fix it.

PR #126, opened 2018-09-12
allow composed relu6 with unknown dims

guschmue:
this should help https://github.com/onnx/tensorflow-onnx/issues/7, at least mobilenetv2 converts with this.


guschmue:
uh, true but only since opset8 ... I can add this for opset8.
But our runtime doesn't support this yet ... won't be able to test it.

pengwa:
@guschmue  I see. 囧😳

PR #129, opened 2018-09-13
make graph.make_const to just add const as onnx initializer

pengwa:


pengwa:
sorry, not familiar with git tools, working hard on get it done... :)

guschmue:
no worry, I think we just resolved it ...

pengwa:
thanks @guschmue , will learn that!!

PR #132, opened 2018-09-13
fix build

guschmue:
ok, this is the last "Const" - grep says so :)


pengwa:
gosh! I will definitely sync to latest next time for a change like this.

guschmue:
all unit tests and model tests are happy. yes sync before pr is easier.

PR #134, opened 2018-09-13
opset 7 supports concat only for float ... workaround with casts

guschmue:


tf2onnx/tfonnx.py
```diff
@@ -742,12 +742,44 @@ def transpose_op(ctx, node, name, args):
     return node
 
 
+def _wrap_concat_with_cast(ctx, node):
+    """wrap concat in casts for opset < 8 since it only supports."""
```
pengwa: seems you missed some words here?

guschmue: will fix it with next pr

PR #144, opened 2018-09-28
Gs/findlstm

rufelo:


pengwa:
I don't think this code is ready to merge into master. lstm support is not complete.

PR #148, opened 2018-10-04
Bidirectional lstm

pengwa:


pengwa:
I enabled TransposeOptimizer by default, and add --fold_const as a switch to enable constant_folding. 

1. when "fold_const" is by default disabled, TransposeOptimizer will work (skip nodes not matching simple usage pattern).
2. when "fold_const" is enabled, TransposeOptimizer  will be  put into fullplay.

For LSTM, 

1. if --fold_const is disabled, some case for LSTM will fail, but I think this should be acceptable for now. 
2. If --fold_const, LSTM conversion would work better.

PR #150, opened 2018-10-08
Removed duplicated g.replace_subgraph

ryotatomioka:
Removed a duplicated `g.replace_subgraph` line which caused the `replace_subgraph` function to be called twice if `ru_op.inputs[0].type == "Shape"`


pengwa:
Thanks @ryotatomioka !!

PR #152, opened 2018-10-09
make pylint happier

guschmue:
replace all ':0' addressing with port_name() in case we need to conform to c names.
fix bunch of pylint errors.


tf2onnx/graph.py
```diff
@@ -269,11 +271,11 @@ def is_target(self, name):
         """Return True if target platform is name."""
         return name in self._target
 
-    def is_initializer(self, input):
-        """Check the input is a constant value. input is in format - node_name:<int> """
-        return input in self._initializers
+    def is_initializer(self, name):
+        """Check the name is a constant value. input is in format - node_name:<int> """
```
pengwa: change this too?

pengwa:
thanks for helping fixing them. 👍

guschmue:
thanks!! I'll see to get pylint down to 0 in the next few days so we can enable it in travis.
And this port_name() change ... possible that we need to change all names to be C names (which I really don't like) ... still trying to find out what the final in onnx is.

PR #155, opened 2018-10-10
Refine lstm

pengwa:
1. change bilstm-rewriter to use simpler strategy. 
2. add case coverage for user specified sequence_length. 
3. allow initializer_c/h Fill zero assignment with dynamic shape [dynamic batch size node, hidden_size_fixed_number]. 
4. set shape for some lstm outputs, this will be helpful for ConcatV2 axis=-1 conversion. 
5. clean up debug information to make the console neat.


pengwa:
thanks @guschmue !!

PR #156, opened 2018-10-10
start to enable pylint in ci

guschmue:
things under tf2onnx/*.py are now find, optimizer, lstm and tests still need to be done.

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -389,6 +389,7 @@ class TransposeOptimizer(object):
         if self._handle_node_having_branches(node):
             node.set_attr("axis", 1)
             return True
+        return False
 
     def _pad_handler(self, trans, node):
         # [N-start, H-start, W-start, C-start, N-end, H-end,  W-end, C-end]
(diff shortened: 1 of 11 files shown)


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -125,8 +136,8 @@ def _update_graph_nodes(self, nodes_to_extend, nodes_to_remove, has_input_change
         if nodes_to_remove:
             for n in nodes_to_remove:
                 ops.remove(n)
-        
-        if nodes_to_extend and len(nodes_to_extend) > 0:
+
+        if nodes_to_extend and nodes_to_extend:
```
pengwa: by intention here?

guschmue: yes, if nodes_to_extend  is the same as len(nodes_to_extend) != 0

pengwa: I mean, we should not have repeated nodes_to_extend. let me fix this in my local change.

guschmue:
thanks!!!

PR #157, opened 2018-10-11
add more tests, fix user passed seq_length when it is a const (which …

pengwa:
…will be folded and rename during fold_constant transformation)


pengwa:
thanks !!

PR #158, opened 2018-10-12
fix get_attr exception when ConvTranspose + Add<Const>, constant is n…

pengwa:
fix get_attr exception when ConvTranspose + Add<Const>, constant is not a  Node of the Graph, so _None_.get_attr() throw the error


pengwa:
@wenbingl

PR #159, opened 2018-10-12
model converter support

zhijxu-MS:
1 make converter support Div op.
2 make converter support Batch-Matmul attribute adjoint
3 avoid generated input data to be zero, for this will make some op failed such as divide.

### DIFF
--- a/tests/run_pretrained_models.py
+++ b/tests/run_pretrained_models.py
@@ -391,6 +391,10 @@ def tests_from_yaml(fname):
 
 
 def main():
+    # suppress log info of tensorflow so that result of test can be seen much easier
+    os.environ['TF_CPP_MIN_LOG_LEVEL'] = '3'
+    tf.logging.set_verbosity(tf.logging.WARN)
+
     args = get_args()
     Test.cache_dir = args.cache
     tests = tests_from_yaml(args.config)
--- a/tests/run_pretrained_models.py
+++ b/tests/run_pretrained_models.py
@@ -46,12 +46,14 @@ def get_beach(shape):
 
 def get_random(shape):
     """Get random input."""
-    return np.random.sample(shape).astype(np.float32)
+    # add 1 to make sure generated data is not zero, for some ops will have a problem when encounter zero value, such as divide
+    return (np.random.sample(shape) + 1).astype(np.float32)
 
 
 def get_random256(shape):
     """Get random imput between 0 and 255."""
-    return np.round(np.random.sample(shape) * 256).astype(np.float32)
+    # add 1 to make sure generated data is not zero, for some ops will have a problem when encounter zero value, such as divide
+    return np.round((np.random.sample(shape) + 1e-2) * 256).astype(np.float32)
 
 
 def get_ramp(shape):
--- a/tests/run_pretrained_models.py
+++ b/tests/run_pretrained_models.py
@@ -46,14 +46,12 @@ def get_beach(shape):
 
 def get_random(shape):
     """Get random input."""
-    # add 1 to make sure generated data is not zero, for some ops will have a problem when encounter zero value, such as divide
-    return (np.random.sample(shape) + 1).astype(np.float32)
+    return np.random.sample(shape).astype(np.float32)
 
 
 def get_random256(shape):
     """Get random imput between 0 and 255."""
-    # add 1 to make sure generated data is not zero, for some ops will have a problem when encounter zero value, such as divide
-    return np.round((np.random.sample(shape) + 1e-2) * 256).astype(np.float32)
+    return np.round(np.random.sample(shape) * 256).astype(np.float32)
 
 
 def get_ramp(shape):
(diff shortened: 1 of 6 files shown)


tf2onnx/tfonnx.py
```diff
@@ -1197,21 +1197,43 @@ def fused_batchnorm_op7(ctx, node, name, args):
...
+    adjoint_b = node.get_attr("adjoint_b")
+    adjoint_b = 0 if adjoint_b is None else adjoint_b.i
+
+    adj_x = node.get_attr("adj_x")
+    adj_x = 0 if adj_x is None else adj_x.i
+
+    adj_y = node.get_attr("adj_y")
+    adj_y = 0 if adj_y is None else adj_y.i
+
+    ##since dtype cann't be complex, so action "adjoint" is same as "transpose"
```
pengwa: follow existing comments style.

tf2onnx/tfonnx.py
```diff
@@ -1197,21 +1197,43 @@ def fused_batchnorm_op7(ctx, node, name, args):
 
 
 def matmul_op(ctx, node, name, args):
+    # onnx matmul only support float16, float, double for now
```
pengwa: I understand the newly added logic just can handle non-complex data, so maybe firstly we can check whether "adjoint_a", "adjoint_b"  appears, if yes, then we check the data type (is complex, then throw, otherwise, do the new logic); if no, then your new logic will not run.

tf2onnx/tfonnx.py
```diff
@@ -1197,21 +1197,43 @@ def fused_batchnorm_op7(ctx, node, name, args):
...
+    adj_x = node.get_attr("adj_x")
+    adj_x = 0 if adj_x is None else adj_x.i
+
+    adj_y = node.get_attr("adj_y")
+    adj_y = 0 if adj_y is None else adj_y.i
+
+    ##since dtype cann't be complex, so action "adjoint" is same as "transpose"
+    transpose_a = (transpose_a + adjoint_a + adj_x) % 2
+    transpose_b = (transpose_b + adjoint_b + adj_y) % 2
+    ##if dtype cann't be complex then adjoint is same as transpose
```
pengwa: remove this comment? seems useless

tests/test_backend.py
```diff
@@ -1096,6 +1100,32 @@ def test_fill(self):
...
+        tf.reset_default_graph()
+        # test integer data
+        x_val = ((np.random.random_sample(shape)+1)*100).astype(np.int32)
+        y_val = ((np.random.random_sample(shape)+1)*100).astype(np.int32)
+        x = tf.placeholder(tf.int32, x_val.shape, name=_TFINPUT)
+        y = tf.placeholder(tf.int32, x_val.shape, name=_TFINPUT1)
+        output = tf.truncatediv(x,y, name=_TFOUTPUT)
+        actual, expected = self._run(output, tf_dict={x: x_val, y: y_val}, onnx_dict={_INPUT: x_val, _INPUT1: y_val})
+        self.assertAllEqual(expected, actual)
+
```
pengwa: as we talked offline, we need add a Div tests.

tests/run_pretrained_models.py
```diff
@@ -42,12 +42,14 @@ def get_beach(shape):
 
 def get_random(shape):
     """Get random input."""
-    return np.random.sample(shape).astype(np.float32)
+    # add 1 to make sure generated data is not zero, for some ops will have a problem when encounter zero value, such as divide
+    return (np.random.sample(shape) + 1).astype(np.float32)
```
pengwa: for cases where data cannot be zero, they should handle at that time. This is generic random function, I don't think we should change it. 

another way to avoid 0, is to use numpy api directly in *.yaml to generate non-zero data.

guschmue: If we need this a lot you could also add get_random_none_zero()

tf2onnx/tfonnx.py
```diff
@@ -1197,21 +1197,29 @@ def fused_batchnorm_op7(ctx, node, name, args):
...
+    attrs = ["transpose_a", "transpose_b", "adjoint_a", "adjoint_b", "adj_x", "adj_y"]
+    attrs_val = [node.get_attr(attr) for attr in attrs]
+    attrs_val = [0 if val is None else val.i for val in attrs_val]
+
+    if any(attrs_val[2:]):
+        # conjugation operation on complex data not supported in onnx for now, so if it's complex than raise exception
+        if node.dtype not in [onnx_pb.TensorProto.FLOAT, onnx_pb.TensorProto.FLOAT16, onnx_pb.TensorProto.DOUBLE]:
+            raise ValueError("dtype " + node.dtype + " is not supported in onnx matmul for now")
+
+    transpose_a = (attrs_val[0] + attrs_val[2] + attrs_val[4]) % 2
```
pengwa: cool! putting this under "if any" makes more sense?

pengwa: this should be fine.

PR #161, opened 2018-10-16
refactor for two api needed by keras converter.

wenbingl:


wenbingl:
Need avoid some rewriters and optimizer is in the up level of process_tf_graph, so the current one could be the sweet spot.

PR #166, opened 2018-10-18
Generalized transpose optimizer to support simplification of no-op transpose of any rank

ryotatomioka:


ryotatomioka:
Is there general guidelines about what optimization should happen in tensorflow-onnx vs. in onnx optimizers?

pengwa:
@ryotatomioka for general ONNX graph optimization, onnx should be the right place I believe. This specific TransposeOptimizer is trying removing thanspose ops introduced by this converter. 

actually after all conversion complete, we called onnx's optimizers before saving the final result model.

pengwa:
thanks @ryotatomioka !!

yuanbyu:
The general rule is that, if an optimization is more generally useful beyond TensorFlow conversion, it should be done in some place else.

PR #168, opened 2018-10-19
refine lstm conversion

pengwa:
1. simplify the rnn output identifying and connecting logic
2. separate logics into lstm specific rewriter and base class (will continue refining this later)


pengwa:
thanks @guschmue !!!

PR #175, opened 2018-10-25
Revert "fix transpose+shape for transposeoptimizer"

guschmue:
Reverts onnx/tensorflow-onnx#172


guschmue:
@pengwa ... reverted this for now ... the gather on int32 created some trouble. 
We'll deal with the trouble and re-apply the PR after that.

pengwa:
yes, sure. @guschmue

PR #178, opened 2018-11-01
add fill support in OPSET7

pengwa:
to unblock LSTM/GRU state zero initializer conversion. 

Fill with Float value can be supported with this change.

The limitation for is: Tile might not support well on int32 value as observed in onnxruntime.


pengwa:
merge this to unblock gru conversion. feel free to let me know if there is any concern.

PR #179, opened 2018-11-01
int32 not supported by Tile, so add cast

pengwa:


tf2onnx/tfonnx.py
```diff
@@ -1263,20 +1263,47 @@ def fill_op7(ctx, node, name, args):
     nodes = [node]
     fill_shape = ctx.get_shape(node.input[0])
     fill_shape_dims = fill_shape[0]
-    attr = {"axes": [0]}
+    val_dtype = ctx.get_dtype(node.input[1])
+    val_shape = ctx.get_shape(node.input[1])
+
+    need_cast = val_dtype != onnx_pb.TensorProto.FLOAT
```
guschmue: add 'or ctx.opset < 8' since opset 8 does support

pengwa: do you mean changing this way? 

need_cast = val_dtype != onnx_pb.TensorProto.FLOAT and ctx.opset < 8

guschmue: yes, my bad ... 'and' please.

pengwa: got it!

PR #180, opened 2018-11-02
add option to transpose input to from nchw to nhwc

guschmue:
- add option to transpose input to from nchw to nhwc
- allow -1 in input shape override
- updates to readme

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,4 +31,4 @@ install:
 
 script:
     - pylint --rcfile=tools/pylintrc --ignore=version.py --disable=cyclic-import tf2onnx/*.py tests/*.py tf2onnx/optimizer/transpose_optimizer.py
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -31,4 +31,4 @@ install:
 
 script:
     - pylint --rcfile=tools/pylintrc --ignore=version.py --disable=cyclic-import tf2onnx/*.py tests/*.py tf2onnx/optimizer/transpose_optimizer.py
-    - python -m pytest --cov=tf2onnx --cov-report=term tests/test_backend.py tests/test_graph.py tests/test_internals.py tests/test_lstm.py
+    - python -m pytest --cov=tf2onnx --cov-report=term tests/test_backend.py tests/backend_test_base.py tests/test_graph.py tests/test_internals.py tests/test_lstm.py
```
fatcat-z, line 34: Why do we add this base class into here? Will it be run by pytest?

PR #189, opened 2018-11-12
add tool to quantitize weights

guschmue:


tools/quantitize_weights.py
```diff
@@ -0,0 +1,182 @@
...
+    w = w_in * scale + zp
+    w = w.astype("float32")
+    return w
+
+
+def eight_bit_quantitize(w_in):
+    """quantitize to 8 bit as scale and zeropoint"""
+    low = np.min(w_in)
+    high = np.max(w_in)
+    scale = (high - low) / 256.
```
pengwa: scale won't be zero, right?

PR #194, opened 2018-11-14
fix shape issue, include tools in pylint

guschmue:
The constant_fold after burner is not active yet.

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -1342,6 +1342,10 @@ def fill_op(ctx, node, name, args):
     return node
 
 
+def erf_op(ctx, node, name, args):
+    raise NotImplementedError("erf op not implemented")
+
+
 # map tensorflow ops to onnx ops. The format below is
 # "TFOP": func_to_map, ["OnnxOp", ...]
 #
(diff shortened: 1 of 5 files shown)


tf2onnx/tfonnx.py
```diff
@@ -1638,6 +1638,47 @@ def rewrite_flatten(g, ops):
...
+    func_map = {"Sub": np.subtract,
+                "Mul": np.multiply,
+                "Add": np.add,
+                "Sqrt": np.sqrt}
+
+    keep_looking = True
+    while keep_looking:
+        keep_looking = False
+        for idx, op in enumerate(ops):
+            if op.is_deleted() or op.type in ["Identity", "TensorArrayV3", "Enter", "Assert"]:
```
pengwa: maybe change to not in func_map.keys?

tf2onnx/tfonnx.py
```diff
@@ -1638,6 +1638,47 @@ def rewrite_flatten(g, ops):
     return ops
 
 
+def rewrite_constant_fold(g, ops):
```
pengwa: cool!

tf2onnx/tfonnx.py
guschmue: yeap, will do

PR #196, opened 2018-11-15
fix bug in utils.get_shape

nbcsm:


pengwa:
add @guschmue  might have better knowledge for this code. I am curious whether  "shape[0]" possibly be None. :)

nbcsm:
a placeholder with partial shape specified?

guschmue:
This is really old code from the time before we figured that it is better to just use the shapes from shape inference ... not sure if we depend anywhere on this info today (that is why there is this FIXME comment on this method).
The ultimate fix would be to remove the method completely.
At a minimum I'd fix it by removing the ```if``` because setting it to 1 is iffy (this if must be some old crime where we could not deal with known dims).

nbcsm:
anyway, feel free to merge or abandon this pr.

pengwa:
I think the change is ok. let's refine this later. @guschmue  let me know if you have concerns here.

PR #197, opened 2018-11-16
pylint and unit test cleanup

guschmue:


pengwa:
@zhijxu-MS  can you check the broken UT case, seems there is accuracy issue there.

PR #200, opened 2018-11-20
add tf.where(e.g. Select) support with Loop+If

pengwa:


pengwa:
@guschmue  please kindly help review this, which would help unblock attention conversion.

tf2onnx/tfonnx.py
```diff
@@ -1536,12 +1547,13 @@ def mknode(op_type, inputs, opname, **kwargs):
     "Relu6": (relu6_op, []),
     "Reshape": (reshape_op, ["Reshape"]),
     "Rsqrt": (rsqrt_op, []),
+    "Select": (select_op, []),
```
guschmue: dpesn't this want to go into the opset8 table?

pengwa: nice catch! will fix it .

pengwa:
thanks a lot !

PR #202, opened 2018-11-21
add cast when tf.shape has int32 as output

pengwa:


guschmue:
as painful pylint is ... it is right most of the time :)

pengwa:
:) I tried "pylint --rcfile=tools/pylintrc --ignore=version.py --disable=cyclic-import tf2onnx tests/*.py tools" locally, is there any approach running that locally?

PR #205, opened 2018-11-22
port test env from travis to azure pipeline

zhijxu-MS:


azure-pipelines_basic_test.yml
```diff
@@ -0,0 +1,47 @@
...
+        opset_version: '7'
+      tf1.10_opset7:
+        tf_version: '1.10'
+        opset_version: '7'
+      tf1.11_opset7:
+        tf_version: '1.11'
+        opset_version: '7'
+      tf1.12_opset7:
+        tf_version: '1.12'
+        opset_version: '7'
```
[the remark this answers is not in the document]
guschmue: +1, we need to do opset 8

ci_build/steps_template_real_model_test.yml
```diff
@@ -0,0 +1,46 @@
...
+      pip install onnxruntime==${{ parameters.onnxruntime_version }}
+    fi
+    # install onnx converter-"tf2onnx"
+    python setup.py install
+    pip list
+  displayName: 'Install dependencies'
+
+- bash:
+    pylint --rcfile=tools/pylintrc --ignore=version.py --disable=cyclic-import tf2onnx tests/*.py tools
+  continueOnError: true
```
guschmue: we still want this build to fail. Does it do this?

tests/backend_test_base.py
```diff
@@ -27,12 +27,15 @@ class Tf2OnnxBackendTestBase(unittest.TestCase):
...
     BACKEND = "onnxruntime"
     OPSET = 7
+    IS_OPSET_SET = False
     DEBUG = None
 
     def debug_mode(self):
         return type(self).DEBUG
 
     def setUp(self):
+        if not self.IS_OPSET_SET and "OPSET_VERSION" in os.environ:
```
guschmue: we should either use environment variables or command line arguments.
What is wrong with the existing command line arguments?

tests/test_backend.py
```diff
@@ -171,7 +173,10 @@ def _conv_test(self, x_val, w, strides=None, padding="VALID", dilations=None, rt
         x = tf.placeholder(tf.float32, shape=x_val.shape, name=_TFINPUT)
         conv = tf.nn.conv2d(x, kernel, strides=strides, padding=padding, dilations=dilations)
         _ = tf.identity(conv, name=_TFOUTPUT)
-        self._run_test_case([_OUTPUT], {_INPUT: x_val}, rtol=rtol)
+        try:
+            self._run_test_case([_OUTPUT], {_INPUT: x_val}, rtol=rtol)
```
guschmue: we lose the original exception info ?

tests/test_gru.py
```diff
@@ -198,6 +198,7 @@ def test_single_dynamic_gru_ch_zero_state_initializer(self):
         output_names_with_port = ["output:0", "cell_state:0"]
         self.run_test_case(feed_dict, input_names_with_port, output_names_with_port, rtol=1e-03)
 
+    @unittest.skip("FIXME: missing rtol")
```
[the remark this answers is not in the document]
zhijxu-MS: sometimes the test will fail for mismatch result.
the reason is the accuracy problem. If we use random data to test, then the rol is uesless.
for example, we set rtol big enough something like 1, and the result of TF and onnxruntime may be 1000 and 1002, then pytest will consider this is an error where actually it's correct.

azure-pipelines_basic_test.yml
```diff
@@ -0,0 +1,47 @@
+# Python package
```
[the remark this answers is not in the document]
zhijxu-MS: you mean we create different version of azure pipeline yaml config?

ci_build/steps_template_real_model_test.yml
zhijxu-MS: azure pipeline will report the status as "success with issue".

reason why write it in this way is that i think maybe it can save test time to run pylint test and computation test

tests/test_backend.py
zhijxu-MS: maybe we could use python function "traceback.print_exc()" to print the exception info.

while after discussion with guoliang, decided to use skipif to skip the test when tensorflow didn't implement the function.

tests/test_backend.py
```diff
@@ -20,7 +21,7 @@
 
 # we can override BACKEND and OPSET from the command line, but that is to late
 # to change the behavior of annotation. If need, pick the backend here.
-OPSET = 7
+OPSET = int(os.environ.get("OPSET_VERSION", default=7))
```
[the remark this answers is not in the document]
zhijxu-MS: done

ci_build/steps_template_real_model_test.yml
```diff
@@ -0,0 +1,46 @@
+parameters:
```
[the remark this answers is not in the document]
zhijxu-MS: done

tests/test_backend.py
```diff
@@ -171,7 +172,10 @@ def _conv_test(self, x_val, w, strides=None, padding="VALID", dilations=None, rt
         x = tf.placeholder(tf.float32, shape=x_val.shape, name=_TFINPUT)
         conv = tf.nn.conv2d(x, kernel, strides=strides, padding=padding, dilations=dilations)
         _ = tf.identity(conv, name=_TFOUTPUT)
-        self._run_test_case([_OUTPUT], {_INPUT: x_val}, rtol=rtol)
+        try:
+            self._run_test_case([_OUTPUT], {_INPUT: x_val}, rtol=rtol)
+        except UnimplementedError:
+            print("encounter UnimplementedError exception")
```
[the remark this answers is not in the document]
zhijxu-MS: done

ci_build/steps_template_real_model_test.yml
guschmue: I just worried that we accidentally merge with pylint errors. Up to you. We can always change it if this becomes an issue.

tests/test_backend.py
guschmue: Its hard to get rid of the skipif() because sometimes results are just wrong.

tests/test_grublock.py
```diff
@@ -186,6 +187,7 @@ def test_single_dynamic_gru_ch_zero_state_initializer(self):
         output_names_with_port = ["output:0", "cell_state:0"]
         self.run_test_case(feed_dict, input_names_with_port, output_names_with_port, rtol=1e-03)
 
+    @unittest.skip("FIXME: missing rtol")
```
pengwa: change to a meaningful name?

tests/run_pretrained_models.py
```diff
@@ -436,6 +437,7 @@ def main():
                 t = tests[test]
                 if t.perf:
                     f.write("{},{},{}\n".format(test, t.tf_runtime, t.onnx_runtime))
+    return failed
```
pengwa: where is it defined?

ci_build/steps_template.yml
```diff
@@ -0,0 +1,57 @@
...
+    set -ex
+    export TF2ONNX_TEST_OPSET=${{ parameters.opset_version }}
+    # unit tests generated artifically
+    python -m pytest --cov=tf2onnx --cov-report=term  --disable-pytest-warnings tests
+
+    if [ "${{ parameters.test_case }}" == "unit_tests" ]; then
+      echo "skip pretrained models tests for now"
+    elif [ "${{ parameters.test_case }}" == "pretrained_models" ]; then
+      # real model tests
+      python tests/run_pretrained_models.py --config tests/unity.yaml --opset ${{ parameters.opset_version }}
```
pengwa: make sure hallway tests case passed.

ci_build/steps_template.yml
```diff
@@ -0,0 +1,57 @@
+parameters:
+  python_version: '3.5'
+  tensorflow_version: ''
+  opset_version: ''
+  onnx_version: ''
+  onnxruntime_version: ''
+  test_case: '' # can be 'unit_tests', 'pretrained_models'
```
nbcsm: onnxruntime_version is not generic, we may enable caffe2 some day.
test_case may not be exclusive, use a variable might be better.

.azure-pipelines_ci_test.yml
```diff
@@ -0,0 +1,34 @@
+# Python package
```
nbcsm: for CI: run unit_test AND pre_trained model with 1 fixed test setup.
2 yml files.

for nightly: run unit_test AND pre_trained model with all test matrix.
2 yml files.

nbcsm: if there are too many azure pipeline files in root dir, u may consider put them to a separate folder.

ci_build/steps_template.yml
nbcsm: personally, i prefer to use variable for non generic purpose.
we may need to add new options in the future, and u don't need to update template parameters in ALL pipeline files.

tests/test_gru.py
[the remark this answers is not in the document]
pengwa: there is some conversation following up this accuracy problem. for us, I think disabling the tests should be ok if we already have similar UT case, where weights is fixed. If there is not, we can just replace the weight with fixed value. how do you think? @zhijxu-MS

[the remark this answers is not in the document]
zhijxu-MS: we already have tests with fixed input and weight .
and relax rtol is useless, for example: when the random result is 0.1 around and we set rtol 1, then the check doesn't make any sense.
if we want tests with random data, then the only way is to retry when tests fail. while even in this way, we just decrease the probability of mismatch result to show up.

ci_build/steps_template_real_model_test.yml
```diff
@@ -0,0 +1,46 @@
...
+    if [ "${{ parameters.onnxruntime_version }}" == "" ]; then
+      pip install onnxruntime
+    else
+      pip install onnxruntime==${{ parameters.onnxruntime_version }}
+    fi
+    # install onnx converter-"tf2onnx"
+    python setup.py install
+    pip list
+  displayName: 'Install dependencies'
+
```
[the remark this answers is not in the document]
zhijxu-MS: done

tests/run_pretrained_models.py
zhijxu-MS: what do you mean by "it"?
if "it" is "failed", then it's defined the same function.

ci_build/steps_template.yml
zhijxu-MS: done

tests/backend_test_base.py
zhijxu-MS: same as guoliang's commit-65fe0338f05
two ways to sest opset version, and setting at command line can override setting at env.

PR #208, opened 2018-11-28
placeholder_with_default_op

nbcsm:
To simulate TF placeholder_with_default behavior, convert const default value to initializer.
user can specify input to override default value in initializer when running session, or specify no input so default value is used.

if the placeholder_with_default default value is non a const node, fall back to placeholder, default value is ignored.

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -286,6 +286,24 @@ def placeholder_op(ctx, node, name, args):
     return None
 
 
+def placeholder_with_default_op(ctx, node, name, args):
+    # Overriding initializer with input is not well supported for these targets
+    if ctx.is_target(TARGET_RS4, TARGET_RS5, TARGET_RS6):
+        log.warning("%s: placeholder_with_default_op, default value is not supported by target, ignore it" % node.name)
+        return placeholder_op(ctx, node, name, args)
+
+    default_value_node = ctx.get_node_by_output(node.input[0])
+    if not default_value_node.is_const():
+        log.warning("%s: placeholder_with_default_op, non const default value is not supported, ignore it" % node.name)
+        return placeholder_op(ctx, node, name, args)
+
+    # Copy the tensor value, set its name to current node's output, add as initializer
+    value = default_value_node.get_tensor_value(as_list=False)
+    tensor = numpy_helper.from_array(value, node.output[0])
+    ctx.add_initializer(tensor)
+    return None
+
+
 def square_op(ctx, node, name, args):
     node.type = "Mul"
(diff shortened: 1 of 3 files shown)


tf2onnx/tfonnx.py
```diff
@@ -292,6 +292,20 @@ def placeholder_op(ctx, node, name, args):
     return None
 
 
+def placeholder_with_default_op(ctx, node, name, args):
+    default_value_node = ctx.get_node_by_name(node.input[0])
+    if default_value_node.is_const():
+        # Clone the tensor, set its name as current node's output, add as initializer
+        t = onnx_pb.TensorProto()
```
pengwa: refer to const_op?

def const_op(ctx, node, name, args):
    """Constants - make those initializers."""
    tensor = node.get_attr("value")
    ctx.add_initializer(tensor.t)

pengwa: talked offline with Guoliang.

two ops in original tf graph. one is placeholder, one is const. the placeholder is referencing the const as default value. So here need to parse tensor value from the const, and rename the tensor's name with placeholder's name. then the placeholder will become a input node, with an initializer containing the tensor.

for the const node in original graph, it will be handled as a normal constant, and will be deleted if no other node referenced it.

guschmue: hm, the runtime supports this?

nbcsm: do not really know yet...
there is bug for required input count check currently, and someone is fixing it.

the runtime should support it since it is totally legal according to onnx spec.

guschmue: but this would mean everything out there would run into this if it uses the placeholder_default

nbcsm: i got your point. 
yes, it will crash on onnxruntime 0.1.3 when placeholder_default is used and user tried to feed input to override default value.

seems we should postpone on this pr for a while.

pengwa: I understand that's the runtime issue, right? the conversion did the right thing? we can postpone this until there is requirement for that. BTW, is there bug tracking the runtime issue?

guschmue: if you need it for some model we could enable or disable with
```ctx.is_target(TARGET_RS5) or `ctx.is_target(TARGET_RS6) ``` (should maybe make is_target accept a list).

nbcsm:
- refined code using latest graph APIs
- fallback to placeholder for target rs4 - rs6. there was a bug when overriding initializer with input, the fix comes in onnxruntime 0.1.4 before rs6.

nbcsm:
thanks for the review

PR #209, opened 2018-11-28
fix reshape input who has more than 1 unknown dims

pengwa:
if sequence inputs has two unknown dims, should get shape dynamically , and use it as input of Reshape


pengwa:
Thanks @guschmue !

PR #213, opened 2018-11-29
fix bilstm/bigru bug when bw lst/gru's output is not followed by reve…

pengwa:
…rse nodes

in a real model, if the author explicitly define a fw lstm, and a bw lstm, feed the input and reversed input respectively, then don't reverse the bw lstm ouput back. In this case, bilstm assumes that the reverse exists after bw lstm y output, so it replaced the wrong node. 

Here we will check the reverse node existence before doing bilstm rewriting.


pengwa:
thanks！

PR #214, opened 2018-11-29
Move from the deprecated FastGFile to GFile

anttisaukko:
tf.gfile.FastGFile is deprecated and will be removed in a future version.


guschmue:
thanks!!

PR #215, opened 2018-11-30
fixes around transpose to make bert models run on onnxruntime

guschmue:
@pengwa ... check the MatMul

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -515,7 +515,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
         x_val = np.array([1.0, 2.0, 3.0, 4.0], dtype=np.float32).reshape((2, 2))
         x = tf.placeholder(tf.float32, [2, 2], name=_TFINPUT)
         shape = tf.constant([4, 1])
-        x_ = tf.reshape(x, tf.transpose(shape))
+        x_ = tf.reshape(x, shape)
         _ = tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: x_val}, check_shape=True)
 
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -366,6 +366,15 @@ class BackendTests(Tf2OnnxBackendTestBase):
         _ = tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: x_val, _INPUT1: y_val})
 
+    def test_matmul3(self):
+        x_shape = [1, 12, 256, 64]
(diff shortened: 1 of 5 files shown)


tests/test_backend.py
```diff
@@ -506,7 +515,7 @@ def test_reshape_dynamic(self):
         x_val = np.array([1.0, 2.0, 3.0, 4.0], dtype=np.float32).reshape((2, 2))
         x = tf.placeholder(tf.float32, [2, 2], name=_TFINPUT)
         shape = tf.constant([4, 1])
-        x_ = tf.reshape(x, tf.transpose(shape))
+        x_ = tf.reshape(x, shape)
```
pengwa: may I know the reason for this? 
If we need do this,  test_reshape_dynamic should be redundant , test_reshape already cover the same case maybe?

tf2onnx/tfonnx.py
```diff
@@ -1304,11 +1304,19 @@ def matmul_op(ctx, node, name, args):
...
+            transpose = ctx.insert_new_node_on_input(node, "Transpose", node.input[0], perm=perm)
+            nodes.insert(0, transpose)
     if transpose_b != 0:
-        transpose = ctx.insert_new_node_on_input(node, "Transpose", node.input[1])
-        nodes.insert(0, transpose)
+        shape = ctx.get_shape(node.input[1])
+        if shape:
+            perm = list(range(0, len(shape)))
+            tmp = perm[-1]; perm[-1] = perm[-2]; perm[-2] = tmp
+            transpose = ctx.insert_new_node_on_input(node, "Transpose", node.input[1], perm=perm)
```
pengwa: why do we need transpose on last two dims?

pengwa: I understand now. 
https://www.tensorflow.org/api_docs/python/tf/linalg/matmul 
> The inputs must, following any transpositions, be tensors of rank >= 2 where the inner 2 dimensions specify valid matrix multiplication arguments, and any further outer dimensions match.

tests/test_backend.py
guschmue: turns out onnxruntime transpose works only on float. In the past that worked only because the onnx optimizer removed this transpose because perm=[0].
But now that I disabled the onnx optimizer the transpose on int is there and onnxruntime ran into it.
And of course that unit test never did what we wanted it to do since the transpose was removed.
You are right, I need to do something with the shape input - maybe pass it as placeholder will make it really dynamic,

pengwa: get it! thanks! yeah, making it into placeholder should work.

guschmue: ok, using placeholder now

PR #217, opened 2018-11-30
tf.range

nbcsm:
support tf.range using loop op

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -194,7 +194,7 @@ def identity_op(ctx, node, name, args):
             for i, parent_name in enumerate(n.input):
                 if parent_name == output_name:
                     n.input[i] = input_name
-        if not ctx.is_model_output(output_name):
+        if not output_name in ctx.output_names:
             return None
 
     ctx.copy_shape(node.input[0], node.output[0])
@@ -203,6 +203,8 @@ def identity_op(ctx, node, name, args):
 
 def range_op(ctx, node, name, args):
     """Range."""
+    # T range = Range(T start, T limit, T delta)
+    # V v_final_and_scan_outputs = Loop(int64 M, B cond, V v_initial)
     start_node = node.inputs[0]
     limit_node = node.inputs[1]
     delta_node = node.inputs[2]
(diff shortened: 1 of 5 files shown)


tf2onnx/tfonnx.py
```diff
@@ -185,14 +185,16 @@ def direct_op(ctx, node, name, args):
...
-        # if identity has a const as input, remove it
+        # if identity has a const as input, remove it if it is not an model output
         input_name = node.input[0]
         output_name = node.output[0]
         for n in ctx.get_nodes():
             for i, parent_name in enumerate(n.input):
                 if parent_name == output_name:
                     n.input[i] = input_name
-        return None
+        if not ctx.is_model_output(node.name):
```
pengwa: shall we just use node.output[0] in ctx.output_names?

tf2onnx/tfonnx.py
```diff
@@ -185,18 +187,105 @@ def direct_op(ctx, node, name, args):
...
-        return None
+        if not ctx.is_model_output(output_name):
+            return None
+
     ctx.copy_shape(node.input[0], node.output[0])
     return node
 
 
+def range_op(ctx, node, name, args):
+    """Range."""
```
pengwa: list Range signature like this? 

    # T output = Select(bool condition, T x, T y)
    # V v_final_and_scan_outputs = Loop(int64 M, B cond, V v_initial)

tf2onnx/tfonnx.py
```diff
@@ -185,18 +187,105 @@ def direct_op(ctx, node, name, args):
...
+
+    nodes = []
+
+    start_name = start_node.output[0]
+    limit_name = limit_node.output[0]
+    delta_name = delta_node.output[0]
+
+    # trip_count
+    diff_name = "{}_diff".format(base_name)
+    nodes.append(Node(helper.make_node("Sub", [limit_name, start_name], [diff_name], name=diff_name), ctx))
```
pengwa: for output id, usually we use port_name(diff_name)

tf2onnx/tfonnx.py
```diff
@@ -1591,6 +1680,7 @@ def mknode(op_type, inputs, opname, **kwargs):
     "RandomUniform": (direct_op, []),
     "RandomNormalLike": (direct_op, []),
     "RandomUniformLike": (direct_op, []),
+    "Range": (range_op, []),
```
pengwa: add into opset 7?

tf2onnx/tfonnx.py
nbcsm: sure

Issue #218, opened 2018-12-01
rewrite_flatten bug


pengwa:
![image](https://user-images.githubusercontent.com/10530022/49325370-23ca2400-f57c-11e8-8770-d10eb0e6ab46.png)

![image](https://user-images.githubusercontent.com/10530022/49325226-c7fe9b80-f579-11e8-9a1a-1627c43957ca.png)

those are the tf.layers.flatten(x), the difference between them are introduced when x has dynamic shape or not.

current logic in rewrite_flatten has a bug, because it is not checking the rewriting condition strictly, so in some models, the graph is modified in a wrong way.

![image](https://user-images.githubusercontent.com/10530022/49325260-5e32c180-f57a-11e8-9e34-dad13e7bc749.png)

this should not a simple Flatten case, so we should not rewrite it as rewrite_flatten did.

PR #219, opened 2018-12-01
fix bugs:1).lstm/gru node not set output type, which caused stridedslice_op failed 2).rerite_flatten is triggered wrongly, so add more strict check

pengwa:
fix bugs:1).lstm/gru node not set output type, which caused stridedslice_op failed 2).rerite_flatten is triggered wrongly, so add more strict check


pengwa:
the pr contains the fix for https://github.com/onnx/tensorflow-onnx/issues/218

alejandrox778:
perfecto

pengwa:
thanks a lot!

PR #220, opened 2018-12-04
fix few issues for real model

pengwa:
1. support expand_dims when the input have shape including more than one None. 
2. support tf.where conversion for condition inputs having more than one rank. 
3. less_op7 support
4. tile_op7 workaround for onnxruntime

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -43,8 +43,7 @@ TARGET_RS4 = "rs4"
 TARGET_RS5 = "rs5"
 TARGET_RS6 = "rs6"
 TARGET_CAFFE2 = "caffe2"
-TARGET_ONNXRUNTIME = "onnxruntime"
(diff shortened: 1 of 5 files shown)


tf2onnx/tfonnx.py
```diff
@@ -43,7 +43,8 @@
 TARGET_RS5 = "rs5"
 TARGET_RS6 = "rs6"
 TARGET_CAFFE2 = "caffe2"
-POSSIBLE_TARGETS = [TARGET_RS4, TARGET_RS5, TARGET_RS6, TARGET_CAFFE2]
+TARGET_ONNXRUNTIME = "onnxruntime"
```
guschmue: I'm pretty sure this will be fixed in some future version - kind of saying there should be a version in the target similar like rs4, rs5, rs6 (I did this wrong for caffe2).
The current version of onnxruntime should have the same issues as rs6 ... maybe use rs6 in this case ?

pengwa: ok, sure, will use rs6.

pengwa:
thanks a lot !

PR #221, opened 2018-12-05
tool to find common patterns

guschmue:


pengwa:
cool, what's the inspiration doing this? find the pattern automatically for fuse optimation in the future?

guschmue:
yes, bunch of people are interested in that. First cut of the tool but need some feedback what the final output should be.
Probably doesn't belong into tensorflow-onnx but few people want to help iterate on it.

Issue #211, opened 2018-11-29
custom rnn bug: in tf1.4 graph, some node lacks of shape from original tf graph, making conversion logic failed.


pengwa:
found on a real model. working on a fix.

PR #222, opened 2018-12-06
fix few issues for real models

pengwa:
1. squeeze axis must be positive in onnx. Fix conversion issue for squeeze. add ut cases.
2. tf graph won't give us all the output shapes even on newer version tf, though it's very easy to infer the shape of missing ones. so add shape_inference.py doing this before all conversion logic (even the rewriter, because rewriter also need the shape infos).
3. other issue like shape/dtype not set, causing later conversion logic failed.
4. refine topk_op using two separate nodes represent 2 results. This require additional change on get_node_by_name, which would be done later.


pengwa:
issue https://github.com/onnx/tensorflow-onnx/issues/211 is covered by this pr

pengwa:
@zhijxu-MS  you can cherry pick this pr for the issue you have on missing shapes.

pengwa:
thanks

PR #223, opened 2018-12-06
tool to experiment with onnx graphs using same libs as tensorflow-onnx

guschmue:
wanted to have this for some time ...


guschmue:
this makes use of the fact that tensorflow-onnx uses onnx as internal graph representation ... so we can use exactly the same libs as tensorflow-onnx to prototype optimizations.

Issue #226, opened 2018-12-10
wrong reshape after transpose optimizer 


nbcsm:
**yaml for repro:**
```
resnet50_v1:
  model_type: saved_model
  url: http://download.tensorflow.org/models/official/20181001_resnet/savedmodels/resnet_v1_fp32_savedmodel_NHWC.tar.gz
  model: resnet_v1_fp32_savedmodel_NHWC/1538686669
  input_get: get_beach
  inputs:
    "input_tensor:0": [64, 224, 224, 3]
  outputs:
    - ArgMax:0
    - softmax_tensor:0
#    - resnet_model/final_reduce_mean:0
```

**error message:**
        run_onnx FAIL [ONNXRuntimeError] : 1 : GENERAL ERROR : [ShapeInferenceError] Incompatible dimensions for matrix multiplication
=== RESULT: 1 failed of 1, backend=onnxruntime

**root cause**
![image](https://user-images.githubusercontent.com/32868157/49738250-e414e200-fcc9-11e8-89fa-271b69d6a2d0.png)

shape attribute in reshape node is wrong, (64, 1, 1, 2048) is expected.
it is added by post_optimize_action in transpose optimizer, which decides the shape value based on input node shape.
the problem is when a transpose is moved over a node, the node SHAPE needs to be backward transposed as well.

nbcsm:
fixed, close it.

PR #227, opened 2018-12-10
fix transpose optimizer bug

nbcsm:
please refer to #226 for details.

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -144,6 +144,8 @@ class TransposeOptimizer(object):
         log.debug("finish after " + str(iteration_cnt) + " iteration(s)")
         self.post_optimize_action()
 
+        self._g.topological_sort(self._g.get_nodes())
+
         current_counter = self._g.dump_node_statistics()
         transpose_cnt = current_counter["Transpose"]
         current_counter.subtract(previous_counter)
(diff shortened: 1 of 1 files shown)


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -144,6 +144,8 @@ def optimize(self):
         log.debug("finish after " + str(iteration_cnt) + " iteration(s)")
         self.post_optimize_action()
 
+        self._g.topological_sort(self._g.get_nodes())
```
pengwa: this is already called by post_optimize_action(), why need another one?

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -234,6 +236,18 @@ def _switch_transpose_and_node(self, node, trans):
...
         trans.input[0] = utils.port_name(node.name)
+
+        # need to transpose node shape in backward direction as well after switch
+        # otherwise, reshape added in post_optimize_action may not work correctly
+        shape = self._g.get_shape(node.output[0])
+        if shape:
+            if is_nhwc_transpose(trans):
+                mapping = [0, 3, 1, 2]
+            else:
+                mapping = [0, 2, 3, 1]
```
pengwa: this should not happen, right?

nbcsm:
thanks for review

PR #228, opened 2018-12-11
unify all node creating logic

pengwa:


pengwa:
Thanks a lot! @guschmue

PR #230, opened 2018-12-11
add get_node_by_output to differentiate get_node_by_name

pengwa:
for example a node has name A
while in most case, its output would be called A:0, A;1, ...

while sometimes, we need update A to have an output called B:0. So get_node_by_name won't work correctly in this case when we pass B:0 to it.


pengwa:
thanks a lot!

PR #235, opened 2018-12-12
fix regression of get_node_by_name change

pengwa:
initializer was not correctly retrieved with get_node_by_output. add related logic to do that.

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -2216,6 +2216,9 @@ def rewrite_incomplete_type_support(g, ops, impacted_ops):
             ignored_inputs = ignored_input_index.get(op.type)
             # insert casts on inputs if the runtime only supports float
             for i, input_node in enumerate(op.inputs):
+                #print(op.input[i])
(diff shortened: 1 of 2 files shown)


tf2onnx/tfonnx.py
```diff
@@ -2216,6 +2216,9 @@ def rewrite_incomplete_type_support(g, ops, impacted_ops):
             ignored_inputs = ignored_input_index.get(op.type)
             # insert casts on inputs if the runtime only supports float
             for i, input_node in enumerate(op.inputs):
+                #print(op.input[i])
```
guschmue: I'm sure pylint is not going to like it :)

pengwa: exactly. my bad.

pengwa: BTW, seems pylint did not notice this. same with my local pylint result.

guschmue: odd, I know it doesn't like #no_space ... oh well.

pengwa:
thanks @guschmue

PR #237, opened 2018-12-12
add tips for converting rnn models

pengwa:


pengwa:
@guschmue  cool, I can move it there when that's ready.

PR #239, opened 2018-12-13
fix expanddims_op regression

pengwa:


pengwa:
merged the single line change directly.

PR #245, opened 2018-12-17
run travis test with onnxruntime 0.1.4

nbcsm:
found test may hang with onnxruntime 0.1.4 with azure pipeline, test it against travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
   - export LD_LIBRARY_PATH="$LD_LIBRARY_PATH$PWD/build/lib"
   - conda config --set always_yes yes --set changeps1 no
   - conda install -c conda-forge protobuf numpy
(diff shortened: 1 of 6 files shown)


guschmue:
uh. If you get this to work just merge this change. We want 1.4.

nbcsm:
hmm, i finally found the root cause.

seems onnxruntime 0.1.4 GRU op implementation may (NOT always) use multiple threads for computation internally.
it will hang on machine that only have 2 cores. i believe it is thread pool related.
unfortunately, azure pipeline hosted agent VM only has 2 cores, so is travis machine i guess.

change GRU batch size to 1 can bypass multi-threading computation logic.

guschmue:
that would be a real bug in onnxruntime that needs to be fixed.

guschmue:
got it. the travis machines only do sse what might have slightly different results.

nbcsm:
@guschmue, seems avgpool in onnxruntime 0.1.4 does not work correctly on machine with avx2, so disable it for 0.1.4.

azure pipeline can pass as well.
https://dev.azure.com/tensorflow-onnx/tensorflow-onnx/_build/results?buildId=531

PR #250, opened 2018-12-19
make transpose optimizer easy to use for an existing onnx model

pengwa:
1. if we have a onnx model, we can use simpler approach to use our TransposeOptimizer.

  ` GraphUtil.opt_transposes_with_model_proto(origin_proto, output_names=output_names_with_port)`

2. add optimizer testing cases, will add more later

3. Add LeakyRelu in transpose optimizer

4. Add ConcatV2 in converter inference, fixing Unity model issue.

5. Catch exception for transpose_optimizer, when exception happens, return original model proto 

6. fix transpose optimizer bug on relu6 opset8 support. When Max/Min have scalar input, transpose optimizer was failed.


pengwa:
where is the CI, it just disappeared....

pengwa:
seems I don't have the permission to bring it back. don't know why.

pengwa:
ah yeah, it's back. still don't know why. :(

setup.cfg
```diff
@@ -2,5 +2,5 @@
 test=pytest
 
 [tool:pytest]
-addopts=--cov=tf2onnx --ignore=tests/test_custom_rnncell.py --ignore=tests/test_const_fold.py
+addopts=--cov=tf2onnx --ignore=tests/test_custom_rnncell.py --ignore=tests/test_const_fold.py --ignore=tests/test_optimizers.py
```
nbcsm: consider something like unittest.skipIf(onnxruntime.__version__ == 0. 1.3) ?

pengwa: 0.1.4 seems to be ready soon, I will enable it after your pr is merged.

pengwa:
thanks for the time to review.

PR #252, opened 2018-12-24
fix rs6 target bug

lucienwang1009:
when the current node isn't the only consumer of the parent Cast, don't modify it.


lucienwang1009:
all tests with `--target rs6` passed.

pengwa:
cool, thanks @lucienwang1009 !

PR #256, opened 2018-12-26
fix Range op bugs

lucienwang1009:
@nbcsm,  If output nodes of Range have not been updated into the graph, replace_all_inputs will throw exceptions.


nbcsm:
what exactly is the exception?

lucienwang1009:
conversion is ok, but onnxruntime will throw "`Input` must be a graph initializer or input" where `Input` is the output of `Range`.

nbcsm:
close it, will get this done in another pr.

PR #262, opened 2019-01-03
fix infinite loop in infer_shape_from_outputs for select op

nbcsm:
- for a select node, if output has shape AND input[0] does not have shape AND input[1] or input[2] has shape, infer_shape_for_graph will hang.
fix the bug

- previous ```infer_shape_from_outputs``` does not depend on output at all
refactor the code a little bit

### DIFF
--- a/tf2onnx/shape_inference.py
+++ b/tf2onnx/shape_inference.py
@@ -11,7 +11,6 @@ from __future__ import unicode_literals
 import logging
 from onnx import onnx_pb
 
-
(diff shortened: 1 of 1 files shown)


nbcsm:
@lucienwang1009

tf2onnx/shape_inference.py
```diff
@@ -11,7 +11,6 @@
 import logging
 from onnx import onnx_pb
 
-
```
pengwa: why remove this?

nbcsm: pycharm likes that style...

PR #266, opened 2019-01-08
change Graph a bit to prepare for loop conversion changes

pengwa:
1. make test_optimizer test pass
2. allow node search in nested graph

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,5 +2,5 @@
 test=pytest
 
 [tool:pytest]
-addopts=--cov=tf2onnx --ignore=tests/test_custom_rnncell.py --ignore=tests/test_const_fold.py --ignore=tests/test_optimizers.py
+addopts=--cov=tf2onnx --ignore=tests/test_custom_rnncell.py --ignore=tests/test_const_fold.py
 #testpaths=tests/test_*.py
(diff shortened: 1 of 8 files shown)


pengwa:
**One key change is to support while_loop:**
Graph can contain Graph by node's attribute. In ONNX spec, GraphProto has NodeProto, NodeProto has Attribute, and Attribute can be a GraphProto. So for us: Graph has Node, Node has attribute, attribute can be a Graph. 

**How Graph and its children Graph connect together**
So I managed the graph in Graph, and stored contained_graphs using node name and attribute name as key. 
Also, there is parent_graph, which is used to track the parent Graph of current Graph. 

**Recursive tensorflow_onnx_mapping calls for nested Graphs**
in pre-rewriters, while_loop/rnn is modified to Scan/Loop, they are in the main Graph, and Scan/Loop's body graphs is stored in main Graph's contained_graphs, tagged with Scan/Loop/s node name and attribute name of GraphProto type. 

So the tensorflow_onnx_mapping will be called firstly for main Graph. When Scan/Loop node is handled, inner tensorflow_onnx_mapping is called for Scan/Loop's contained graphs, once it is done, Scan/Loop node is converted. The main Graph conversion continues. 

This is the nested data structure, and nested conversion logic I am proposing. Mostly it is decided by the nature of ONNX resursive graph feature.

**When the attribute of GraphProto type is updated to proto?**
for each node, attribute proto update is in Node's update_proto(). For attribute of GraphProto, 
1. we firstly get the Graph from node.Graph.contained_graphs, 
2. and call Graph.make_graph() to get the GraphProto.
3. set_attr() for the attribute. 
So this is also a recursive call.

pengwa:
thanks @guschmue.

PR #274, opened 2019-01-15
nested graph changes and bug fixes

pengwa:
1. make replace_all_inputs/find_output_consumers/topological_sort support recursive call across Graphs.
2. transpose optimizer use conversion Graph instead of re-parsing for built onnx model proto; add logic to recursively mapping nodes, because parsing onnx model proto and running onnx shape inferencing is not stable. 
3. allow tf2onnx/tfonnx.py to use customized input boundary checker
4. fix stridedslice conversion bug
5. allow late rewriters to run recursively

BTW, the resnet v1 transpose optimizer bug is also fixed.


pengwa:
thanks a lot! @guschmue

PR #275, opened 2019-01-15
fix some bugs

lucienwang1009:
- greater: in opset7, greater only accepts float, float16 and double input. If inputs are other types, we need to cast them to supported type first.
- reverse_sequence: in the time major setting, current converter only supports 3-dimension sequences, it's easy to generate it to 2 and larger dimensions.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1306,6 +1306,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
         _ = tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: x_val})
         tf.reset_default_graph()
+
         x_val = np.array([[1, 2, 3], [1, 2, 3], [1, 2, 3],
                           [4, 5, 6], [4, 5, 6], [1, 1, 1],
                           [0, 0, 0], [7, 8, 9], [0, 0, 0]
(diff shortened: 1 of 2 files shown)


tf2onnx/tfonnx.py
```diff
@@ -1491,7 +1492,10 @@ def reverse_op8(ctx, node, name, args):
     if time_major:
         old_shape = ctx.get_shape(node.input[0])
         old_dtype = ctx.get_dtype(node.input[0])
-        perm_val = [1, 0, 2]
+        perm_val = [1, 0]
+        rank = len(old_shape)
+        utils.make_sure(rank>=2, "rank of reverse_sequence input {} is at least 2".format(node.input[0]))
```
pengwa: spaces besides >= ?

lucienwang1009: fixed already

tests/test_backend.py
```diff
@@ -1309,6 +1309,27 @@ def test_reverse_sequence_time_major(self):
...
 
+    @unittest.skip("FIXME: the newest onnxruntime wheel hasn't been published to PYPI, so scan op is not supported")
+    def test_reverse_sequence_time_major_two_dimension(self):
+        x_val = np.array([[1, 2, 3], [1, 2, 3], [1, 2, 3],
+                          [4, 5, 6], [4, 5, 6], [1, 1, 1],
+                          [0, 0, 0], [7, 8, 9], [0, 0, 0]
+                         ],
+                         dtype=np.float32)
+        x = tf.placeholder(tf.float32, [9, None], name=_TFINPUT)
+        x_ = tf.reverse_sequence(x, seq_axis=0, batch_axis=1, seq_lengths=[9, 9, 9])
```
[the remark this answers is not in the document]
lucienwang1009: the current conversion which uses Scan only support reversing the entire sequence or the case seq_length is larger than the non-zero length of sequence along seq_axis.

PR #276, opened 2019-01-15
Graph Input as "Dummpy" Node (just as initializer) for easier manipulation. 

pengwa:
1. Graph Input as "Dummpy" Node (just as initializer) for easier manipulation. 
2. add output name checks.


pengwa:
thanks!

PR #278, opened 2019-01-16
Tf.while loop support & custom rnn refactoring

pengwa:
Changes include:

1. **Enable nested graph node searching**. 
 1.1 get_implicit_inputs is refactored to get sub graph based on Graph's containedgraphs instead of searching attributes of type GraphProto. 
 1.2 The problem: node output dtype/shape should also be supported for recursive search because inner graph will have reference to outer scope node, so we have some places allowing retrieving shape/dtype with output id including port. Another thing we need consider is that: we sometime want to set a shape with output id (like topk_k:1), so we may hold a Graph instance g, but what if topk_k:1 does not exist in existing Graph._dtypes or _output_shapes? With current implementation, we cannot figure out where(e.g. which Graph) to store the newly inserted output id's dtyep/shape info. 
       The proposal: we need put dtype/shape in Node instead of Graph. Then, if we want to get/set a dtype with an output id, firstly get the node (recursively), then get/set the dtype. In this PR, I did this. BUT we still need a change to move the storage into Node. 

2. **Loop parsing and conversion** 
   2.1 Refactor custom_rnn_rewriter, put some logic into its base class LoopRewriterBase. 
   2.2 LoopRewriter is implemented inheriting LoopRewriterBase

3. **Remove useless code**:
   Originally the process we handle custom rnn conversion is: 1). parse loop 2). check is rnn 3). find subgraph boundary node (ids), put it into a global dict indexed with Scan node's name. 4). op mapping 5). check all Scan node, find  its' subgraphs inputs/outputs, get all related nodes, create a Graph and set it into Scan's attribute. 
   So we have rewrite_custom_rnn_cell at the beginning , and also have a rewrite_custom_rnn_body_graph, which is in the later rewriter.

   Now, when we do rewriting in the beginning, we organize nodes in recursive mode, and op mapping is done recursively as well, we don't have extra need in later rewriter except the rs6 target flag.

4. TF graph shape inference for TensorArray ops.
    It is observed in internel models, some output don't have shape returned calling TF API, so enriching the TA ops shape inferencing (which is very simple) help us get there easily. The reason I did not check it in seperately: I found enabling this in a single change will make some testing failed. 

**About testing:**

- python tests\backend.py --opset 7/8
- python tests\lstm.py
- python tests\gru(block).py
- python tests\test_loops.py
- python tests\test_graph.py
- python tests\test_optimizer.py
- python tests\test_custom_rnn.py --opset 8
- python tests\run_pretrained_models.py --backend onnxruntime  --config tests\run_pretrained_models.yaml  --fold_const --onnx-file ".\tmp" --opset 8
- python tests\run_pretrained_models.py --backend onnxruntime --config tests\internal-rnn.yaml   --fold_const --onnx-file ".\tmp" --opset 8 --target rs6   **(13 RNN models we have converted are all successful)**
- python tests\run_pretrained_models.py --backend onnxruntime --config tests\internal-cnn.yaml   --fold_const --onnx-file ".\tmp" --opset 8 --target rs6  **(5 of 6 total CNN model are successful, the other 1 is an known issue)**
- python -m tf2onnx.convert --input tests\internal\cnns\classical\resnet_v2_152_inf_graph.pb --inputs input:0 --outputs resnet_v2_152/predictions/Reshape_1:0 --output res152.onnx --verbose --fold_const

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,5 +2,5 @@
 test=pytest
 
 [tool:pytest]
-addopts=--cov=tf2onnx --ignore=tests/test_custom_rnncell.py --ignore=tests/test_const_fold.py
+addopts=--cov=tf2onnx --ignore=tests/test_custom_rnncell.py --ignore=tests/test_const_fold.py --ignore=tests/test_loops.py
 #testpaths=tests/test_*.py
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -251,9 +251,9 @@ class Tf2OnnxGraphTests(unittest.TestCase):
 
             g = process_tf_graph(sess.graph)
             self.assertEqual(
-                'digraph { kernel [op_type=Reshape] Conv2D__3 [op_type=Transpose] Conv2D__2 [op_type=Transpose] '
+                'digraph { Conv2D__2 [op_type=Transpose] kernel [op_type=Reshape] Conv2D__3 [op_type=Transpose] '
                 'Conv2D [op_type=Conv] Conv2D__4 [op_type=Transpose] output [op_type=Identity] '
-                'k:0 -> kernel "kernel/shape":0 -> kernel kernel:0 -> Conv2D__3 input1:0 -> Conv2D__2 '
+                'input1:0 -> Conv2D__2 k:0 -> kernel "kernel/shape":0 -> kernel kernel:0 -> Conv2D__3 '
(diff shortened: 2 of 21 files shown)


pengwa:
Sorry for the huge change, while I have tried to checked in changes not closely separately.

guschmue:
no worry, I'll try to get to it today.

pengwa:
thanks so much!

PR #284, opened 2019-01-22
support auto_pad for add_padding when input_shape is unknown

zhijxu-MS:
when input_shape is unknown, use auto_pad instead of calculating pads.


nbcsm:
pool_op is used by Conv, MaxPool, AveragePool. could you add tests for the other ops to make sure auto_pad work as expected?

PR #286, opened 2019-01-24
fix onnx version to 1.3, as onnx 1.4 is unstable

lucienwang1009:


guschmue:
what's the problem with onnx-1.4 ?

lucienwang1009:
@guschmue installation of onnx-1.4.0 with pip requires building something in local machine which needs protobuf. onnx-1.4.1 has fixed the issue.

PR #288, opened 2019-01-25
fix shape inference bug about All, Any and Min

lucienwang1009:
@pengwa axis in All, Any and Min can be a list which may lead to current shape inferencing fails.


pengwa:
fix the pylint pls.

PR #295, opened 2019-01-29
update greater and less for opset9

nbcsm:
- merge current greater and less op map functions
- do not cast inputs of greater/less for opset9, since all numerical types are supported in opset9 now
- add test case for less


nbcsm:
link https://github.com/onnx/tensorflow-onnx/issues/270

PR #296, opened 2019-01-29
Fix Erf__sign divided by zero issue.

zhanghuanrong:
We meet Erf_sign divide by zero when running converting bert model. Fix it.


nbcsm:
@zhanghuanrong, thanks for the fix.
could you help to fix the CLA check please?

zhanghuanrong:
Seems I used wrong email when push the change. I will create a new pull request with same content but correct email. please ignore this one. I will close this after the other one finished.

PR #299, opened 2019-02-01
back out experimental op dynamicslice support

pengwa:
should not use experimental op, otherwise it's hard to get all converted models available for common runtimes.


pengwa:
thanks!

PR #302, opened 2019-02-04
remain Const/Placeholder as it is, util we make_graph at last

pengwa:
In this change, Const and Placeholder nodes are not converted during op mapping, instead, we make them into inputs/initializers during make_graph. If we want to add new input, we actually add new Placeholder node.

as mentioned in https://github.com/onnx/tensorflow-onnx/issues/280:,this change would be helpful since:
1). we can avoid creating dummy nodes just for code convenience and need take care of deletion when calling set_nodes(). 
2). we don't need main initializer in the Graph object (they are represented as node in the Graph), initializer is created in make_graph() only.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1538,7 +1538,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
         res = tf.nn.sparse_softmax_cross_entropy_with_logits(labels=label, logits=logits)
         _ = tf.identity(res, name=_TFOUTPUT)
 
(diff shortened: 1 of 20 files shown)


pengwa:
thank a lot!

wenbingl:
@pengwa @guschmue , The Keras converter depends on the converted ONNX-initializer as one kind of 'node to node' conversion API we agreed before. This change will break the 'node to node' conversion. Can we add some option to keep the old behavior?

pengwa:
@wenbingl I don't understand what is broken. can you explain how keras converter depends on tf-onnx converter? or you can call me online.

PR #306, opened 2019-02-13
fix bug when only "state" of gru is consumed;

zhijxu-MS:


tests/test_gru.py
```diff
@@ -320,6 +368,38 @@ def test_dynamic_bigru_output_consumed_only(self):
...
+    def test_dynamic_bigru_state_consumed_only(self):
+        units = 5
+        batch_size = 1
+        x_val = np.array([[1., 1.], [2., 2.], [3., 3.]], dtype=np.float32)
+        x_val = np.stack([x_val] * batch_size)
+
+        x = tf.placeholder(tf.float32, x_val.shape, name="input_1")
+        initializer = init_ops.constant_initializer(0.5)
+
+        gru_list = []
```
nbcsm: gru_list seems not used?
so are the others.

tests/run_pretrained_models.py
```diff
@@ -23,6 +23,7 @@
 import tensorflow as tf
 from tensorflow.core.framework import graph_pb2
 from tensorflow.python.framework.graph_util import convert_variables_to_constants
+from tensorflow.contrib.rnn import GRUBlockCell  # pylint: disable=unused-import
```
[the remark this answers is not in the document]
zhijxu-MS: done.

tests/test_gru.py
zhijxu-MS: done.

PR #307, opened 2019-02-13
fix bugs from ops

zhijxu-MS:
1. "ConstantLike" is renamed to "ConstantOfShape"

2. "pack" does nothing when receiving only one input tensor

3. conv kernel may be used by multiple nodes, so it can't be transposed in-place

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -1267,6 +1267,12 @@ def minmax_op(ctx, node, name, args):
 
 
 def pack_op(ctx, node, name, args):
+    # in tf, "pack" can accept one input tensor which means doing nothing,
+    # so remove the node in ONNX
+    if len(node.inputs) == 1:
+        ctx.replace_all_inputs(ctx.get_nodes(), node.output[0], node.input[0])
+        return None
+
     # hack to make up for the missing onnx pack op
     axis = node.get_attr("axis").i
     if axis < 0:
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -422,22 +422,17 @@ def conv_convert_inputs(ctx, node, with_kernel=False, new_kernel_shape=None,
     # kernel must to be transposed
     if with_kernel:
         parent = node.inputs[1]
-        if node.inputs[1].is_const():
-            # kernel is const - transpose the const
-            if not parent.data_format:
-                val = parent.get_tensor_value(as_list=False)
-                val = val.transpose(HWCN_TO_NCHW)
-                parent.set_tensor_value(val)
-        else:
-            # kernel comes from op, insert transpose op
-            input_name = node.input[1]
-            transpose = ctx.insert_new_node_on_input(node, "Transpose", input_name)
-            transpose.set_attr("perm", HWCN_TO_NCHW)
-            transpose.inserted_nchw = True
-            ctx.copy_shape(input_name, transpose.output[0])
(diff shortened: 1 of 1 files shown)


tf2onnx/tfonnx.py
```diff
@@ -422,22 +422,15 @@ def conv_convert_inputs(ctx, node, with_kernel=False, new_kernel_shape=None,
...
-            # kernel comes from op, insert transpose op
-            input_name = node.input[1]
-            transpose = ctx.insert_new_node_on_input(node, "Transpose", input_name)
-            transpose.set_attr("perm", HWCN_TO_NCHW)
-            transpose.inserted_nchw = True
-            ctx.copy_shape(input_name, transpose.output[0])
-            new_shape = spatial_map(ctx.get_shape(input_name), HWCN_TO_NCHW)
-            ctx.set_shape(transpose.output[0], new_shape)
-            nodes.append(transpose)
+        # kernel comes from op, insert transpose op
```
nbcsm: plz copy & paste the commit description here as well.

tf2onnx/tfonnx.py
```diff
@@ -422,22 +422,15 @@ def conv_convert_inputs(ctx, node, with_kernel=False, new_kernel_shape=None,
...
+        input_name = node.input[1]
+        transpose = ctx.insert_new_node_on_input(node, "Transpose", input_name)
+        transpose.set_attr("perm", HWCN_TO_NCHW)
+        transpose.inserted_nchw = True
+        ctx.copy_shape(input_name, transpose.output[0])
+        new_shape = spatial_map(ctx.get_shape(input_name), HWCN_TO_NCHW)
```
nbcsm: please add a test for this case when you check in the transpose optimizer change.

tf2onnx/tfonnx.py
```diff
@@ -1267,6 +1267,11 @@ def minmax_op(ctx, node, name, args):
 
 
 def pack_op(ctx, node, name, args):
+    # in tf, "pack" can accept one input tensor which means doing nothing,
+    # so remove the node in ONNX
+    if len(node.inputs) == 1:
+        ctx.replace_all_inputs(ctx.get_nodes(), node.output[0], node.input[0])
+        return None
```
nbcsm: [optional] personally, i prefer to insert a new line here.

tf2onnx/tfonnx.py
```diff
@@ -1267,6 +1260,11 @@ def minmax_op(ctx, node, name, args):
...
     # hack to make up for the missing onnx pack op
```
nbcsm: [optional] personally, i prefer to insert a new line here.

tf2onnx/tfonnx.py
zhijxu-MS: I thought the code reader may not need to know there is a bug here before.

tf2onnx/tfonnx.py
zhijxu-MS: this func is used in many ops and they already have unit tests.

 of course, this change may lead to more transpose nodes since sometime transposing a constant can be done in previous code while now transposing is done at runtime. And we had a discussion before the we may need to add a optimization phase to eliminate unnecessary ops and ops that can be done statically.

tf2onnx/tfonnx.py
zhijxu-MS: agree, showing that this is a code path for special case.

PR #310, opened 2019-02-18
fix few issues found in real model

pengwa:
1. Remove the logic in _convert_shapenode_to_int64 which was trying to update const node value(actually type) in place. This will be bad when the constant is used by other node as well.  
 if we modify its value(including type) in place, other node might be failed to use it. I know this might introduce an extra Cast when the const node is really used once. While we can optimize this after the model is converted successfully.
2. Pad op more types support. ONNX pad support onnx_pb.TensorProto.FLOAT16, onnx_pb.TensorProto.FLOAT, onnx_pb.TensorProto.DOUBLE only, so add Cast logic to support more types.
3. reduce_logic_op. There is a bug when there are multiple conjoint Cast for example "int const node 7 --> bool const node True --> float const node 7.0".  https://github.com/Microsoft/onnxruntime/issues/433
We can use a very cheap change to avoid this.

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -147,14 +147,6 @@ def _convert_shapenode_to_int64(ctx, node, input_number):
     """cast int32 shape into int64 shape."""
     shape_node = node.inputs[input_number]
     name = node.input[input_number]
-    if shape_node.is_const():
-        # if it is a const, change the const to be int64
-        shape = shape_node.get_tensor_value()
-        shape = np.array(list(shape), dtype=np.int64)
-        shape_node.set_tensor_value(shape)
-        ctx.set_dtype(shape_node.output[0], onnx_pb.TensorProto.INT64)
-        ctx.copy_shape(name, shape_node.output[0])
-        return [node]
 
     cast_node = ctx.insert_new_node_on_input(node, "Cast", name)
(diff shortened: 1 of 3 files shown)


tf2onnx/tfonnx.py
```diff
@@ -1222,11 +1232,6 @@ def minmax_op(ctx, node, name, args):
 
 
 def pack_op(ctx, node, name, args):
-    # in tf, "pack" can accept one input tensor which means doing nothing,
-    # so remove the node in ONNX
-    if len(node.inputs) == 1:
-        ctx.replace_all_inputs(ctx.get_nodes(), node.output[0], node.input[0])
-        return None
```
nbcsm: @zhijxu-MS to take a look at this part.

pengwa: @zhijxu-MS  please help confirm this to unblock this PR. I think this should be a regression.

zhijxu-MS: original logic is right, my modification is not needed.

my fault, my issues in my real model is triggered by other problems.

pengwa: ok, got it. Thanks @zhijxu-MS for the double check.

pengwa:
thanks for the review.

PR #314, opened 2019-02-19
Update tf2onnx version due to an interface change

jiafatom:
This [PR](https://github.com/onnx/tensorflow-onnx/commit/b9f95783dd71c57532325d658782439547c1327f) removes initializers in graph.py, so keras-onnx breaks. Therefore, we want to update the version here so that we can consume the package from GitHub.


guschmue:
We don't really update this file manual in master and do so only when we publish to pypi.
The plan was to publish 0.5 by now with opset9 support but opset9 is a bit late so we haven't.
There are maybe enough changes to push a new version to pypi before opset9.

guschmue:
let me check if team is ok to push a new version this week.

guschmue:
we are going to push a release to pypi in the next few days. 
If you take a dependency, use the release branches - that will have version.py filled out with the correct info. On master the version will always next_version.0 and ```python setup.py *`` will update the git commit id in version.py.

PR #316, opened 2019-02-20
add test cases of map_fn, fix bug of shape_inference

zhijxu-MS:


pengwa:
looks good to me unless we should part map test case into test_loops.py

PR #320, opened 2019-02-20
Fall back to non optimized model proto

csteegz:
Fixes a bug where if the transpose optimizer fails, the proto will not be written.


pengwa:
thanks for the fixing.

PR #324, opened 2019-02-23
fix CI failure, refine test code

nbcsm:
- fix ci build failure when use tf 1.10
tf 1.10 requires numpy <=1.14.5 (tf tests break due to numpy api change in 1.15, but tf functionality is ok), but onnxruntime 0.2.1 crashes with numpy <=1.14.5.
upgrade numpy to higher version for tf 1.10

- use latest onnx for all test matrix
1.4 currently.

- add test config, refine test filter, refactor test base
  - introduce test config, which allows overriding via env var and command line arguments.
  - when run from script, load test config before test filters decorators, so the test filters can take effect correctly. previously test filters does not work properly when run from script.
  - extract common test filters, which can be used within all test scripts, not just backend tests.
  - refactor test base, simplify main for test class 

- enable all test scripts for ci, fix/disable some tests.


nbcsm:
thanks for the review.

nbcsm:
finally, it is green again. :-)

image

PR #325, opened 2019-02-26
run_pretrained_model delete TMPPATH after test

nbcsm:
TMPPATH causes out of disk often on internal agent, delete it after test.


nbcsm:
tks for the review.

Issue #231, opened 2018-12-11
merge duplicated transpose ops


nbcsm:
after transpose optimizer, a transpose op might be broadcasted to MULTIPLE downstream edges.
some of them should be merged if we cannot move the transpose deeper or eliminate it.
this may help performance potentially.

here is an example:
![image](https://user-images.githubusercontent.com/32868157/49782794-87a9d500-fd52-11e8-98b7-fc95c427d28c.png)

PR #326, opened 2019-02-27
merge transpose nodes which share same input and perm attr are also same

zhijxu-MS:
current strategy is to move down transpose nodes if possible while this is
is not always a good idea for it may generate more transpose nodes which
not removed in later procedures. so it's better to merge them back into
one after removing procedures.

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -1,5 +1,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT license.
+
 """Transpose Optimizer."""
 
 from __future__ import unicode_literals
@@ -114,9 +115,9 @@ class TransposeOptimizer(object):
         self._g.update_proto()
         self._g.topological_sort(self._g.get_nodes())
 
-    def merge_transpose_with_same_input(self):
+    def merge_duplicated_transposes(self):
         # strategy used in previous procedure is to move transpose nodes down if possible,
-        # and it means that when a node has n outputs then n transpose will be generated ,
+        # and it means that when a node has n outputs then n transpose will be generated,
         # so we should merge them back to one if they can't be eliminated in previous procedure.
         graph = self._g
         input_transposes_map = defaultdict(list)
@@ -125,7 +126,7 @@ class TransposeOptimizer(object):
                 key = (node.input[0], str(node.get_attr("perm").ints))
                 input_transposes_map[key].append(node)
 
-        for _, transposes in input_transposes_map.items():
+        for transposes in input_transposes_map.values():
             # merge transpose nodes into one: make nodes use the output of the first transpose node
             transpose_out = transposes[0].output[0]
             for node in transposes[1:]:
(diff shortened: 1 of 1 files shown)


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -140,7 +162,9 @@ def optimize(self):
                 break
 
         log.debug("finish after " + str(iteration_cnt) + " iteration(s)")
+
         self.post_optimize_action()
```
pengwa: why not put it before self.post_optimize_action

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -113,6 +114,27 @@ def post_optimize_action(self):
         self._g.update_proto()
         self._g.topological_sort(self._g.get_nodes())
 
+    def merge_transpose_with_same_input(self):
+        # strategy used in previous procedure is to move transpose nodes down if possible,
+        # and it means that when a node has n outputs then n transpose will be generated ,
```
pengwa: extra space

nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/231

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -113,6 +114,27 @@ def post_optimize_action(self):
...
+        # so we should merge them back to one if they can't be eliminated in previous procedure.
+        graph = self._g
+        input_transposes_map = defaultdict(list)
+        for node in graph.get_nodes():
+            if node.type == "Transpose":
+                key = (node.input[0], str(node.get_attr("perm").ints))
+                input_transposes_map[key].append(node)
+
+        for _, transposes in input_transposes_map.items():
```
nbcsm: [optional] for transposes in input_transposes_map.values(): ?

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -113,6 +114,27 @@ def post_optimize_action(self):
         self._g.update_proto()
         self._g.topological_sort(self._g.get_nodes())
 
+    def merge_transpose_with_same_input(self):
```
nbcsm: a better name? technically, not just same input, also same perm.
e.g. merge_duplicated_transpose?

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -113,6 +114,27 @@ def post_optimize_action(self):
         self._g.update_proto()
         self._g.topological_sort(self._g.get_nodes())
 
+    def merge_transpose_with_same_input(self):
+        # strategy used in previous procedure is to move transpose nodes down if possible,
```
nbcsm: please also add corresponding test, either in this pr or new pr.

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -114,9 +115,9 @@ def post_optimize_action(self):
         self._g.update_proto()
         self._g.topological_sort(self._g.get_nodes())
 
-    def merge_transpose_with_same_input(self):
+    def merge_duplicated_transpose(self):
```
pengwa: nit: merge_duplicated_transposes

PR #328, opened 2019-02-27
add leakyrelu rewriter and related test

zhijxu-MS:


tf2onnx/rewriter/leakyrelu_rewriter.py
```diff
@@ -0,0 +1,55 @@
...
+"""
+
+from tf2onnx.graph_matcher import OpTypePattern, GraphMatcher
+
+
+# pylint: disable=missing-docstring
+
+
+def rewrite_leakyrelu(g, ops):
+    if g.opset < 6:
```
pengwa: why do this check? the reason?

tf2onnx/rewriter/leakyrelu_rewriter.py
```diff
@@ -0,0 +1,55 @@
...
+                OpTypePattern('*', name='mul_input'),
+            ]),
+            OpTypePattern('*', name='max_input'),
+        ])
+
+    matcher = GraphMatcher(pattern, allow_reorder=True)
+    match_results = list(matcher.match_ops(ops))
+    for match in match_results:
+        max_node = match.get_op('max')
+        max_input_node = match.get_op('max_input')
```
pengwa: don't need check max_input node's value to be 0?  what if it be 1, is this conversion correct?

tf2onnx/rewriter/leakyrelu_rewriter.py
zhijxu-MS: because leakyrelu is supported since opset 6

tf2onnx/rewriter/leakyrelu_rewriter.py
zhijxu-MS: if it's 0 then both input to max is 0 and output is 0
if it's 1 then max will choose 1 instead of alpha*1

but alpha needs to be checked, if it's larger than 1 then converting to leakyrelu is not right....

tf2onnx/rewriter/leakyrelu_rewriter.py
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT license.
+
+"""
+tf2onnx.rewrite - rewrite tensorflow subgraph to onnx leakyrelu op
```
pengwa: rewrite**r** ....

tf2onnx/rewriter/leakyrelu_rewriter.py
pengwa: so mul_input MUST be the same node as max_input, right?

tf2onnx/rewriter/leakyrelu_rewriter.py
```diff
@@ -0,0 +1,57 @@
...
+    return ops
+
+
+def _find_edges_name_btw_nodes(sender, sinker):
+    res = []
+    for sinker_end in sinker.input:
+        for sender_end in sender.output:
+            if sinker_end == sender_end:
+                res.append(sinker_end)
+    return res
```
pengwa: res should contain only 1 element?

tf2onnx/rewriter/leakyrelu_rewriter.py
```diff
@@ -0,0 +1,57 @@
...
+    match_results = list(matcher.match_ops(ops))
+    for match in match_results:
+        max_node = match.get_op('max')
+        max_input_node = match.get_op('max_input')
+        mul_node = match.get_op("mul")
+        mul_input_node = match.get_op('mul_input')
+
+        max_input_edge_name = _find_edges_name_btw_nodes(max_input_node, max_node)
+        mul_input_edge_name = _find_edges_name_btw_nodes(mul_input_node, mul_node)
+        if max_input_edge_name == mul_input_edge_name:
```
pengwa: what does these two line check? 

        max_input_edge_name = _find_edges_name_btw_nodes(max_input_node, max_node) 
        mul_input_edge_name = _find_edges_name_btw_nodes(mul_input_node, mul_node)

pengwa:
maybe add UT cases, for example alpha > 1 or the pattern check failed ?

tf2onnx/rewriter/leakyrelu_rewriter.py
zhijxu-MS: yes.

tf2onnx/rewriter/leakyrelu_rewriter.py
zhijxu-MS: thought this function may be reused in other places

tf2onnx/rewriter/leakyrelu_rewriter.py
zhijxu-MS: to see if one input of max and mul is same.

tf2onnx/rewriter/leakyrelu_rewriter.py
pengwa: this is kind of over-design I think. let's just do it once we really need it. here I think we just need check the two nodes (max_input, mul_input) to be the same node.

zhijxu-MS: if the same node has multiple outputs, max and mul use different output then it can't be converted.

tf2onnx/rewriter/leakyrelu_rewriter.py
zhijxu-MS: I copied from other rewrite file, and found that it's the only one has this typo.

I am so lucky I think.

nbcsm:
@zhijxu-MS, we should add tests for rewriter. e.g. make sure mul+max is changed to leakyrelu.
maybe in another PR later.

PR #329, opened 2019-02-27
auto infer shape and dtype for onnx nodes

lucienwang1009:
1. Infer shapes and dtypes for ONNX nodes in the method `make_node` when creating a ONNX node. Only support the node whose inputs' shape and dtype are already set. Nodes involved in `body_graph`, such as Loop, Scan, If, are not supported for now. 
2. Explicitly set `parent_garph` if generated body graph is a sub-graph. 
3. Use `g.make_node` rather than `Node(helper.make_node())` in `insert_new_node_on_output`/`insert_new_node_on_input`.
4. Remove some useless `set_dtype` and `set_shape`.

### DIFF
--- a/tests/test_shape_inference.py
+++ b/tests/test_shape_inference.py
@@ -1,0 +1,424 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT license.
+
+"""Unit Tests for shape inference."""
+
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import numpy as np
+from onnx import TensorProto
+from tf2onnx import utils
+from tf2onnx.graph import Graph
+from backend_test_base import Tf2OnnxBackendTestBase
+from common import *  # pylint: disable=wildcard-import,unused-wildcard-import
+
+# pylint: disable=missing-docstring
+
+INPUT1 = "input1"
+INPUT2 = "input2"
+INPUT3 = "input3"
+
+
+class ShapeInferenceTests(Tf2OnnxBackendTestBase):
+    """
+    Test shape inference, it's just a subset of all cases that can be inferred shape.
+    For more information, please refer to onnx shape inference test:
+    https://github.com/onnx/onnx/blob/master/onnx/test/shape_inference_test.py
+    """
+
+    def _run_test_case(self, graph, feed_dict):
+        """Run model with onnxruntime and compare results' shape with internal shape inference."""
+        outputs = graph.outputs
+        results = []
+        raised = False
+        try:
+            results = self.run_backend(graph, outputs, feed_dict)
+        except:  # pylint: disable=bare-except
+            raised = True
+        self.assertFalse(raised)
+
+        raised = False
+        for actual, inferred in zip(results, outputs):
+            actual_shape = actual.shape
+            inferred_shape = tuple(graph.get_shape(inferred))
+            try:
+                utils.merge_shapes(actual_shape, inferred_shape)
+            except:  # pylint: disable=bare-except
+                raised = True
+            self.assertFalse(raised)
+
+            actual_dtype = actual.dtype
+            inferred_dtype = utils.ONNX_TO_NUMPY_DTYPE[graph.get_dtype(inferred)]
+            self.assertEqual(actual_dtype, inferred_dtype)
+
+    def _create_empty_graph(self, inputs, shapes, dtypes):
+        graph = Graph([], target=self.config.target, opset=self.config.opset)
+        for inp, shape, dtype in zip(inputs, shapes, dtypes):
+            graph.add_graph_input(inp, dtype, shape)
+        return graph
+
+    def _generate_random_inputs(self, inputs, shapes, dtypes):
+        """Generate random input of shape and ONNX dtypes"""
+        res = {}
+        for inp, shape, dtype in zip(inputs, shapes, dtypes):
+            new_shape = [1 if s == -1 else s for s in shape]
+            res[inp] = np.random.random(new_shape).astype(utils.ONNX_TO_NUMPY_DTYPE[dtype])
+        return res
+
+    # one input
+    def test_identity(self):
+        inputs = [INPUT1]
+        shapes = [[1, 3, 4, 1]]
+        dtypes = [TensorProto.FLOAT]
+        graph = self._create_empty_graph(inputs, shapes, dtypes)
+        node = graph.make_node("Identity", [INPUT1])
+        graph.add_graph_output(node.output[0])
+        self._run_test_case(graph, self._generate_random_inputs(inputs, shapes, dtypes))
+
+    def test_transpose(self):
+        inputs = [INPUT1]
+        shapes = [[1, 3, 4, 1]]
+        dtypes = [TensorProto.FLOAT]
+        graph = self._create_empty_graph(inputs, shapes, dtypes)
+        node = graph.make_node("Transpose", [INPUT1], attr={"perm": [1, 0, 2, 3]})
+        graph.add_graph_output(node.output[0])
+        self._run_test_case(graph, self._generate_random_inputs(inputs, shapes, dtypes))
+
+    def test_shape(self):
+        inputs = [INPUT1]
+        shapes = [[1, 3, 4, 1]]
+        dtypes = [TensorProto.FLOAT]
+        graph = self._create_empty_graph(inputs, shapes, dtypes)
+        node = graph.make_node("Shape", [INPUT1])
+        graph.add_graph_output(node.output[0])
+        self._run_test_case(graph, self._generate_random_inputs(inputs, shapes, dtypes))
+
+    @check_opset_min_version(2, "Split")
+    def test_split(self):
+        inputs = [INPUT1]
+        shapes = [[5, 6, 7]]
+        dtypes = [TensorProto.FLOAT]
+        graph = self._create_empty_graph(inputs, shapes, dtypes)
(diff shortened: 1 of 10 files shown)


nbcsm:
add more background:
we found several dtype/shape missing issues when converting real models.
this pr tries to address some of such issues automatically.
so developer does not need to specify dtype/shape **everywhere**, only need to do it when necessary.

tf2onnx/graph.py
```diff
@@ -252,6 +257,22 @@ def update_proto(self):
         if attr:
             self._op.attribute.extend(attr)
 
+    def build_onnx_node(self):
```
nbcsm: seems it is only used in infer_onnx_shape_dtype?
move to utils or a inline function of infer_onnx_shape_dtype?

tf2onnx/graph.py
```diff
@@ -410,8 +431,49 @@ def make_node(self, op_type, inputs, attr=None, output_count=1, outputs=None, sk
...
+            return
+        log.debug("Infer shape and dtype for [%s]", node.name)
+        input_shapes = [self.get_shape(i) for i in node.input]
+        input_dtypes = [self.get_dtype(i) for i in node.input]
+
+        dtypes = {}
+        shapes = {}
+        try:
+            shapes, dtypes = utils.infer_onnx_shape_dtype(node, input_shapes, input_dtypes, self._opset)
+        except Exception:
```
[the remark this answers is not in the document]
lucienwang1009: Now, we won't throw any exception in `infer_onnx_shape_dtype`, so the exception caught here should be caused by ONNX infer_shape.

tf2onnx/graph.py
```diff
@@ -410,8 +431,49 @@ def make_node(self, op_type, inputs, attr=None, output_count=1, outputs=None, sk
...
+            dtype = dtypes[output]
+            shape = shapes[output]
+            if dtype == TensorProto.UNDEFINED:
+                log.debug("Inferred dtype for [%s, type: %s] is None", node.name, node.type)
+            else:
+                existing_dtype = self.get_dtype(output)
+                if existing_dtype is not None and not override:
+                    dtype = existing_dtype
+                self.set_dtype(output, dtype)
+                log.debug("Set dtype of [%s] to %s", output, dtype)
```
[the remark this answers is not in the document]
lucienwang1009: class methods have only one new line inserted.

tf2onnx/utils.py
```diff
@@ -347,3 +347,39 @@ def tf_name_scope(name):
...
     return [-1 for i in enumerate(shape)]
+
+
+def infer_onnx_shape_dtype(node, input_shapes, input_dtypes, opset=8):
+    """infer shapes and dtypes for outputs of the node"""
+    shapes = {}
+    dtypes = {}
+    inputs = []
+    outputs = []
+    for inp, shape, dtype in zip(node.input, input_shapes, input_dtypes):
```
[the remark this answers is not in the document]
lucienwang1009: Yes, you're right. ONNX infer_shape can work with None shape and `UNDEFINED` dtype as inputs.
I remove the requirement that shape and dtype of inputs cannot be None, and set unknown dtype of `tensor_info` to `UNDEFINED`.

pengwa:
as we talked, I will look at once the set_nodes() is complete (if everybody agrees on that PR)

tf2onnx/utils.py
```diff
@@ -437,3 +448,52 @@ def is_onnx_domain(domain):
...
+            copied_sub_graph = copy.deepcopy(sub_graph)
+            graph_proto = copied_sub_graph.make_graph("graph for " + node.name + " " + attr_name)
+            attr.append(helper.make_attribute(attr_name, graph_proto))
+    attr.extend([a for a in node.attr_onnx.values()])
+    if attr:
+        onnx_node.attribute.extend(attr)
+    return onnx_node
+
+
+def infer_onnx_shape_dtype(node, input_shapes, input_dtypes, opset=constants.PREFERRED_OPSET):
```
nbcsm: let's make opset required.

tf2onnx/utils.py
```diff
@@ -437,3 +448,52 @@ def is_onnx_domain(domain):
     if domain is None or domain == "":
         return True
     return False
+
+
+def build_onnx_op(node):
```
nbcsm: make it an inline function of infer_onnx_shape_dtype

tf2onnx/graph.py
```diff
@@ -535,6 +541,50 @@ def reset_nodes(self, ops):
         self._dtypes = remained_dtypes
         self._output_shapes = remained_shapes
 
+    def update_node_shape_dtype(self, node, override=True):
+        """try the best to infer shapes and dtypes for outputs of the node"""
```
nbcsm: skip if node is not in onnx domain.

tf2onnx/utils.py
nbcsm: i am considering to move this to tf2onnx/schemas.py?
put these onnx def/schema related api calls together.

tf2onnx/utils.py
```diff
@@ -437,3 +448,52 @@ def is_onnx_domain(domain):
...
+    """Build onnx op"""
+    onnx_node = helper.make_node(node.type, node.input, node.output, name=node.name)
+    # deal with attributes
+    attr = []
+    attr_graphs = node.get_body_graphs()
+    if attr_graphs:
```
nbcsm: i am curious about whether we can infer node with sub graph correctly.
you may want to add tests, e.g. a simple If/Loop/Scan?

tf2onnx/graph.py
```diff
@@ -535,6 +541,50 @@ def reset_nodes(self, ops):
...
+
+        for output in node.output:
+            dtype = dtypes[output]
+            shape = shapes[output]
+            if dtype == TensorProto.UNDEFINED:
+                log.debug("Inferred dtype for [%s, type: %s] is None", node.name, node.type)
+            else:
+                existing_dtype = self.get_dtype(output)
+                if existing_dtype is not None:
+                    if override:
```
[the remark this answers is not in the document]
nbcsm: warning or just debug?
we already have too many warnings...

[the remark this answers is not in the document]
lucienwang1009: no worries. only see few in unittests.

tf2onnx/graph.py
```diff
@@ -535,6 +541,50 @@ def reset_nodes(self, ops):
         self._dtypes = remained_dtypes
         self._output_shapes = remained_shapes
 
+    def update_node_shape_dtype(self, node, override=True):
```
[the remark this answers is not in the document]
lucienwang1009: em.. currently not use this option.

lucienwang1009:
To make shape inference work, when creating new nodes, one needs to:
1. make sure the node's inputs are already in the graph.
2. if the node refers to outputs in parent graph, make sure the parent graph is set by `g.parent_graph = parent_graph`.

lucienwang1009:
> please also add tests

added some tests stolen from ONNX: https://github.com/onnx/onnx/blob/master/onnx/test/shape_inference_test.py. :)

tests/test_shape_inference.py
```diff
@@ -0,0 +1,426 @@
...
+            self.logger.error(ex)
+            raised = True
+        self.assertFalse(raised)
+
+        raised = False
+        for actual, inferred in zip(results, outputs):
+            actual_shape = actual.shape
+            inferred_shape = tuple(graph.get_shape(inferred))
+            try:
+                utils.merge_shapes(actual_shape, inferred_shape)
```
nbcsm: are_shapes_compatible? so no need to try/catch

tests/test_shape_inference.py
```diff
@@ -0,0 +1,426 @@
...
+    """
+
+    def _run_test_case(self, graph, feed_dict):
+        """Run model with onnxruntime and compare results' shape with internal shape inference."""
+        outputs = graph.outputs
+        results = []
+        raised = False
+        try:
+            results = self.run_backend(graph, outputs, feed_dict)
+        except Exception as ex:  # pylint: disable=broad-except
```
nbcsm: no need to try/catch, just let it fail?

tf2onnx/graph.py
```diff
@@ -419,7 +421,7 @@ def make_const(self, name, np_val, skip_conversion=False, raw=True):
         return node
 
     def make_node(self, op_type, inputs, attr=None, output_count=1, outputs=None, skip_conversion=True,
-                  op_name_scope=None, name=None, shapes=None, dtypes=None, domain=None):
+                  op_name_scope=None, name=None, shapes=None, dtypes=None, domain=None, auto_infer_shape_dtype=True):
```
nbcsm: auto_infer_shape_dtype => infer_shape_dtype?

tf2onnx/graph.py
```diff
@@ -533,6 +537,68 @@ def reset_nodes(self, ops):
...
+        if node.domain != constants.ONNX_DOMAIN:
+            return
+
+        logger.debug("Infer shape and dtype for [%s]", node.name)
+        # NOTE: shape inference for some ops need the input values of the op, e.g., Reshape
+        # op needs the "Shape" value to infer output shape.
+        initializers = []
+        for i, inp in enumerate(node.inputs):
+            if not inp:
+                if utils.is_debug_mode():
```
nbcsm: logger.isEnabledFor(xxx)?

lucienwang1009: which level we should use? VERBOSE?

nbcsm:
@lucienwang1009, i will merge it later after the random uniform bug is fixed.

PR #332, opened 2019-02-28
Fetch valid attributes from ONNX dynamically

lucienwang1009:
@nbcsm please take a look.


tf2onnx/schema.py
```diff
@@ -0,0 +1,53 @@
...
+from collections import defaultdict, OrderedDict
+from onnx import defs
+
+def version_mapping(schemas):
+    """Obtain version to schema mapping."""
+    schemas_with_version = defaultdict(dict)
+    for s in schemas:
+        schemas_with_version[s.name][int(s.since_version)] = s
+    for name, s in schemas_with_version.items():
+        schemas_with_version[name] = OrderedDict(
```
nbcsm: not 100% sure about python, but for some other languages, modify a dict during traversing may cause issue.
anyway, use a new dict?

tf2onnx/schema.py
```diff
@@ -0,0 +1,53 @@
...
+
+class ONNXSchema:
+    """Wrapper for ONNX schema"""
+
+    all_schemas = defs.get_all_schemas_with_history()
+    schemas_with_version = version_mapping(all_schemas)
+
+    @staticmethod
+    def get_schema(name, version):
+        """Get schema by name within specific version."""
```
nbcsm: version => opset_version, and make the doc string more clear about whether ```version``` is inclusive.

tf2onnx/schema.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class ONNXSchema:
+    """Wrapper for ONNX schema"""
+
+    all_schemas = defs.get_all_schemas_with_history()
+    schemas_with_version = version_mapping(all_schemas)
+
+    @staticmethod
+    def get_schema(name, version):
```
nbcsm: need to take domain into consideration.
we only deal with onnx for now, may have onnx.ml in the future.
see this for reference https://github.com/onnx/onnx/blob/master/onnx/defs/schema.h#L602-L610

tf2onnx/schema.py
```diff
@@ -0,0 +1,53 @@
...
+    for s in schemas:
+        schemas_with_version[s.name][int(s.since_version)] = s
+    for name, s in schemas_with_version.items():
+        schemas_with_version[name] = OrderedDict(
+            sorted(schemas_with_version[name].items(), key=lambda x: x[0])
+        )
+    return schemas_with_version
+
+
+class ONNXSchema:
```
nbcsm: really need a class with all methods static?

tf2onnx/schema.py
```diff
@@ -0,0 +1,53 @@
...
+"""
+
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from collections import defaultdict, OrderedDict
+from onnx import defs
+
+def version_mapping(schemas):
```
nbcsm: make it internal

tf2onnx/graph.py
```diff
@@ -19,6 +19,7 @@
 from tf2onnx import utils, __version__
 from tf2onnx.utils import port_name, find_opset
 from tf2onnx.optimizer.transpose_optimizer import TransposeOptimizer
+from tf2onnx.schema import ONNXSchema
```
nbcsm: schemas?

tf2onnx/schema.py
```diff
@@ -0,0 +1,53 @@
...
+            return None
+        schemas = ONNXSchema.schemas_with_version[name]
+        versions = list(schemas.keys())
+        for i, v in enumerate(versions):
+            if version < v:
+                return schemas[versions[i-1]]
+        return schemas[versions[-1]]
+
+    @staticmethod
+    def get_attribute(name, version):
```
nbcsm: remove it?
a not existing op is an error, should be handled from caller side.

tf2onnx/schema.py
```diff
@@ -0,0 +1,53 @@
...
+    """Obtain version to schema mapping."""
+    schemas_with_version = defaultdict(dict)
+    for s in schemas:
+        schemas_with_version[s.name][int(s.since_version)] = s
+    for name, s in schemas_with_version.items():
+        schemas_with_version[name] = OrderedDict(
+            sorted(schemas_with_version[name].items(), key=lambda x: x[0])
+        )
+    return schemas_with_version
+
```
nbcsm: it might be better to define a schema class, which is a wrapper of onnx schema, may provide some schema level utilities.

lucienwang1009: good point!

tf2onnx/schemas.py
```diff
@@ -0,0 +1,79 @@
...
+from __future__ import unicode_literals
+
+from collections import defaultdict, OrderedDict
+from onnx import defs
+
+
+ONNX_DOMAIN = ""
+
+
+class ONNXOpSchema(object):
```
nbcsm: OnnxOpSchema feels better.

tf2onnx/schemas.py
```diff
@@ -0,0 +1,79 @@
...
+        Args:
+            name (str): op name
+            attributes (List[str]): valid attributes
+            domain (str): default value "" means it's ONNX domain
+            since_version (int): opset version, default is 1
+        """
+        self._name = name
+        self._domain = domain
+        self._attributes = attributes
+        self._since_version = since_version
```
nbcsm: let's expose name, domain, and since version as property, although they are not in use for now.

tf2onnx/schemas.py
```diff
@@ -0,0 +1,79 @@
...
+                    version_schema_map.items(),
+                    key=lambda x: x[0]
+                )
+            )
+    return ordered_version_map
+
+
+all_schemas = defs.get_all_schemas_with_history()
+# format is <OpName, <Domain, <SinceVersion, OpSchema>>>
+schemas = map_schemas(all_schemas)
```
nbcsm: schemas should be internal, so is map_schemas

tf2onnx/schemas.py
```diff
@@ -0,0 +1,79 @@
...
+        self._domain = domain
+        self._attributes = attributes
+        self._since_version = since_version
+
+    @property
+    def attributes(self):
+        return self._attributes
+
+
+def map_schemas(onnx_schemas):
```
nbcsm: register_schemas?

tf2onnx/schemas.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        self._name = name
+        self._domain = domain
+        self._attributes = attributes
+        self._since_version = since_version
+
+    @property
+    def attributes(self):
+        return self._attributes
+
```
nbcsm: better to add a static method, something like ```parse```, which parse 1 of the schema that get from ```defs.get_all_schemas_with_history()``` into an ONNXOpSchema and return.

keep the logic in this class, so in case we need parse other fields, e.g. type constraints, just change here.

tf2onnx/schemas.py
```diff
@@ -0,0 +1,79 @@
...
+            ordered_version_map[name][domain] = OrderedDict(
+                sorted(
+                    version_schema_map.items(),
+                    key=lambda x: x[0]
+                )
+            )
+    return ordered_version_map
+
+
+all_schemas = defs.get_all_schemas_with_history()
```
nbcsm: move into map_schemas?

tf2onnx/schemas.py
```diff
@@ -0,0 +1,79 @@
...
+    """Map schema in ONNX to tf2onnx.ONNXOpSchema"""
+    name_domain_version_schema_map = defaultdict(lambda: defaultdict(dict))
+    for s in onnx_schemas:
+        name = s.name
+        domain = s.domain
+        since_version = int(s.since_version)
+        name_domain_version_schema_map[name][domain][since_version] = \
+            ONNXOpSchema(name, domain, since_version, s.attributes)
+
+    ordered_version_map = defaultdict(lambda: defaultdict(OrderedDict))
```
nbcsm: hmm, don't feel good about ```defaultdict``` here..
when someone uses schema mappings, lookup with a wrong key (either op name, domain, or version) is an error.
while using defaultdict might hide this error.

tf2onnx/schemas.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def get_schema(name, max_inclusive_opset_version, domain=ONNX_DOMAIN):
+    """Get schema by name within specific version."""
+    if name not in schemas:
+        return None
+    if max_inclusive_opset_version < 1:
+        return None
+    domain_version_schema_map = schemas[name]
+    version_schema_map = domain_version_schema_map[domain]
```
nbcsm: what if domain does not exist?

tf2onnx/schemas.py
```diff
@@ -0,0 +1,79 @@
...
+    versions = list(version_schema_map.keys())
+    for i, v in enumerate(versions):
+        if max_inclusive_opset_version < v:
```
nbcsm: you may want to store version from high to low, so if version <= max_inclusive_opset_version, return.
can make the code here simpler.

tf2onnx/schemas.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+all_schemas = defs.get_all_schemas_with_history()
+# format is <OpName, <Domain, <SinceVersion, OpSchema>>>
+schemas = map_schemas(all_schemas)
+
+
+def get_schema(name, max_inclusive_opset_version, domain=ONNX_DOMAIN):
+    """Get schema by name within specific version."""
+    if name not in schemas:
```
nbcsm: no need to check, you will lookup for name anyway.

tf2onnx/graph.py
```diff
@@ -73,8 +74,10 @@ def attr(self):
     def attr_onnx(self):
         onnx_attrs = {}
         for a in self._attr.values():
-            if a.name in utils.ONNX_VALID_ATTRIBUTES:
-                onnx_attrs[a.name] = a
+            schema = get_schema(self.type, self.graph.opset)
+            if schema:
+                if a.name in schema.attributes:
```
nbcsm: schema.has_attribute(name) or schema.contains_attribute(name)?
in case, we may change attributes type (e.g. a dict with name->type constraint?) in the future.

.gitignore
```diff
@@ -1,4 +1,4 @@
-.coverage
+.coverage*
```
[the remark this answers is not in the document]
lucienwang1009: running python setup.py will generate .coverage.xxx files. it's annoying to delete it before push a new commit.

Issue #201, opened 2018-11-20
From tf frozen graph to onnx tf2onnx.convert failed to export padding size in correct order


stonexjr:
OK, I tried to make it short. 
To recreate the issue, do as follows:
1. Download the public pre-trained Resnet-50 v2 model from tensorflow git [repo](https://github.com/tensorflow/models/tree/master/official/resnet), select the one with ResNet-50 v2, fp32 , savedModel [NHWC](http://download.tensorflow.org/models/official/20181001_resnet/savedmodels/resnet_v2_fp32_savedmodel_NHWC.tar.gz)
2. Use `tensorflow.python.tools.freeze_graph` to export a frozen graph from savedModel (see instruction [here](https://github.com/tensorflow/models/tree/master/research/slim#freezing-the-exported-graph)).  Here is the command I used to create frozen graph from savedModel
```bash
bazel-bin/tensorflow/python/tools/freeze_graph \
--input_saved_model_dir="/root/models/resnet50/tensorflow/resnet_v2_fp32_savedmodel_NHWC/1538687283" \ 
--output_graph="/root/models/resnet50/tensorflow/frozen_graph/resnet50_v2_frozen_graph.pb" \
--output_node_names="softmax_tensor"
```
For your convenience here is my pregenerated [frozen graph](https://drive.google.com/open?id=1Yqml6YIoNe5g55isfkSBgIqR4exexxmP)

3. Now use tf2onnx.convert to convert the frozen graph from step (2) to onnx format. 
Here is the command I used to create onnx from frozen graph
```bash
INPUT_FROZEN_GRAPH_PB="/root/models/resnet50/tensorflow/frozen_graph/resnet50_v2_frozen_graph.pb"
OUTPUT_ONNX_FILE="/root/models/resnet50/onnx/resnet50_v2.onnx"
INPUTS="input_tensor:0"
OUTPUTS="softmax_tensor:0"
# --inputs-as-nchw $INPUTS # It doesn't make any difference with this line regarding this issue.
python3 -m tf2onnx.convert \
 --input  $INPUT_FROZEN_GRAPH_PB \
 --inputs $INPUTS \
 --outputs $OUTPUTS \
 --output $OUTPUT_ONNX_FILE \
 --opset 7 \
 --verbose
```
For your convenience here is my pregenerated [onnx](https://drive.google.com/open?id=1GGi1E7Kz8Tw3subge3xeWYxxW-J9SxW6).

[...]

Issue #103, opened 2018-08-07
tensorflow Pad operator


dmoham1476:
Is there an option in tf2onnx to fuse the Pad operator into the Conv? The caffe2->onnx model has no operators and instead has a conv layer with padding, in tensorflow and the tf2onnx model, there is a separate pad operator followed by conv and the conv operator does not have a paddings input (unlike caffe)

guschmue:
we follow the tensorflow graph: if the conv op has padding we map it to onnx and pass it to the conv op.
If there is a seperate pad op after the conv we'll add the same to onnx.
I assume you are asking if we can fuse the pad op into the previous conv op?
Hm, yes - might be doable. Need to look into details.

guschmue:
On second thought: a tf.pad() after tf.conv2d() would pad the output while onnx conv() pads=[...] would pad the input. So maybe if you have something like conv1->pad->conv2->pad->conv3 you could fuse the pad into the following conv but I'd expect model models will do this already in the source code ... have not seen many models that call tf.pad()

dmoham1476:
Hi, I see that the pad operator is before the convolution. (pad->conv), I see this pattern in resnet50 tensorflow graph in three places.

guschmue:
I can take a look, maybe there is a way of fusing this into the pads for conv()

tlepley-cadence:
We've also got a  different TF models with this Pad node before Conv2D. We had to handle it after tf2onnx. But if tf2onnx does the merge of Pad with Conv2D, I think this makes the generated ONNX model nicer.

pengwa:
this should be fixed in https://github.com/onnx/tensorflow-onnx/pull/334 by @lucienwang1009 . please let us know if you still have issue on this.

PR #334, opened 2019-02-28
Fuse Pad op before Conv2D op into Conv op as an attribute

lucienwang1009:
@guschmue For resnet-50, three Pads are fused into Conv.

See also #103 #201

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -383,6 +383,7 @@ def conv_convert_inputs(ctx, node, with_kernel=False, new_kernel_shape=None,
                 transpose = ctx.insert_new_node_on_input(node, "Transpose", input_name)
                 transpose.set_attr("perm", NHWC_TO_NCHW)
                 transpose.inserted_nchw = True
+                transpose.skip_conversion = True
                 shape = ctx.get_shape(input_name)
                 new_shape = spatial_map(shape, NHWC_TO_NCHW)
                 ctx.set_shape(transpose.output[0], new_shape)
@@ -399,6 +400,7 @@ def conv_convert_inputs(ctx, node, with_kernel=False, new_kernel_shape=None,
         transpose = ctx.insert_new_node_on_input(node, "Transpose", input_name)
         transpose.set_attr("perm", HWCN_TO_NCHW)
         transpose.inserted_nchw = True
+        transpose.skip_conversion = True
         ctx.copy_shape(input_name, transpose.output[0])
         new_shape = spatial_map(ctx.get_shape(input_name), HWCN_TO_NCHW)
         ctx.set_shape(transpose.output[0], new_shape)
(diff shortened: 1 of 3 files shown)


tf2onnx/tfonnx.py
```diff
@@ -2280,6 +2285,47 @@ def rewrite_incomplete_type_support_rs6(g, ops):
...
+        OpTypePattern("Conv2D", name="conv", inputs=[
+            OpTypePattern("Pad", name="pad"),
+            OpTypePattern("*")
+        ])
+    matcher = GraphMatcher(pattern)
+    match_results = list(matcher.match_ops(ops))
+    for match in match_results:
+        conv = match.get_op("conv")
+        pad = match.get_op("pad")
+        paddings = pad.inputs[1]
```
pengwa: any bad impact if padding node is used more than once?

tf2onnx/tfonnx.py
```diff
@@ -2280,6 +2285,47 @@ def rewrite_incomplete_type_support_rs6(g, ops):
...
+
+        log.debug("merge pad [%s] into conv [%s]", pad.name, conv.name)
+        paddings_val = np.array(paddings.get_tensor_value())
+        # can't pad on batch or channel dimensions
+        if np.any(paddings_val[0]) or np.any(paddings_val[3]):
+            return ops
+        paddings_val = paddings_val[1:3]
+        paddings_val = paddings_val.transpose().flatten()
+        g.replace_input(conv, conv.input[0], pad.input[0])
+        # convert Conv2D
```
pengwa: if you want to do conversion here, I think you should also consider opset check.

pengwa: shall we consider to do the fusion in the later rewriter? then we won't have the trick to call mapping function in this stage.

pengwa: or it can even be part of the optimizer.

tf2onnx/tfonnx.py
```diff
@@ -2280,6 +2285,47 @@ def rewrite_incomplete_type_support_rs6(g, ops):
...
+        ])
+    matcher = GraphMatcher(pattern)
+    match_results = list(matcher.match_ops(ops))
+    for match in match_results:
+        conv = match.get_op("conv")
+        pad = match.get_op("pad")
+        paddings = pad.inputs[1]
+
+        if not paddings.is_const():
+            return ops
```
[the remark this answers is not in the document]
lucienwang1009: yes, my fault.

tf2onnx/tfonnx.py
lucienwang1009: Here, I keep the Pad op intact and change the input of following Conv2D by `replace_input`. It won't affect other nodes using this Pad op.

tf2onnx/tfonnx.py
lucienwang1009: Have thought of this. but then we will need to handle the case in which converted Pad and Conv nodes are separated by Cast and Transpose nodes inserted by the converter.

lucienwang1009: > if you want to do conversion here, I think you should also consider opset check.

Talk with @pengwa offline. We decide not to check OPSET here, as `Conv` op has no change since OPSET 1.
But we find many pre-rewriters create ONNX nodes without doing this check. It will screw up if we use nodes of the wrong version in these rewriters.

guschmue: Not checking sounds correct. I guess if in a future opset something changes we'll depend on unittests to remind us.

PR #335, opened 2019-02-28
add tf_graph_tool

nbcsm:
try to centralize **common** tf graph operations into a dedicated script.
this is an initial version, some are borrowed from existing tool scripts.

### DIFF
--- a/tools/tf_graph_tool.py
+++ b/tools/tf_graph_tool.py
@@ -1,0 +1,198 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT license.
+
+""" Tool for common tf graph operations. """
+
+from __future__ import division
+from __future__ import print_function
+
+import argparse
+import logging
+import os
+import sys
+
+from google.protobuf import text_format
+import tensorflow as tf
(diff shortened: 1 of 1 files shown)


tools/tf_graph_tool.py
```diff
@@ -0,0 +1,198 @@
...
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers()
+
+    # pb2txt
+    subparser = subparsers.add_parser("pb2txt", help="convert pb to pbtxt")
+    subparser.add_argument("--input", dest="input_path", required=True, help="input pb path")
+    subparser.add_argument("--output", dest="output_path", help="output pbtxt path")
+    subparser.set_defaults(func=main.convert_pb_to_pbtxt)
+
+    # pb2summary
```
pengwa: add some example for using this?

nbcsm: sometimes, it is easier to view large graph using tensorboard instead of netron.
netron cannot navigate among nodes by edges easily...

nbcsm: i can add something to MD or wiki later.

PR #337, opened 2019-03-01
rewrite the graph and body graph recursively

lucienwang1009:
The prior converter cannot rewrite the subgraph in nodes. In this PR, the rewriters will be performed on the original graph and nodes' subgraph generated by converter recursively. 

Other changes:
1. set the shape and dtype for `If` node in `cond_rewriter`.
2. fix a bug in `make_node` when `outputs` is given while `output_count` is not set.
3. enable some unit tests, in which `tf.cond` is inside `tf.while_loop` and `tf.cond` contains `tf.while_loop`.


tf2onnx/tfonnx.py
```diff
@@ -2486,19 +2508,7 @@ def process_tf_graph(tf_graph, continue_on_error=False, verbose=False, target=No
...
-            g.set_nodes(ops)
-    except Exception as ex:
-        type_, value_, traceback_ = sys.exc_info()
-        log.error("node %s: exception %s" % (rewrite, ex))
-        ex_ext = traceback.format_exception(type_, value_, traceback_)
-        if continue_on_error:
-            log.info(ex_ext)
-        else:
-            raise ex
+    tensorflow_onnx_rewrite(g, rewriters)
```
nbcsm: there is a run_late_rewriters, which are doing similar things, can we re-use? 
@pengwa for more comments.

nbcsm: so the question is does the graph processing order matter?
rewrite parent then child vs. rewrite child then parent?

tf2onnx/rewriter/cond_rewriter.py
```diff
@@ -104,28 +104,63 @@ def run(self):
 
         return self.g.get_nodes()
 
+    def _get_output_shape_dtype(self, cond_context):
```
pengwa: I actually refactored this a bit in https://github.com/onnx/tensorflow-onnx/pull/340. But you can go ahead with this, I'll merge it.

tf2onnx/tfonnx.py
```diff
@@ -2353,6 +2353,19 @@ def tensorflow_onnx_mapping(g, continue_on_error, custom_op_handlers):
     return mapped_op, unmapped_op
 
 
+def run_pre_rewriters(g, rewriters):
```
pengwa: shall we just use run_late_rewriters, while rename it to like run_rewriters()?

tf2onnx/tfonnx.py
```diff
@@ -2475,17 +2475,18 @@ def topological_sort(g, continue_on_error):
...
-def run_late_rewriters(g, funcs, continue_on_error):
-    if g.contained_graphs:
-        for dict_val in g.contained_graphs.values():
-            for attr_name, b_g in dict_val.items():
-                run_late_rewriters(b_g, funcs, attr_name)
-
-    topological_sort(g, continue_on_error)
+def run_rewriters(g, funcs, continue_on_error=False, need_sort=True):
+    if need_sort:
+        topological_sort(g, continue_on_error)
```
nbcsm: as discussed, let's move topo sort out.

tf2onnx/tfonnx.py
```diff
@@ -2475,16 +2475,26 @@ def topological_sort(g, continue_on_error):
             pass
 
 
-def run_late_rewriters(g, funcs, continue_on_error):
+def run_rewriters(g, funcs, continue_on_error):
+    """Rewrite the original graph and body graphs of nodes"""
+    # NOTE: not topologically sort graph here
```
nbcsm: please add more words here about:
- rewriter is expected to sort nodes on its own
- graph may contains circle, current topo sort cannot handle it.
- why process parent graph first then child graphs.

lucienwang1009: added some comments, the third question is obvious if one reads the code.

PR #338, opened 2019-03-01
fix bug found when option "custom-ops" is used

zhijxu-MS:


nbcsm:
this issue is caused by https://github.com/onnx/tensorflow-onnx/pull/305

will add test to cover this.

PR #340, opened 2019-03-03
Graph node management in real time when add/delete

pengwa:
Originally, Graph manage the nodes this way: 

- When we create a node, we actually won't add it into the Graph's _nodes directly, we will need retrieve all existing nodes using Graph's get_nodes() from current Graph, into a list, and append the new node to it, call Graph's set_nodes(ops) to reset the Graph.

- When we delete a node, we call Graph's get_nodes() to get all existing nodes into a list, then remove those that we want to remove from this list, and call Graph's set_nodes(ops) to update the nodes. 

With this change:

-  All actions to Graph including adding node/removing node, it will be reflected immediately on Graph's _nodes property. 

- This would help us to have simpler logic to manage nodes.


pengwa:
testing matrix:

python tests\test_backend.py

python tests\test_lstm.py

python tests\test_gru.py

python tests\test_grublock.py

python tests\test_loops.py --opset 8

python tests\test_graph.py

python tests\test_optimizers.py

python tests\test_custom_rnncell.py --opset 8

python tests\test_internals.py

python tests\run_pretrained_models.py --backend onnxruntime --config tests\run_pretrained_models.yaml --fold_const --onnx-file ".\tmp" --opset 8

python tests\run_pretrained_models.py --backend onnxruntime --config tests\internal-rnn.yaml --fold_const --onnx-file ".\tmp" --opset 8 --target rs6

python tests\run_pretrained_models.py --backend onnxruntime --config tests\internal-cnn.yaml --fold_const --onnx-file ".\tmp" --opset 8 --target rs6 

python -m tf2onnx.convert --input tests\internal\cnns\classical\resnet_v2_152_inf_graph.pb --inputs input:0 --outputs resnet_v2_152/predictions/Reshape_1:0 --output res152.onnx --verbose --fold_const

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -73,13 +73,12 @@ def pre_optimize_action(self):
...
-            ops.append(self._g.make_const(const_name, new_data))
+            self._g.make_const(const_name, new_data)
 
             # need call this to make input update synced to protobuf val
             self._g.update_proto()
-            ops.remove(reshape_op)
-            self._g.set_nodes(ops)
-            self._g.topological_sort(ops)
+            self._g.remove_node(reshape_op.name)
+            self._g.topological_sort(self._g.get_nodes())
```
nbcsm: with this change, still need the ```ops``` as argument?
at least it can be optional.

tf2onnx/tfonnx.py
```diff
@@ -2484,7 +2384,7 @@ def run_late_rewriters(g, funcs, continue_on_error):
     topological_sort(g, continue_on_error)
     for func in funcs:
         ops = func(g, g.get_nodes())
```
nbcsm: rewriter interface still needs ```ops``` argument?

tf2onnx/optimizer/transpose_optimizer.py
pengwa: yeah, we might not need them now. let's do it separately next time.

tf2onnx/tfonnx.py
nbcsm: seems rewriter does not need ```ops``` either

guschmue:
hm, I'm not sure I like working directly on the node list of the graph ... it breaks the immutable nature of
```
for rewriter in rewriters:
   ops = rewriter(ops)
```
Let me see if I can convince myself that operating on on the nodes in the graph in a mutable way is not avoidable.

pengwa:
@guschmue for this specific case, yes, it is. with this change, actually we just need pass rewriter a Graph (which I did not change in this PR), that's enough,  rewriter  as a modifier of a graph.

guschmue:
yeah, I know, we are already not doing the immutable rewrite in some cases ... was just hoping we could get back to immutable  :)

pengwa:
@guschmue thanks for the feedback. Are you mainly concerning about rewriters only? Or you also concerns about the mutable node update during op mapping stage? 

If it is the first case, I think we can still use older way, e.g. maintain a node list during rewriter, and finally reset the Graph with nodes.

pengwa:
Or we have a dicussion during sync-up.

guschmue:
yeah, lets discuss tomorrow.

pengwa:
thanks for the reviews. as we discussed together, this change is against original immutable design. while we did have few reasons doing this: 

1. during op mapping, some newly added nodes not reflected in Graph, so if subsequent op mapping need update those nodes' property, they won't be retrieved from the Graph.  Recently we just  have such a case https://github.com/onnx/tensorflow-onnx/issues/356. 

2. everytime we operate the graph, it's a bit heavy to manage a node list (either add or remove), as you can see in rewriters. 

So we will let this change in, and see whether it is good enough. If not, let do clean up.

Another thing to mention:

There are few things that need to be improved to make the new design better to understand, for example, make the failure happened earlier in case it's hard to know where we fail since we are operating the graph in real time. I'll try to mitigate with few minor changes later.

PR #341, opened 2019-03-03
fix const transpose

guschmue:


tf2onnx/tfonnx.py
```diff
@@ -393,19 +391,30 @@ def conv_convert_inputs(ctx, node, with_kernel=False, new_kernel_shape=None,
...
-        transpose.set_attr("perm", HWCN_TO_NCHW)
-        transpose.inserted_nchw = True
-        transpose.skip_conversion = True
-        ctx.copy_shape(input_name, transpose.output[0])
-        new_shape = spatial_map(ctx.get_shape(input_name), HWCN_TO_NCHW)
-        ctx.set_shape(transpose.output[0], new_shape)
-        nodes.append(transpose)
-        parent.data_format = "NCHW"
+
+        need_transpose = True
```
pengwa: why not use if- else, instead of need_transpose?

tf2onnx/tfonnx.py
```diff
@@ -393,19 +391,30 @@ def conv_convert_inputs(ctx, node, with_kernel=False, new_kernel_shape=None,
...
+        if node.inputs[1].is_const():
+            # kernel is const - transpose the const if we are the only consumer of const
+            # TODO: maybe we should make a copy of the const, or look at the other consumers
+            # if they'd want a transose as well.
```
pengwa: a typo - transose

tf2onnx/tfonnx.py
guschmue: ah, because there are 2 conditions ... if cons and if consumers==1

pengwa: sorry...misunderstood

tf2onnx/tfonnx.py
guschmue: ah, will fix with next commit

nbcsm:
@guschmue, seems the CLA is not signed?

PR #342, opened 2019-03-04
enhance transpose optimizer

zhijxu-MS:


zhijxu-MS:
> please add test to cover this case and #326

related tests has been added

tests/test_optimizers.py
```diff
@@ -104,6 +104,38 @@ def test_max(self):
         self.run_and_compare(["Z1"], {"X": np.random.randn(2, 3, 4, 5).astype(np.float32)},
                              model_proto)
 
+    def test_transpose_merge(self):
+        node0 = helper.make_node("Transpose", ["X"], ["Y"], perm=[0, 2, 3, 1], name="trans")
+        node1 = helper.make_node("Transpose", ["X"], ["Y_1"], perm=[0, 2, 3, 1], name="trans_1")
+        node2 = helper.make_node("Mul", ["Y", "Y_1"], ["Z"], name="mul")
+        node3 = helper.make_node("Transpose", ["Z"], ["OUT"], perm=[0, 3, 1, 2], name="trans_2")
+
```
nbcsm: currently, we only compare whether there are LESS transpose ops.:
```self.assertTrue(current["Transpose"] < previous["Transpose"], msg="transpose ops count not changed")```

this is too relaxed, we should check whether left transpose count **equals to an expected value**.

zhijxu-MS: done.

nbcsm:
please rebase onto latest mater to get the fix for pretrained model test.

PR #343, opened 2019-03-05
fix cond_test bug

lucienwang1009:


nbcsm:
@lucienwang1009, i will merge this PR to make CI happy.
but it is better to enhance shape inference side to get it fixed.

PR #346, opened 2019-03-06
optimize upsample's conversion logic

zhijxu-MS:
input shape info is known only when input's shape is fixed. Under this
case the scale can be a const instead of a subgraph to compute it when
running the model.

the reason why I change the way to create initializers is because sometime we want to put data at related field of proto object such as float_data instead of raw_data(such as onnx sometimes will only access float_data when dtype of proto obj is specified)

import a mechanism to check the generated onnx graph is expected.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -92,7 +92,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
     def _run_test_case(self, output_names_with_port, feed_dict, **kwargs):
         kwargs["convert_var_to_const"] = False
         kwargs["constant_fold"] = False
-        self.run_test_case(feed_dict, [], output_names_with_port, **kwargs)
+        return self.run_test_case(feed_dict, [], output_names_with_port, **kwargs)
 
     def _test_expand_dims(self, idx):
         tf.reset_default_graph()
@@ -1276,7 +1276,10 @@ class BackendTests(Tf2OnnxBackendTestBase):
         x_new_size_ = tf.constant(x_new_size)
         x_ = tf.image.resize_nearest_neighbor(x, x_new_size_)
         _ = tf.identity(x_, name=_TFOUTPUT)
-        self._run_test_case([_OUTPUT], {_INPUT: x_val})
+        graph = self._run_test_case([_OUTPUT], {_INPUT: x_val})
+        if self.config.opset >= 9:
+            scale_node = group_nodes_by_type(graph)["Upsample"][0].inputs[1]
+            self.assertTrue(validate_const_node(scale_node, [1.0, 1.0, 2.0, 2.0]))
 
     @check_opset_min_version(9, "resize_nearest_neighbor")
     def test_resize_nearest_neighbor_with_non_const(self):
@@ -1301,7 +1304,10 @@ class BackendTests(Tf2OnnxBackendTestBase):
         x_new_size_ = tf.constant(x_new_size)
         x_ = tf.image.resize_bilinear(x, x_new_size_)
         _ = tf.identity(x_, name=_TFOUTPUT)
-        self._run_test_case([_OUTPUT], {_INPUT: x_val})
+        graph = self._run_test_case([_OUTPUT], {_INPUT: x_val})
+        if self.config.opset >= 9:
+            scale_node = group_nodes_by_type(graph)["Upsample"][0].inputs[1]
+            self.assertTrue(validate_const_node(scale_node, [1.0, 1.0, 2.0, 2.0]))
 
     @check_opset_min_version(9, "resize_bilinear")
     def test_resize_bilinear_with_non_const(self):
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -373,14 +373,19 @@ class Graph(object):
         """Return True if target platform contains any name."""
         return any(name in self._target for name in names)
 
-    def make_const(self, name, np_val, skip_conversion=False):
+    def make_const(self, name, np_val, skip_conversion=False, raw=True):
         """Make a new constant in the graph.
         Args:
             name: const node name, must be unique.
             np_val: value of type numpy ndarray.
             skip_conversion: bool, indicate whether this created node would be mapped during conversion.
+            raw: whether to store data at field of raw_data or the specific field according to its dtype
         """
-        onnx_tensor = numpy_helper.from_array(np_val, name)
+        if raw:
+            onnx_tensor = numpy_helper.from_array(np_val, name)
+        else:
+            onnx_tensor = helper.make_tensor(name, utils.map_numpy_to_onnx_dtype(np_val.dtype),
+                                             np_val.shape, np_val, raw=False)
         node = self.make_node("Const", [], outputs=[name], name=name, attr={"value": onnx_tensor},
                               skip_conversion=skip_conversion)
         self.set_shape(name, np_val.shape)
@@ -732,8 +737,12 @@ class Graph(object):
         # create initializers for constant nodes
         const_ops = [op for op in const_ops if op not in placeholder_default_const_ops]
         for op in const_ops:
-            const_val = op.get_tensor_value(as_list=False)
-            tensor = numpy_helper.from_array(const_val, op.output[0])
+            # not to use numpy_helper.from_array to create a new tensor
+            # because sometimes onnx will have a bug that only check the tensor data in specific field
+            # such as at upsample it only checks the float_data field.
+            t = op.get_attr("value")
+            tensor = helper.get_attribute_value(t)
+            tensor.name = op.output[0]
             initializers.append(tensor)
 
         # create input_tensor_values
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -1056,18 +1056,37 @@ def upsample_op9(ctx, node, name, args):
     # float32 out = ResizeBilinear/ResizeNearestNeighbor(T images, int size)
     # https://www.tensorflow.org/api_docs/python/tf/image/resize_nearest_neighbor
     # wants the input to be NHWC - adjust target_shape to this.
-    ori_shape = ctx.make_node("Shape", [node.input[0]])
-    ori_shape_hw = ctx.make_node("Slice", ori_shape.output, {"axes": [0], "starts": [1], "ends": [3]})
-    ori_shape_hw_float = ctx.make_node("Cast", ori_shape_hw.output, attr={"to": onnx_pb.TensorProto.FLOAT})
 
-    target_hw = node.inputs[1]
-    target_hw_float = ctx.make_node("Cast", target_hw.output, attr={"to": onnx_pb.TensorProto.FLOAT})
+    # first create "scales" info for onnx upsample
+    # if shape of input and output known then  "scale" is calculated statically and set as a const node
+    shape = ctx.get_shape(node.input[0])
+    added_nodes = []
+    if shape and shape[2] != -1 and shape[1] != -1 and node.inputs[1].is_const():
+        target_shape = node.inputs[1].get_tensor_value()
+        n, h, w, c = shape
+        nh, nw = target_shape
+        # scales is nchw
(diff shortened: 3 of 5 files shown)


tests/test_backend.py
```diff
@@ -1726,6 +1742,26 @@ def test_isnan(self):
...
+
+        ops = graph.get_nodes()
+        for node in ops:
+            if checker(graph, node):
+                return True
+        # doesn't find anything
+        return False
+
+    @staticmethod
+    def _is_upsample_with_const_input(graph, node, const_val):
```
nbcsm: how about something like ```validate_const_node(node, value)```
put it in tests/common.py, so other test scripts can use it.

tests/test_backend.py
```diff
@@ -1305,7 +1318,10 @@ def test_resize_bilinear_with_non_const(self):
 
         x_ = tf.image.resize_bilinear(x, x_new_size_)
         _ = tf.identity(x_, name=_TFOUTPUT)
-        self._run_test_case([_OUTPUT], {_INPUT: x_val, _INPUT1: x_new_size})
+        graph = self._run_test_case([_OUTPUT], {_INPUT: x_val, _INPUT1: x_new_size})
+        make_sure(not self._has_specific_node(graph, partial(self._is_upsample_with_const_input,
+                                                             const_val=[1.0, 1.0, 2.0, 2.0])),
+                  "scale shouldn't be a const")
```
nbcsm: may not be necessary?

tests/test_backend.py
```diff
@@ -1280,7 +1286,10 @@ def test_resize_nearest_neighbor_with_non_const(self):
 
         x_ = tf.image.resize_nearest_neighbor(x, x_new_size_)
         _ = tf.identity(x_, name=_TFOUTPUT)
-        self._run_test_case([_OUTPUT], {_INPUT: x_val, _INPUT1: x_new_size})
+        graph = self._run_test_case([_OUTPUT], {_INPUT: x_val, _INPUT1: x_new_size})
+        make_sure(not self._has_specific_node(graph, partial(self._is_upsample_with_const_input,
```
nbcsm: may not be necessary? onnx result won't be equal to tf result anyway.

tests/test_backend.py
```diff
@@ -1726,6 +1742,26 @@ def test_isnan(self):
             self._run_test_case([_OUTPUT], {_INPUT: x_val})
             tf.reset_default_graph()
 
+    @staticmethod
+    def _has_specific_node(graph, checker):
```
nbcsm: how about a ```group_nodes_by_type(g)```  returns a dict of node op type => list[node]?
then user may use the dict to do further validation. e.g. compare transpose count.
or for your case:
```group_nodes_by_type(g)["Upsample"].inputs[1]```

also, put it in tests/common.py, so other test scripts can use it.

File: tf2onnx/tfonnx.py [hunk 5]
```diff
@@ -1148,24 +1148,43 @@ def upsample_op9(ctx, node, name, args):
...
+        scaler_node = ctx.make_node("Concat", [const_one_array.output[0], scales_hw.output[0]], {"axis": 0})
+        added_nodes.extend([ori_shape, ori_shape_hw, ori_shape_hw_float, target_hw_float,
+                            scales_hw, const_one_array, scaler_node])
 
-    const_one_array = ctx.make_const(utils.make_name("one"), np.array([1.0, 1.0]).astype(np.float32))
-    # scaler is nchw
-    scales = ctx.make_node("Concat", [const_one_array.output[0], scales_hw.output[0]], {"axis": 0})
     input_nchw = ctx.make_node("Transpose", [node.input[0]], {"perm": [0, 3, 1, 2]})
-    upsample = ctx.make_node("Upsample", [input_nchw.output[0], scales.output[0]], attr={"mode": args[0]})
+    upsample = ctx.make_node("Upsample", [input_nchw.output[0], scaler_node.output[0]], attr={"mode": args[0]})
```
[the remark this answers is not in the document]
zhijxu-MS: onnxruntime only supports to scale the last two dims

tests/test_backend.py
```diff
@@ -1267,7 +1267,11 @@ def test_resize_nearest_neighbor(self):
         x_new_size_ = tf.constant(x_new_size)
         x_ = tf.image.resize_nearest_neighbor(x, x_new_size_)
         _ = tf.identity(x_, name=_TFOUTPUT)
-        self._run_test_case([_OUTPUT], {_INPUT: x_val})
+        graph = self._run_test_case([_OUTPUT], {_INPUT: x_val})
+        if get_test_config().opset >= 9:
```
nbcsm: btw: ```self.config``` has same effect.

tf2onnx/graph.py
```diff
@@ -664,8 +667,9 @@ def make_graph(self, doc, graph_name="tf2onnx"):
         # create initializers for constant nodes
         const_ops = [op for op in const_ops if op not in placeholder_default_const_ops]
         for op in const_ops:
-            const_val = op.get_tensor_value(as_list=False)
-            tensor = numpy_helper.from_array(const_val, op.output[0])
+            t = op.get_attr("value")
+            tensor = helper.get_attribute_value(t)
```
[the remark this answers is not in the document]
nbcsm: btw: plz open an issue on onnx repo for that shape inference bug.

tf2onnx/tfonnx.py
```diff
@@ -1084,24 +1084,41 @@ def upsample_op9(ctx, node, name, args):
...
+        target_hw = node.inputs[1]
+        target_hw_float = ctx.make_node("Cast", target_hw.output, attr={"to": onnx_pb.TensorProto.FLOAT})
 
-    const_one_array = ctx.make_const(utils.make_name("one"), np.array([1.0, 1.0]).astype(np.float32))
-    # scaler is nchw
-    scales = ctx.make_node("Concat", [const_one_array.output[0], scales_hw.output[0]], {"axis": 0})
+        scales_hw = ctx.make_node("Div", [target_hw_float.output[0], ori_shape_hw_float.output[0]])
+
+        const_one_array = ctx.make_const(utils.make_name("one"), np.array([1.0, 1.0]).astype(np.float32))
+        # scaler is nchw
```
nbcsm: ```scales``` seems a better name, align with attribute name in onnx spec.

tf2onnx/graph.py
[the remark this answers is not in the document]
zhijxu-MS: issue [1852 ](https://github.com/onnx/onnx/issues/1852)is created for the bug.

tf2onnx/graph.py
```diff
@@ -360,14 +360,22 @@ def is_target(self, *names):
...
         """
-        onnx_tensor = numpy_helper.from_array(np_val, name)
+        if raw:
+            onnx_tensor = numpy_helper.from_array(np_val, name)
+        else:
+            # onnx sometimes will have bug that only get data from specific data field, so we have to keep data
+            # at specific field instead of raw_data field
+            tensor_val = np_val.tobytes() if raw else np_val
+            onnx_tensor = helper.make_tensor(name, utils.map_numpy_to_onnx_dtype(np_val.dtype),
+                                             np_val.shape, tensor_val, raw=raw)
```
nbcsm: now the condition for else is already ```raw=False```, should just be 
```
onnx_tensor = helper.make_tensor(name, utils.map_numpy_to_onnx_dtype(np_val.dtype),
                                             np_val.shape, np_val, raw=False)
```

tf2onnx/graph.py
```diff
@@ -360,14 +360,22 @@ def is_target(self, *names):
...
             np_val: value of type numpy ndarray.
             skip_conversion: bool, indicate whether this created node would be mapped during conversion.
+            raw: whether to store data at field of raw_data or the specific field according to its dtype
         """
-        onnx_tensor = numpy_helper.from_array(np_val, name)
+        if raw:
+            onnx_tensor = numpy_helper.from_array(np_val, name)
+        else:
+            # onnx sometimes will have bug that only get data from specific data field, so we have to keep data
+            # at specific field instead of raw_data field
```
nbcsm: just remove these 2 lines

tf2onnx/tfonnx.py
```diff
@@ -1084,24 +1084,41 @@ def upsample_op9(ctx, node, name, args):
...
-    scales = ctx.make_node("Concat", [const_one_array.output[0], scales_hw.output[0]], {"axis": 0})
+        scales_hw = ctx.make_node("Div", [target_hw_float.output[0], ori_shape_hw_float.output[0]])
+
+        const_one_array = ctx.make_const(utils.make_name("one"), np.array([1.0, 1.0]).astype(np.float32))
+        # scales is nchw
+        scales = ctx.make_node("Concat", [const_one_array.output[0], scales_hw.output[0]], {"axis": 0})
+        added_nodes.extend([ori_shape, ori_shape_hw, ori_shape_hw_float, target_hw_float,
+                            scales_hw, const_one_array, scales])
+
+    # because onnxruntime only supports to scale the last two dims so transpose is inserted
```
nbcsm: let's keep it this way for now.
but we need to confirm with Ke whether this behavior is expected by onnx spec or it is another onnxruntime limitation.

if it is onnxruntime limitaton, then we may treat the transpose logic as target specific.

tf2onnx/tfonnx.py
```diff
@@ -1084,24 +1084,41 @@ def upsample_op9(ctx, node, name, args):
...
+    # first create "scales" info for onnx upsample
+    # if shape of input and output known then  "scale" is calculated statically and set as a const node
+    shape = ctx.get_shape(node.input[0])
+    added_nodes = []
+    if shape and shape[2] != -1 and shape[1] != -1 and node.inputs[1].is_const():
+        target_shape = node.inputs[1].get_tensor_value()
+        n, h, w, c = shape
+        nh, nw = target_shape
+        # scales is nchw
+        scale_val = np.array([1.0, 1.0, float(nh) / h, float(nw) / w]).astype(np.float32)
```
nbcsm: we should add a comment here about why raw is false, and put the issue link here.

tf2onnx/graph.py
```diff
@@ -360,14 +360,20 @@ def is_target(self, *names):
...
             np_val: value of type numpy ndarray.
             skip_conversion: bool, indicate whether this created node would be mapped during conversion.
+            raw: whether to store data at field of raw_data or the specific field according to its dtype
         """
-        onnx_tensor = numpy_helper.from_array(np_val, name)
+        if raw:
+            onnx_tensor = numpy_helper.from_array(np_val, name)
+        else:
+            tensor_val = np_val.tobytes() if raw else np_val
+            onnx_tensor = helper.make_tensor(name, utils.map_numpy_to_onnx_dtype(np_val.dtype),
```
nbcsm: now the condition for else is already ```raw=False```, should just be 
```
onnx_tensor = helper.make_tensor(name, utils.map_numpy_to_onnx_dtype(np_val.dtype),
                                             np_val.shape, np_val, raw=False)
```

nbcsm:
please resolve conflict.

nbcsm:
@zhijxu-MS please rebase onto latest mater to get the fix for pretrained model test.

tf2onnx/tfonnx.py
```diff
@@ -1056,18 +1056,37 @@ def upsample_op9(ctx, node, name, args):
...
-    ori_shape = ctx.make_node("Shape", [node.input[0]])
-    ori_shape_hw = ctx.make_node("Slice", ori_shape.output, {"axes": [0], "starts": [1], "ends": [3]})
-    ori_shape_hw_float = ctx.make_node("Cast", ori_shape_hw.output, attr={"to": onnx_pb.TensorProto.FLOAT})
 
-    target_hw = node.inputs[1]
-    target_hw_float = ctx.make_node("Cast", target_hw.output, attr={"to": onnx_pb.TensorProto.FLOAT})
+    # first create "scales" info for onnx upsample
+    # if shape of input and output known then  "scale" is calculated statically and set as a const node
+    shape = ctx.get_shape(node.input[0])
+    added_nodes = []
```
nbcsm: this seems no longer needed?

PR #347, opened 2019-03-07
replace conversion logic of "select" with a simpler one

zhijxu-MS:
also release accuracy check of gru tests

### DIFF
--- a/tf2onnx/function/select.py
+++ b/tf2onnx/function/select.py
@@ -4,236 +4,42 @@
 """
 tf2onnx.tf2onnx - select op conversion
 """
-import numpy as np
-from onnx.onnx_pb import TensorProto
+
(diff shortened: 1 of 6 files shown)


tf2onnx/function/select.py
```diff
@@ -4,236 +4,43 @@
...
-from tf2onnx.utils import port_name, make_sure
 
 
 # pylint: disable=unused-argument,missing-docstring
 
 
-def select_op8(ctx, node, name, args):
+def select_op7(ctx, node, name, args):
     # T output = Select(bool condition, T x, T y)
     # V v_final_and_scan_outputs = Loop(int64 M, B cond, V v_initial)
```
pengwa: change the comment pls

PR #350, opened 2019-03-07
shape inference support LogicalOr, relax shape check for cond rewriter

nbcsm:
fix several issues found in real model conversion.

- shape inference support LogicalOr
- relax cond_rewriter shape check
it is possible that shapes for true and false branches are **compatible but not exactly the same**, which is still valid.
- add shape utilities to simplify shape operations.

### DIFF
--- a/tf2onnx/utils.py
+++ b/tf2onnx/utils.py
@@ -342,11 +342,6 @@ def tf_name_scope(name):
     return '/'.join(name.split('/')[:-1])
 
 
-def create_vague_shape_like(shape):
-    make_sure(len(shape) >= 0, "rank should be >= 0")
-    return [-1 for i in enumerate(shape)]
-
-
 def get_temp_directory():
     return os.environ.get("TF2ONNX_TEMP_DIRECTORY", tempfile.mkdtemp())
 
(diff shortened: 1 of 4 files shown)


nbcsm:
@lucienwang1009 to take a look at cond rewriter.

lucienwang1009:
@nbcsm Cool!

tf2onnx/utils.py
```diff
@@ -364,3 +359,70 @@ def save_protobuf(path, message, as_text=False):
...
+    None is compatible with all shapes, different ranks are not considered as compatible
+    """
+    try:
+        merge_shapes(src, dest)
+        return True
+    except Exception:
+        return False
+
+
+def are_shapes_equal(src, dest):
```
pengwa: I guess we should avoid use this check in product code.

nbcsm: i feel this method is useful when user TRIES to merge 2 shapes, e.g. in cond rewriter.
it is easier to write code like
```
if are_shapes_compatible(s1, s2):
	s1 = merge_shapes(s1, s2)
```
otherwise, user may need to try/catch on merge_shapes just in case there is conflict.

tf itself has similar api: https://www.tensorflow.org/api_docs/python/tf/TensorShape#is_compatible_with

nbcsm: not quite sure, may have some scenarios that need to check whether 2 shapes need to be **exactly the same**.

nbcsm:
thanks.

PR #352, opened 2019-03-07
ArgMax/ArgMin support int32

nbcsm:
found in real model.
TF ArgMax/ArgMin may return int32 or int64, while ONNX ArgMax/ArgMin returns int64 only.
need to add cast when necessary.

enable dtype check in test by default by the way.


nbcsm:
thanks.

PR #358, opened 2019-03-08
implement Sign for opset9

mindest:
- add `Sign` op in `_OPSET_9`
- add `sign_op9`

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -670,7 +670,7 @@ def cast_op(ctx, node, name, args):
     return node
 
 
-def sign_op(ctx, node, name, args):
+def sign_op4(ctx, node, name, args):
     """Sign op."""
     # T sign = Sign(T Input)
     nodes = []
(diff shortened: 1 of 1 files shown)


tf2onnx/tfonnx.py
```diff
@@ -690,6 +690,18 @@ def sign_op(ctx, node, name, args):
...
 
 
+def sign_op9(ctx, node, name, args):
+    # Currently supported: `float32`
+    # Ignored: `bfloat16`
+    # TODO: add support for `int32`, `int64`
+    node_dtype = ctx.get_dtype(node.output[0])
+    utils.make_sure(node_dtype, "Dtype of {} is None".format(node.name))
+    if node_dtype in [onnx_pb.TensorProto.BOOL, onnx_pb.TensorProto.FLOAT16,
+                      onnx_pb.TensorProto.COMPLEX64, onnx_pb.TensorProto.COMPLEX128]:
```
nbcsm: float16 is supported according to onnx spec, remove it from here plz.
i mean we need to implement conversion logic based on onnx spec, but we may not need to test float16 for now,

tf2onnx/tfonnx.py
```diff
@@ -690,6 +690,18 @@ def sign_op(ctx, node, name, args):
     return nodes
 
 
+def sign_op9(ctx, node, name, args):
+    # Currently supported: `float32`
+    # Ignored: `bfloat16`
+    # TODO: add support for `int32`, `int64`
```
nbcsm: i mean you can (optionally) add a TODO at test side. :-)
just remove these 3 lines and keep the test as it is.

you can fix the cast issue we mentioned today and update test with more types in another PR.

tf2onnx/tfonnx.py
```diff
@@ -690,6 +690,18 @@ def sign_op(ctx, node, name, args):
     return nodes
 
 
+def sign_op9(ctx, node, name, args):
```
nbcsm: please also rename previous ```sign_op``` to ```sign_op4```

nbcsm:

PR #361, opened 2019-03-10
tf2onnx converted model cannot go to onnx-experimental.py because of naming conflicting

pengwa:
tf2onnx converted model cannot go to onnx-experimental.py because of naming conflicting


pengwa:
thanks @guschmue !

PR #364, opened 2019-03-11
modifiy sign_op4 and test_sign

mindest:
- in `sign_op4`, add support for dtypes `int32` and `int64`
- in the test case `test_sign`, add tests for the above two int dtypes

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1059,14 +1059,6 @@ class BackendTests(Tf2OnnxBackendTestBase):
         _ = tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: x_val})
 
-    @check_opset_min_version(9)
-    def test_sign_int(self):
-        x_val = np.array([1, 2, 0, -1, 0, -2], dtype=np.int).reshape((2, 3))
-        x = tf.placeholder(tf.int32, [2, 3], name=_TFINPUT)
-        x_ = tf.sign(x)
-        _ = tf.identity(x_, name=_TFOUTPUT)
-        self._run_test_case([_OUTPUT], {_INPUT: x_val})
-
     def test_sign(self):
         x_val1 = np.array([1, 2, 0, -1, 0, -2], dtype=np.int32).reshape((2, 3))
         x_val2 = np.array([1, 2, 0, -1, 0, -2], dtype=np.int64).reshape((2, 3))
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1068,11 +1068,15 @@ class BackendTests(Tf2OnnxBackendTestBase):
         self._run_test_case([_OUTPUT], {_INPUT: x_val})
 
     def test_sign(self):
-        x_val = np.array([1.0, 2.0, 0.0, -1.0, 0.0, -2.0], dtype=np.float32).reshape((2, 3))
-        x = tf.placeholder(tf.float32, [2, 3], name=_TFINPUT)
-        x_ = tf.sign(x)
-        _ = tf.identity(x_, name=_TFOUTPUT)
-        self._run_test_case([_OUTPUT], {_INPUT: x_val})
+        x_val1 = np.array([1, 2, 0, -1, 0, -2], dtype=np.int32).reshape((2, 3))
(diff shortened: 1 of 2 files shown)


nbcsm:
@mindest, please rebase your code onto latest master, there is a fix for pretrained_model_test.

tf2onnx/tfonnx.py
```diff
@@ -644,12 +644,16 @@ def sign_op4(ctx, node, name, args):
...
     utils.make_sure(node_dtype, "Dtype of {} is None".format(node.name))
     if node_dtype in [onnx_pb.TensorProto.COMPLEX64, onnx_pb.TensorProto.COMPLEX128]:
-        raise ValueError("dtype " + node_dtype + " is not supported in onnx for now")
-    input_tensor_type = utils.map_onnx_to_numpy_type(node_dtype)
+        raise ValueError("dtype " + str(node_dtype) + " is not supported in onnx for now")
     zero_name = utils.make_name("{}_zero".format(node.name))
-    ctx.make_const(zero_name, np.array(0, dtype=input_tensor_type))
-    greater_node = ctx.make_node("Greater", [node.input[0], zero_name])
-    less_node = ctx.make_node("Less", [node.input[0], zero_name])
+    ctx.make_const(zero_name, np.array(0, dtype=utils.ONNX_TO_NUMPY_DTYPE[1]))
```
nbcsm: ```1``` is not good to understand, use type name instead.

tf2onnx/tfonnx.py
```diff
@@ -644,12 +644,16 @@ def sign_op4(ctx, node, name, args):
...
+    if node_dtype in [onnx_pb.TensorProto.INT32, onnx_pb.TensorProto.INT64]:
+        cast_node_0 = ctx.make_node("Cast", [node.input[0]], {"to": 1})
```
nbcsm: so is the ```1``` here.

tests/test_backend.py
```diff
@@ -1068,11 +1068,15 @@ def test_sign_int(self):
         self._run_test_case([_OUTPUT], {_INPUT: x_val})
```
nbcsm: btw: does test_sign_int need opset9?

nbcsm:
@lucienwang1009 to take a look as well.

tests/test_backend.py
```diff
@@ -1068,11 +1068,15 @@ def test_sign_int(self):
...
 
     def test_sign(self):
```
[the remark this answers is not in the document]
mindest: The three opsets were all tested before this PR. I will test again after I update the codes.

tests/test_backend.py
mindest: > btw: does test_sign_int need opset9?

it does not. opset7 and opset8 also work fine with it.

mindest: > and do we still need test_sign_int since int type is already covered in test_sign

it seems unnecessary now, should I just remove it?

nbcsm: then just remove it.

tf2onnx/tfonnx.py
```diff
@@ -648,12 +648,16 @@ def sign_op4(ctx, node, name, args):
...
     if node_dtype in [onnx_pb.TensorProto.COMPLEX64, onnx_pb.TensorProto.COMPLEX128]:
-        raise ValueError("dtype " + node_dtype + " is not supported in onnx for now")
-    input_tensor_type = utils.map_onnx_to_numpy_type(node_dtype)
+        raise ValueError("dtype " + str(node_dtype) + " is not supported in onnx for now")
     zero_name = utils.make_name("{}_zero".format(node.name))
-    ctx.make_const(zero_name, np.array(0, dtype=input_tensor_type))
-    greater_node = ctx.make_node("Greater", [node.input[0], zero_name])
-    less_node = ctx.make_node("Less", [node.input[0], zero_name])
+    ctx.make_const(zero_name, np.array(0, dtype=np.float32))
+    if node_dtype in [onnx_pb.TensorProto.INT32, onnx_pb.TensorProto.INT64]:
```
nbcsm: @lucienwang1009, should this be if dtype NOT in [float16, float, double]?

PR #369, opened 2019-03-12
fix conv output shape

nbcsm:
found in real model.
onnx conv op output shape is not set correctly, which causes later transpose optimizer bug here

https://github.com/onnx/tensorflow-onnx/blob/master/tf2onnx/optimizer/transpose_optimizer.py#L94-L98


nbcsm:
@lucienwang1009

nbcsm:
thanks.

PR #371, opened 2019-03-14
shape inference support ExpandDims

nbcsm:
enhance tf shape inference to support ExpandDims

### DIFF
--- a/tf2onnx/shape_inference.py
+++ b/tf2onnx/shape_inference.py
@@ -175,7 +175,22 @@ def infer_shape_for_node(g, node):
 
         g.set_shape(node.output[0], new_shape)
         log.debug("set %s node [%s] with new shape %s", node.type, node.output[0], new_shape)
+        return True
 
+    if node.type == "ExpandDims":
+        # https://www.tensorflow.org/api_docs/python/tf/expand_dims
+        input_shape = g.get_shape(node.input[0])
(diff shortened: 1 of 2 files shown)


tf2onnx/tfonnx.py
```diff
@@ -907,9 +907,10 @@ def expanddims_op7(ctx, node, name, args):
     dim_node = node.inputs[1]
     if dim_node.is_const():
         node.type = "Unsqueeze"
-        input_rank = len(ctx.get_shape(node.input[0]))
         dim = dim_node.get_tensor_value()
-        dim = dim + input_rank + 1 if dim < 0 else dim
+        if dim < 0:
```
pengwa: change expanddims_op as well?

tf2onnx/shape_inference.py
```diff
@@ -175,7 +175,22 @@ def infer_shape_for_node(g, node):
 
         g.set_shape(node.output[0], new_shape)
         log.debug("set %s node [%s] with new shape %s", node.type, node.output[0], new_shape)
+        return True
+
+    if node.type == "ExpandDims":
```
pengwa: maybe we need start adding test for the inference.

nbcsm: yep, need to think about it.

PR #372, opened 2019-03-14
node get/set data_format via attribute internally

nbcsm:
fix bug found in real model

currently, ```node.data_format``` field is set in node constructor only.

when we create a new node from an existing node, e.g. in ```construct_graph_from_nodes```, existing node's attributes are passed into ```g.make_node```.

while in ```make_node```, ```data_format``` is treated like ```onnx_attrs``` if it is a ```AttributeProto```, so it calls ```node.set_attr_onnx``` directly, so the ```data_format``` field remains None in new node. it will cause problem for logic later that depends on ```data_format```

always get/set data_format via attribute to fix this issue.

### DIFF
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -46,10 +46,6 @@ class Node(object):
         # dict to original attributes
         for a in node.attribute:
             self._attr[a.name] = a
-
-        self.data_format = self.get_attr("data_format")
-        if self.data_format:
-            self.data_format = self.data_format.s.decode("utf-8")
         self._skip_conversion = skip_conversion
 
     @property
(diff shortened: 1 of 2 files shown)


tf2onnx/graph.py
```diff
@@ -141,17 +147,22 @@ def __repr__(self):
...
-        attr = self.attr.get(name)
+        """Get attribute value as int."""
+        attr = self.get_attr(name)
         utils.make_sure(attr is not None, "attribute %s is None", name)
         attr = attr.i
         return attr
 
+    def get_attr_str(self, name, encoding="utf-8"):
+        """Get attribute value as string."""
+        attr = self.get_attr(name)
```
pengwa: shall we keep the same behavior of get_attr_int, e.g. fast fail. if the attr is None.

nbcsm: for get_attr_int, a None attr value cannot be converted to int, so the check is required.
but for get_attr_str, return a None attr value means value does not exist, caller can decide what to do next, either fail or continue.

pengwa: get_attr() should be the function that check the attribute existence. when calling get_attr_str(), the dev is expecting the attribute exists and it's type is string in my option.

nbcsm:
thanks.

PR #373, opened 2019-03-14
support tf.math.round

nbcsm:
- support tf.math.round approximately
- fix loop_rewriter unused output shape

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -1638,6 +1638,24 @@ def where_op(ctx, node, name, args):
     return [node, transpose_node]
 
 
+def round_op7(ctx, node, name, args):
+    dtype = ctx.get_dtype(node.input[0])
+    if dtype in [onnx_pb.TensorProto.INT32, onnx_pb.TensorProto.INT64]:
+        node.type = "Identity"
+        return
+
+    # Workaround: for floating point, TF round mode is half to even, while round half up is used here.
+    # Will switch to onnx Round when it is ready.
+    # The difference now is for numbers like 2.5, 4.5 etc.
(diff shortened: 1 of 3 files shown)


tf2onnx/tfonnx.py
```diff
@@ -1638,6 +1638,24 @@ def where_op(ctx, node, name, args):
     return [node, transpose_node]
 
 
+def round_op7(ctx, node, name, args):
+    dtype = ctx.get_dtype(node.input[0])
```
pengwa: add TF signature like we did for better understanding. 

"    # T output = Softmax(T logits). The axis softmax would be performed on is always on -1.
     # T output = Softmax(T input, @int axis). Default axis is 1."

tf2onnx/tfonnx.py
```diff
@@ -1638,6 +1638,24 @@ def where_op(ctx, node, name, args):
...
+    if dtype in [onnx_pb.TensorProto.INT32, onnx_pb.TensorProto.INT64]:
+        node.type = "Identity"
+        return
+
+    # Workaround: for floating point, TF round mode is half to even, while round half up is used here.
+    # Will switch to onnx Round when it is ready.
+    # The difference now is for numbers like 2.5, 4.5 etc.
+    # In practise, it is supposed to have minor impact.
```
pengwa: if there are model having *.5, we will get result mismatch. shall we avoid this?

pengwa: it will not be a good experience to plan for a fix when we have a case with tons of efforts identifying it's this node causing mismatch.

nbcsm: make sense, will close this pr.

PR #378, opened 2019-03-15
check specific op (lstm, gru) in converted graph

lucienwang1009:
In unittests, check if converted graph contains specific ops, such as LSTM, GRU.

Other changes:
1. add time major and forget bias unittests for LSTM.
2. remove the requirement for forget bias in lstm_rewriter.

### DIFF
--- a/tests/test_gru.py
+++ b/tests/test_gru.py
@@ -16,7 +16,7 @@ from tensorflow.contrib import rnn
 from tensorflow.python.ops import init_ops
 from tensorflow.python.ops import variable_scope
 from backend_test_base import Tf2OnnxBackendTestBase
-from common import unittest_main
+from common import unittest_main, check_gru_count
 
 
 # pylint: disable=missing-docstring,invalid-name,unused-argument,using-constant-test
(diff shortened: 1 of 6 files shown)


nbcsm:
as we discussed, let's find a better way instead of ```must_contain_op="GRU", must_contain_op_num=2```

tf2onnx/rewriter/lstm_rewriter.py
```diff
@@ -73,7 +73,7 @@ def get_weight_and_bias(self, match):
         if not ft:
             return None
 
-        if not (ft.value == 1 and self.g.get_dtype(b_e.output[0]) == self.g.get_dtype(ft_bias.output[0])):
+        if not self.g.get_dtype(b_e.output[0]) == self.g.get_dtype(ft_bias.output[0]):
```
pengwa: maybe we should check ft_bias being a scalar?

lucienwang1009:
> as we discussed, let's find a better way instead of `must_contain_op="GRU", must_contain_op_num=2`

as we talked offline, a graph validator that checks graph attributes will be passed to run_test_case.

tf2onnx/rewriter/lstm_rewriter.py
lucienwang1009: Sure! using `np.ndim(ft.value) == 0` to check whether ft is a scalar

tests/test_gru.py
```diff
@@ -47,7 +47,8 @@ def test_single_dynamic_gru(self):
         input_names_with_port = ["input_1:0"]
         feed_dict = {"input_1:0": x_val}
         output_names_with_port = ["output:0", "cell_state:0"]
-        self.run_test_case(feed_dict, input_names_with_port, output_names_with_port, rtol=1e-03, atol=1e-06)
+        self.run_test_case(feed_dict, input_names_with_port, output_names_with_port, rtol=1e-03, atol=1e-06,
+                           graph_validator=lambda g: len(group_nodes_by_type(g)["GRU"]) == 1)
```
nbcsm: you may want to define a common helper function, e.g. ```check_op_count(g, op_type, expected_count)```

nbcsm: and maybe also define```check_gru_count```, ```check_lstm_count``` in corresponding files to save some typing.

PR #380, opened 2019-03-15
fix regression for bert, we broke if multiple model outputs came from the same node

guschmue:
less than idea fix for now but some folks are blocked on it.

### DIFF
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -338,8 +338,20 @@ class Graph(object):
         self.reset_nodes(ops)
 
         # add identity node after each output, in case it is renamed during conversion.
+        nodes_seen = set()
+        multi_output_nodes = set()
         for o in self.outputs:
             n = self.get_node_by_output_in_current_graph(o)
+            if n in nodes_seen:
+                multi_output_nodes.add(n)
+            else:
+                nodes_seen.add(n)
+
+        for o in self.outputs:
+            n = self.get_node_by_output_in_current_graph(o)
+            # TODO: below doesn't work for nodes with multiple outputs. A work around, keep those intact.
+            if n in multi_output_nodes:
+                continue
             new_output_name = port_name(n.name + "_" + utils.make_name("raw_output_"))
             n_shapes = n.output_shapes
             n_dtypes = n.output_dtypes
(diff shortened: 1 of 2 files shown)


tf2onnx/tfonnx.py
```diff
@@ -250,11 +250,18 @@ def arg_minmax_op(ctx, node, name, args):
...
 def reduce_op(ctx, node, name, args):
     axes_node = node.inputs[1]
-    input_rank = len(ctx.get_shape(node.input[0]))
     axes = axes_node.get_tensor_value()
     if np.isscalar(axes):
         axes = [axes]
-    axes = [val + input_rank if val < 0 else val for val in axes]
+    input_shape = ctx.get_shape(node.input[0])
+    if input_shape is None:
+        for val in axes:
```
pengwa: maybe use any([val < 0 for val in axes])

tf2onnx/graph.py
```diff
@@ -338,8 +338,20 @@ def __init__(self, nodes, output_shapes=None, dtypes=None, target=None, opset=No
...
         for o in self.outputs:
             n = self.get_node_by_output_in_current_graph(o)
+            if n in nodes_seen:
+                multi_output_nodes.add(n)
+            else:
+                nodes_seen.add(n)
+
+        for o in self.outputs:
+            n = self.get_node_by_output_in_current_graph(o)
+            # TODO: below doesn't work for nodes with multiple outputs. A work around, keep those intact.
```
pengwa: sorry for this regression. We can add the test just we did usually in the yaml file, could you please provide bert model and yaml file?

tf2onnx/tfonnx.py
```diff
@@ -250,11 +250,18 @@ def arg_minmax_op(ctx, node, name, args):
...
+            if val < 0:
+                raise ValueError("reduce_op: can have negative axis because we don't know input rank")
```
pengwa: do you intent to use "can**not** have negative axis"?

tf2onnx/graph.py
guschmue: no worry, this is the only model I know that takes multiple outputs from the same node as model output. bert needs a little extra training before its ready to convert ... need to find a place to download it from and its large. Maybe I can add a unit test to simulate the thing.

tf2onnx/tfonnx.py
guschmue: sold

tf2onnx/tfonnx.py
guschmue: yep.

PR #382, opened 2019-03-18
remove useless Identity node in nested graphs

pengwa:
Identity node may comes for different reasons:

1. original model has Identity node. 
2. converters inserted Identity node for easier handling. For example, graph output should never be changed, so for each op mapping function, we MUST make sure we consider "what if the node is generating the graph output". This is kind of heavy. What we did currently is: for a given graph output, we insert an Identity node generating it before all ops mapping happens. And for node not being Identity, the mapping function just assume the node won't be the one generating the graph output directly. For Identity node, it is easy to make sure it does not change the graph output.
3. In subgraphs such as Loop's Scan, sometimes tensor A:0, might be used as two different sub-graph outputs, (image LSTM case, hidden state is passed to next iteration, and also returned in each iteration, so it will be used as two different outputs). So in those cases, we intentionally insert a Identity node for each output.

It is better to do a clean up after all nodes get mapped correctly, so I added a IdentityOptimizer and triggered it at the end.

IdentityOptimizer just like TransposeOptimizer, can be used along with GraphUtils/Graph/Node structures, and easily used by other tools. But there is a potential unstable problem on onnx shape inferencing.

### DIFF
--- a/tests/run_pretrained_models.py
+++ b/tests/run_pretrained_models.py
@@ -266,7 +266,7 @@ class Test(object):
                 onnx_graph = self.to_onnx(sess.graph, opset=opset, shape_override=shape_override,
                                           input_names=inputs.keys())
                 model_proto = onnx_graph.make_model("converted from tf2onnx")
-                new_model_proto = GraphUtil.opt_transposes_with_graph(onnx_graph, "test", debug=debug)
+                new_model_proto = GraphUtil.optimize_graph(onnx_graph, "test", debug=debug)
                 if new_model_proto:
                     model_proto = new_model_proto
                 else:
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -19,6 +19,7 @@ import numpy as np
 from onnx import helper, numpy_helper, optimizer, shape_inference, OperatorSetIdProto, AttributeProto
 from tf2onnx import utils, __version__
 from tf2onnx.utils import port_name, find_opset
+from tf2onnx.optimizer.identity_optimizer import IdentityOptimizer
 from tf2onnx.optimizer.transpose_optimizer import TransposeOptimizer
 from tf2onnx.schemas import get_schema
 
@@ -842,6 +843,10 @@ class Graph(object):
         op_cnt = collections.Counter()
         for n in self.get_nodes():
             op_cnt[n.type] += 1
+            body_graphs = n.get_body_graphs()
+            if body_graphs:
+                for _, b_g in body_graphs.items():
+                    op_cnt += b_g.dump_node_statistics()
 
         return op_cnt
 
@@ -1025,16 +1030,19 @@ class GraphUtil(object):
     """Utilities for Graph manipulation."""
 
     @staticmethod
-    def opt_transposes_with_graph(graph, doc_string, optimize=None, debug=False):
-        """Optimize the graph, eliminating all useless Transpose pairs.
+    def optimize_graph(graph, doc_string, optimize=None, debug=False):
+        """Optimize the graph, for example: eliminating all useless Transpose/Identity pairs.
 
         Returns:
             model proto after optimization, if optimizer run successfully
             or None, if exceptions happen
         """
         try:
-            opt = TransposeOptimizer(graph, output_names=graph.outputs, debug=debug)
-            opt.optimize()
+            opts = [TransposeOptimizer(graph, output_names=graph.outputs, debug=debug),
+                    IdentityOptimizer(graph, output_names=graph.outputs, debug=debug)
+                   ]
+            for opt in opts:
+                opt.optimize()
             model_proto = graph.make_model(doc_string, optimize=optimize)
             return model_proto
         except Exception:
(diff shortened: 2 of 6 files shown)


tf2onnx/optimizer/identity_optimizer.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def optimize(self):
+        previous_counter = self._g.dump_node_statistics()
+        self._optimize_recursively(self._g)
+        current_counter = self._g.dump_node_statistics()
+        identity_cnt = current_counter["Identity"]
+        current_counter.subtract(previous_counter)
+        log.info(" %d identity op(s) left, ops diff after identity optimization: %s", identity_cnt, current_counter)
+
+    def _optimize_recursively(self, g):
```
nbcsm: seems to be a common requirement to traverse graph recursively, consider something like ```g.traverse_recursively(lamda g: func)```?

tf2onnx/optimizer/identity_optimizer.py
```diff
@@ -0,0 +1,104 @@
...
+                else:
+                    ret = self._handle_non_graph_output_identity(g, n)
+                has_update = ret
+
+        self._g.topological_sort(self._g.get_nodes())
+
+
+    def _handle_non_graph_output_identity(self, graph, identity):
+        graph.replace_all_inputs(graph.get_nodes(), identity.output[0], identity.input[0])
+        graph.remove_node(identity.name)
```
nbcsm: need to copy shape/dtype?

tf2onnx/optimizer/__init__.py
```diff
@@ -6,4 +6,4 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
-__all__ = ["transpose_optimizer"]
+__all__ = ["identity_optimizer", "transpose_optimizer"]
```
nbcsm: consider export IdentityOptimizer and TransposeOptimizer directly?
so user can use ```from tf2onnx.optimizer import IdentityOptimizer ``` to import specific optimizer or ```from tf2onnx.optimizer import *``` to import all.

```
from tf2onnx.optimizer.identity_optimizer import IdentityOptimizer
from tf2onnx.optimizer.transpose_optimizer import TransposeOptimizer
```
 feels too verbose

tf2onnx/optimizer/identity_optimizer.py
nbcsm: optional

nbcsm:
@zhijxu-MS is working on adding other optimizers currently, may need to unify/refactor some interfaces later.

tf2onnx/optimizer/identity_optimizer.py
pengwa: I would prefer to do this later. there are a few places to replace, which require some careful validation.

tf2onnx/optimizer/identity_optimizer.py
pengwa: not necessary actually, because this is just connecting consumers to another existing node.

tf2onnx/optimizer/__init__.py
pengwa: sounds good.

tf2onnx/optimizer/identity_optimizer.py
```diff
@@ -0,0 +1,98 @@
...
+            # input id already be graph output, so we cannot make that be another graph output.
+            # this Identity must be kept.
+            log.debug("identity input already be graph output")
+            return False
+
+        graph.remove_node(identity.name)
+        new_output = [output_id if o == input_id else o for o in input_node.output]
+        input_node.output = new_output
+
+        graph.set_shape(output_id, output_shape)
```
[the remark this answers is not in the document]
pengwa: yes we need. the optimizer all have Graph as input. it is assumed we should maintain a valid Graph, which would be optimized by another pass after this optimizer.

tests/run_pretrained_models.py
```diff
@@ -266,7 +266,7 @@ def run_test(self, name, backend="caffe2", debug=False, onnx_file=None, opset=No
                 onnx_graph = self.to_onnx(sess.graph, opset=opset, shape_override=shape_override,
                                           input_names=inputs.keys())
                 model_proto = onnx_graph.make_model("converted from tf2onnx")
-                new_model_proto = GraphUtil.opt_transposes_with_graph(onnx_graph, "test", debug=debug)
+                new_model_proto = GraphUtil.optimize_graph(onnx_graph, "test", debug=debug)
```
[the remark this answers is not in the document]
pengwa: there is another api that can call with onnx_model_proto. called optimize_graph_with_model_proto.

tf2onnx/graph.py
```diff
@@ -1025,16 +1030,19 @@ class GraphUtil(object):
     """Utilities for Graph manipulation."""
 
     @staticmethod
-    def opt_transposes_with_graph(graph, doc_string, optimize=None, debug=False):
-        """Optimize the graph, eliminating all useless Transpose pairs.
+    def optimize_graph(graph, doc_string, optimize=None, debug=False):
```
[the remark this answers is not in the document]
pengwa: let's do it when it is necessary. considered a optimizer registry and api entry, but we don't have strong requirement for that. so postponed it.

tf2onnx/graph.py
```diff
@@ -19,6 +19,7 @@
 from onnx import helper, numpy_helper, optimizer, shape_inference, OperatorSetIdProto, AttributeProto
 from tf2onnx import utils, __version__
 from tf2onnx.utils import port_name, find_opset
+from tf2onnx.optimizer.identity_optimizer import IdentityOptimizer
```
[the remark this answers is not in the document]
pengwa: let keep it consistent with tf2onnx.function/rewriter.

PR #384, opened 2019-03-18
opset 9 scan support

pengwa:


tests/common.py
```diff
@@ -155,6 +156,13 @@ def check_opset_min_version(min_required_version, message=""):
     return unittest.skipIf(config.opset < min_required_version, reason)
 
 
+def skip_specific_opset_version(opset_v, message=""):
```
nbcsm: just skip_opset(version, message)? seems clear enough.

tests/test_custom_rnncell.py
```diff
@@ -284,6 +284,7 @@ def test_multi_rnn_lstm(self, state_is_tuple=True):
 
     @check_opset_min_version(8, "Scan")
     @check_tf_min_version("1.8")
+    @skip_specific_opset_version(9, "ReverseSequence cannot be efficient mapped in OPSET 9")
```
nbcsm: just "ReverseSequence not supported" like others above? there is enough information in reverse_op9

tf2onnx/rewriter/custom_rnn_rewriter.py
```diff
@@ -105,13 +105,19 @@ def rewrite(self, context):
 
             state_inputs_initial_values = []
             for state_input in scan_props.state_inputs_initial_values:
-                nodes = self._adapt_scan_sequence_input_or_output("input", state_input, False)
-                state_inputs_initial_values.append(nodes[-1].output[0])
+                if self.g.opset == 8:
+                    nodes = self._adapt_scan_sequence_input_or_output("input", state_input, False)
+                    state_inputs_initial_values.append(nodes[-1].output[0])
+                else:
```
nbcsm: how about add a check somewhere, if opset < 8, skip rewrite.
maybe in need_rewrite or at the beginning of rewrite?

it will make this ```else``` more meaningful.
otherwise, we may end up with Scan in a opset 7 model

nbcsm: [optional] maybe also print some message like "rnn found, but needs higher opset"?

tf2onnx/tfonnx.py
```diff
@@ -1454,6 +1454,15 @@ def reverse_op8(ctx, node, name, args):
     node.input[0] = node.input[1]
     node.input[1] = tmp
 
+def reverse_op9(ctx, node, name, args):
+    # T output = ReverseSequence(T input, int32|int64 seq_lengths, @int seq_dim, @int batch_dim)
+    # we cannot easily construct reverse_sequence equivalence in opset 9, so we will not support it
+    # here. Actually using loops to do that is kind of meaningless since there will be performance
+    # issue there for sure.
+
+    raise RuntimeError("ReverseSequence is not supported to convert in OPSET9,"
```
nbcsm: NotImplementedError seems better.

pengwa:
thanks for the review.

nbcsm:

PR #388, opened 2019-03-19
Support ellipse mask in stridedslice op

jiafatom:


guschmue:
Do you have a test case ?

jiafatom:
> Do you have a test case ?



> Do you have a test case ?

@guschmue updated unit test

PR #389, opened 2019-03-19
fix tf.greater_equal

guschmue:


tf2onnx/tfonnx.py
```diff
@@ -1633,6 +1633,11 @@ def logical_compare_op(ctx, node, name, args):
                 ctx.copy_shape(inp, inp_cast.output[0])
                 ctx.set_dtype(inp_cast.output[0], target_dtype)
 
+def logical_compareeq_op(ctx, node, name, args):
+    logical_compare_op(ctx, node, name, [])
+    ctx.insert_new_node_on_output("Not", node.output[0], name=utils.make_name(name),
```
pengwa: insert_new_node_on_output does not have shapes/dtypes params. so we have to set_shape/set_dtype after the insertion.

pengwa: this can fix some of the failed ut cases I guess.

guschmue: yeah, I was just debugging it :) Thought it was passing the kwargs through. Going to fix it.

pengwa: we could later add the params for it to keep it consistent with Graph.make_node().

guschmue: I think that fixed it.

guschmue:
thanks!!

tf2onnx/tfonnx.py
```diff
@@ -1812,9 +1821,7 @@ def where_op(ctx, node, name, args):
     "Cosh": (direct_op, []),
     "Erf": (direct_op, []),
     "Fill": (fill_op, []),
-    "Greater": (logical_compare_op, []),
```
nbcsm: @guschmue, why Greater/Less are removed?
it is for opset9.

PR #392, opened 2019-03-20
fix python3.5 issue - subgraph input orders are lost after loading Gr…

pengwa:
…aph from graph_proto

@zhijxu-MS


pengwa:
thanks @zhijxu-MS !! merged.

Issue #391, opened 2019-03-20
Fill_op for opset 9 causes onnxruntime type mismatch


jiafatom:
I am using the following code in the test framework
` def test_fill(self):
        from keras import backend as K
        class_box_scores = K.variable(np.random.random((2, 3)))
        c = 20
        classes_ = K.ones_like(class_box_scores, 'int32') * K.constant(value=c, dtype='int32')
        _ = tf.identity(classes_, name="output")
        input_names_with_port = []
        feed_dict = {}
        output_names_with_port = ["output:0"]
        self.run_test_case(feed_dict, input_names_with_port, output_names_with_port, rtol=1e-06)`

then it causes onnx runtime error when I using **opset=9**
RuntimeError: [ONNXRuntimeError] : 1 : GENERAL ERROR : Load model from C:\Users\jiafa\AppData\Local\Temp\tmpkn0tod4m\test_simple_while_loop\test_simple_while_loop.onnx failed:Type Error: Type parameter (T) bound to different types (tensor(float) and tensor(int32) in node (mul).

nbcsm:
it seems to be caused by onnx version.
we will add check for onnx version, and skip attributes check for unknown ops.

nbcsm:
@jiafatom, could you try again please? thanks.

nbcsm:
will close it, please reopen if there is any issue.

PR #393, opened 2019-03-20
upgrade required min onnx version to 1.4.1

nbcsm:
currently, we need to get op schema from onnx python api to filter out attributes that not belong to onnx op.
onnx < 1.4 does not support opset9, which will cause issue when people try to use opset 9.
e.g. https://github.com/onnx/tensorflow-onnx/issues/391

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,5 +80,5 @@ setup(
     author='onnx@microsoft.com',
     author_email='onnx@microsoft.com',
     url='https://github.com/onnx/tensorflow-onnx',
-    install_requires=['numpy>=1.14.1', 'onnx>=1.2.2', 'six']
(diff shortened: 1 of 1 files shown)


guschmue:
Some stuff like README mentions the onnx version .... it would need to be updated as well.
In general it not good to have a dependency on the onnx version - it is artificial, we never had this.
If you depend on the newer onnx version for opset 9, that newer onnx version doesn't know much about older opsets you break those.
We should avoid this, it is going to create trouble.

guschmue:
Maybe its ok as long older opsets are not purged from the onnx package. Let me check with Ke on that,

nbcsm:
onnx package itself has backward compatibility, new package supports old opsets as well.
thanks for review.

PR #394, opened 2019-03-20
refine onnx node attributes check

nbcsm:
- check whether currently installed onnx package version supports opset version specified
- skip onnx node attributes check for unknown op
- support domain argument when create node
- when user creates node with op in non onnx domain, make sure domain is kept when dealing with graph output node

### DIFF
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -378,7 +378,8 @@ class Graph(object):
 
             new_outputs = [o if o != output else new_output_name for output in n.output]
             new_node = self.make_node(n.type, n.input, outputs=new_outputs, attr=n.attr, name=n.name,
-                                      skip_conversion=n._skip_conversion, dtypes=n_dtypes, shapes=n_shapes)
+                                      skip_conversion=n._skip_conversion, dtypes=n_dtypes, shapes=n_shapes,
+                                      domain=n.domain)
 
             if body_graphs:
                 for attr_name, body_graph in body_graphs.items():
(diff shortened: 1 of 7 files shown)


nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/391

nbcsm:
@lucienwang1009

tf2onnx/schemas.py
```diff
@@ -77,16 +76,39 @@ def _register_all_schemas_with_history():
     return ordered_map
 
 
+def _parse_domain_opset_versions(schemas):
```
pengwa: why we don't simply check onnx package version instead of this? if the version is not for opset 9 release, then we fail it.

tf2onnx/graph.py
```diff
@@ -449,7 +459,7 @@ def make_node(self, op_type, inputs, attr=None, output_count=1, outputs=None, sk
             n = self.get_node_by_output_in_current_graph(o)
             utils.make_sure(n is None, "output tensor named %s already exists in node: \n%s", o, n)
 
-        onnx_node = helper.make_node(op_type, inputs, outputs, name=name, **raw_attr)
+        onnx_node = helper.make_node(op_type, inputs, outputs, name=name, domain=domain, **raw_attr)
```
pengwa: is a None domain okay for this call?

nbcsm: yes.

tf2onnx/schemas.py
nbcsm: then we have to keep an onnx package version -> max opset version mapping.
and every time an onnx package is released with new opset, we need to update this mapping.

nbcsm:
thanks for review.

PR #395, opened 2019-03-20
use onehot to support some ops found in real tf model and fix bug of lstm

zhijxu-MS:
1   use onehot to softmax with cross entropy
2   fix bug of lstm
3   workaround for the bug of tf fold_const
4 relax atol from 0 to 1e-6


tf2onnx/tfonnx.py
```diff
@@ -2229,6 +2231,28 @@ def rewrite_conv2d_with_pad(g, ops):
     return ops
 
 
+def rewrite_const_sub_with_pack(g, ops):
```
guschmue: why not enable rewrite_constant_fold ?

tf2onnx/tfonnx.py
```diff
@@ -1806,6 +1806,7 @@ def where_op(ctx, node, name, args):
     "ResizeNearestNeighbor": (upsample_op7, ["Upsample", "nearest"]),
     "Sin": (direct_op, []),
     "Sub": (broadcast_op7, []),
+    "SoftmaxCrossEntropyWithLogits": (softmax_cross_entropy_with_logits_op, []),
```
nbcsm: rename to softmax_cross_entropy_with_logits_op7?

tf2onnx/tfonnx.py
```diff
@@ -1835,6 +1836,7 @@ def where_op(ctx, node, name, args):
     "ReverseSequence": (reverse_op9, []),
     "Sign": (sign_op9, []),
     "Sinh": (direct_op, []),
+    "SparseSoftmaxCrossEntropyWithLogits": (sparse_softmax_cross_entropy_with_logits_op9, []),
```
nbcsm: please also add a mapping for OneHot, maybe a new PR.

tf2onnx/function/sparse_softmax_cross_entropy_with_logits.py
```diff
@@ -12,6 +12,37 @@
 
 # pylint: disable=unused-argument,missing-docstring
 
+
+def softmax_cross_entropy_with_logits_computation(ctx, label, logit, tf_ori_node):
```
nbcsm: make it internal.

nbcsm: rename to make_softmax_cross_entropy_with_logits?

tf2onnx/tfonnx.py
zhijxu-MS: will add another PR to avoid making the PR too complex.

tf2onnx/function/sparse_softmax_cross_entropy_with_logits.py
```diff
@@ -12,6 +12,37 @@
 
 # pylint: disable=unused-argument,missing-docstring
```
nbcsm: replace this file to softmax_cross_entropy_with_logits.py?

tests/run_pretrained_models.py
```diff
@@ -80,7 +80,7 @@ class Test(object):
     target = []
 
     def __init__(self, url, local, make_input, input_names, output_names,
-                 disabled=False, more_inputs=None, rtol=0.01, atol=0.,
```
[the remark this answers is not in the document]
zhijxu-MS: done

tf2onnx/function/softmax_cross_entropy_with_logits.py
```diff
@@ -12,6 +12,37 @@
...
+    logit_dtype = ctx.get_dtype(logits.output[0])
+    labels = node.inputs[1]
+    label_dtype = ctx.get_dtype(labels.output[0])
+    if label_dtype != logit_dtype:
+        labels = ctx.make_node("Cast", labels.output, attr={"to": logit_dtype}, dtypes=[logit_dtype])
+
+    _make_softmax_cross_entropy_with_logits(ctx, labels, logits, node)
+
+
 def sparse_softmax_cross_entropy_with_logits_op(ctx, node, name, args):
```
nbcsm: sparse_softmax_cross_entropy_with_logits_op5?
5, right?

tf2onnx/tfonnx.py
nbcsm: @zhijxu-MS, let's withdraw rewrite_const_sub_with_pack from this PR to unblock onehot.
const fold can be in another pr.

tf2onnx/function/softmax_cross_entropy_with_logits.py
zhijxu-MS: 7. mul doesn't support broadcasting automatically before.

tf2onnx/tfonnx.py
zhijxu-MS: @guschmue there is a bug in rewrite_constant_fold when I tried with a real model and after discussion with @nbcsm we decided to enhance const fold in another way.

I will withdraw my changes in this PR, and send the enhancement in another PR.

guschmue: Works for me.  I think fixing rewrite_constant_fold is the better option so we have all const fold in one place. rewrite_constant_fold used to be needed for older tensorflow versions (for example to get bert work) but since 1.12 I have not seen issues.

tests/test_backend.py
```diff
@@ -1612,6 +1612,24 @@ def test_shape_int64(self):
         kwargs = {"check_dtype": True}
         self._run_test_case([_OUTPUT], {_INPUT: x_val}, **kwargs)
 
+    # @check_opset_min_version(7, "broadcasting op")
+    @unittest.skip("disable it for node, since fold const has bug")
```
nbcsm: node => now?

tf2onnx/tfonnx.py
nbcsm: https://github.com/onnx/tensorflow-onnx/issues/417

PR #397, opened 2019-03-21
fix an issue when subgraph output node has more than one output

pengwa:


pengwa:
@guschmue  I am using Split to test graph output node having multiple outputs. Do you think this can cover Bert's failure at that time?

guschmue:
bert converts fine on master, need to check if it works with your change.
But bert has some issue at runtime, I know it works a few weeks back, looking at that first.

PR #398, opened 2019-03-21
fix some bugs

lucienwang1009:
1. correct input dtype when run pretrained model.
2. support more dtype for Maximum and Minimum and add unittests.
3. any empty input of concat op will lead to seg fault on onnxruntime. here I workaround it by remove concat op when its input is empty.
4. fix bugs when parsing tensor value of const op.

### DIFF
--- a/tests/run_pretrained_models.py
+++ b/tests/run_pretrained_models.py
@@ -15,6 +15,7 @@ import tarfile
 import time
 import traceback
 import zipfile
+import logging
 
 import PIL.Image
 import numpy as np
@@ -33,6 +34,9 @@ from tf2onnx.graph import GraphUtil
 from tf2onnx.tfonnx import process_tf_graph
 
 # pylint: disable=broad-except,logging-not-lazy,unused-argument,unnecessary-lambda
+
+logging.basicConfig(level=logging.INFO)
+log = logging.getLogger("tf2onnx")
 
 TEMP_DIR = os.path.join(utils.get_temp_directory(), "run_pretrained")
 PERFITER = 1000
@@ -246,9 +250,10 @@ class Test(object):
             for k in inputs.keys():  # pylint: disable=consider-iterating-dictionary
                 t = sess.graph.get_tensor_by_name(k)
                 dtype = tf.as_dtype(t.dtype).name
-                if type != "float32":
-                    v = inputs[k]
-                    inputs[k] = v.astype(dtype)
+                v = inputs[k]
+                if dtype != v.dtype:
+                    log.warning("input dtype doesn't match tensorflow's")
+                    inputs[k] = np.array(v, dtype=dtype)
             if self.force_input_shape:
                 for k, v in inputs.items():
                     shape_override[k] = list(v.shape)
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -769,8 +769,11 @@ def concatv2_op(ctx, node, name, args):
     # if any input is empty, remove the input and concat the others
     # NOTE: workaround for https://github.com/Microsoft/onnxruntime/issues/681
     for i, inp in enumerate(node.inputs):
-        if inp.is_const() and inp.get_tensor_value() == []:
+        if inp.is_const() and inp.get_tensor_value(as_list=False).size == 0:
             ctx.remove_input(node, node.input[i])
+    # all inputs are deleted
+    if not node.input:
+        raise RuntimeError("all inputs of {} are empty".format(name))
 
     axis_node = node.inputs[-1]
     axis_val = axis_node.get_tensor_value()
@@ -1175,7 +1178,7 @@ def minmax_op(ctx, node, name, args):
             ctx.copy_shape(inp, inp_cast.output[0])
             ctx.set_dtype(inp_cast.output[0], target_dtype)
     origin_dtype = ctx.get_dtype(node.output[0])
-    utils.make_sure(origin_dtype, "dtype of {} is None".format(node.output[0]))
+    utils.make_sure(origin_dtype is not None, "dtype of {} is None".format(node.output[0]))
     ctx.set_dtype(node.output[0], target_dtype)
     cast_name = utils.make_name(name)
     cast_node = ctx.insert_new_node_on_output("Cast", node.output[0], name=cast_name, to=origin_dtype)
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -766,6 +766,12 @@ def concat_op(ctx, node, name, args):
 def concatv2_op(ctx, node, name, args):
     # T output = ConcatV2(T values, Tidx axis, @int N, @type Tidx)
     # T concat_result = Concat(T inputs, @INT axis)
+    # if any input is empty, remove the input and concat the others
+    # NOTE: workaround for https://github.com/Microsoft/onnxruntime/issues/681
+    for i, inp in enumerate(node.inputs):
+        if inp.is_const() and inp.get_tensor_value() == []:
+            ctx.remove_input(node, node.input[i])
+
     axis_node = node.inputs[-1]
     axis_val = axis_node.get_tensor_value()
     ctx.remove_input(node, node.input[-1])
@@ -1154,6 +1160,28 @@ def minmax_op(ctx, node, name, args):
     # handle this by doing something like:
     # y = min(x1, add(x2, sub(x1, x1))), where x1, x2 are the inputs and x2 is a scalar
     # this will create a tensor of zeros of the shape of x1, adds x2 to it (which broadcasts) and use that for min.
+    # support more dtype
+    supported_dtypes = [
+        onnx_pb.TensorProto.FLOAT,
+        onnx_pb.TensorProto.FLOAT16,
+        onnx_pb.TensorProto.DOUBLE
+    ]
+    target_dtype = onnx_pb.TensorProto.FLOAT
+    for inp in node.input:
+        dtype = ctx.get_dtype(inp)
+        utils.make_sure(dtype, "dtype of {} is None".format(inp))
+        if dtype not in supported_dtypes:
+            inp_cast = ctx.insert_new_node_on_input(node, "Cast", inp, to=target_dtype)
+            ctx.copy_shape(inp, inp_cast.output[0])
+            ctx.set_dtype(inp_cast.output[0], target_dtype)
+    origin_dtype = ctx.get_dtype(node.output[0])
+    utils.make_sure(origin_dtype, "dtype of {} is None".format(node.output[0]))
+    ctx.set_dtype(node.output[0], target_dtype)
+    cast_name = utils.make_name(name)
+    cast_node = ctx.insert_new_node_on_output("Cast", node.output[0], name=cast_name, to=origin_dtype)
+    to_replace = [n for n in ctx.get_nodes() if n != cast_node]
+    ctx.replace_all_inputs(to_replace, node.output[0], cast_node.output[0])
+
     shapeo = ctx.get_shape(node.output[0])
     needs_broadcast_op = []
     has_correct_shape = []
@@ -1676,6 +1704,7 @@ _OPSET_4 = {
     "BiasAdd": (biasadd_op, []),
     "BiasAddV1": (biasadd_op, []),
     "Cast": (cast_op, []),
+    "CheckNumerics": (identity_op, ["Identity"]),
     "Concat": (concat_op, ["Concat"]),
     "ConcatV2": (concatv2_op, ["Concat"]),
     "Const": (direct_op, []),
--- a/tf2onnx/utils.py
+++ b/tf2onnx/utils.py
@@ -131,9 +131,9 @@ def tf_to_onnx_tensor(tensor, name=""):
         tensor_content: empty
         tensor_shape.dim: [0]
         DTYPE_val: 1
-    3. empty tensor, e.g., np.array([], dtype=DTYPE):
+    3. empty tensor, e.g., np.array([], dtype=DTYPE) and np.array([[]], dtype=DTYPE):
         tensor_content: empty
-        tensor_shape.dim: [0]
+        tensor_shape.dim: [0] and [1, 0]
         DTYPE_val: empty
     """
     new_type = TF_TO_ONNX_DTYPE[tensor.dtype]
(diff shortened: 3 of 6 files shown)


tf2onnx/tfonnx.py
```diff
@@ -1746,6 +1774,7 @@ def where_op(ctx, node, name, args):
     "SquaredDifference": (squareddifference_op, []),
     "Softmax": (softmax_op, ["Softmax"]),
     "StopGradient": (identity_op, ["Identity"]),
+    "CheckNumerics": (identity_op, ["Identity"]),
```
nbcsm: order

tf2onnx/utils.py
```diff
@@ -151,16 +154,13 @@ def get_tf_tensor_data(tensor):
...
-        data = [0.]
-    elif tensor.dtype == tf.float16:
-        data = [0]
     elif tensor.string_val:
         data = tensor.string_val
+    elif tensor.dtype == tf.int32 or \
+        tensor.dtype == tf.int64 or \
+        tensor.dtype == tf.float32 or \
+        tensor.dtype == tf.float16:
+        data = None
```
nbcsm: tensor.dtype in [x, x, x,]?

tf2onnx/tfonnx.py
```diff
@@ -766,6 +766,14 @@ def concat_op(ctx, node, name, args):
 def concatv2_op(ctx, node, name, args):
     # T output = ConcatV2(T values, Tidx axis, @int N, @type Tidx)
     # T concat_result = Concat(T inputs, @INT axis)
+    # if any input is empty, remove this concat op
+    # TODO: file a issue to onnxruntime that empty input of concat op
+    # will leads seg fault
```
nbcsm: just file it and fill the link here. :-)

tf2onnx/utils.py
```diff
@@ -121,17 +121,20 @@ def tf_to_onnx_tensor(tensor, name=""):
...
-        dims = [1]
     is_raw, data = get_tf_tensor_data(tensor)
+    # data is [], i.e., empty array
+    if dims == [0] and not is_raw and not data:
+        np_data = np.array([], dtype=map_onnx_to_numpy_type(new_type))
+        return numpy_helper.from_array(np_data, name=name)
+    # data is a scalar
+    if dims == [0] and not is_raw and len(data) == 1:
+        return helper.make_tensor(name, new_type, [], data, False)
+    # which case?
```
nbcsm: seems it is for https://github.com/onnx/tensorflow-onnx/pull/162
just remove this comment.

tf2onnx/utils.py
```diff
@@ -121,17 +121,20 @@ def tf_to_onnx_tensor(tensor, name=""):
     new_type = TF_TO_ONNX_DTYPE[tensor.dtype]
     tdim = tensor.tensor_shape.dim
     dims = [d.size for d in tdim]
-    # FIXME: something is fishy here
-    if dims == [0]:
-        dims = [1]
     is_raw, data = get_tf_tensor_data(tensor)
+    # data is [], i.e., empty array
```
nbcsm: please add more comments with examples, otherwise, it is hard for others to understand.

tests/run_pretrained_models.py
```diff
@@ -246,9 +246,9 @@ def run_test(self, name, backend="caffe2", debug=False, onnx_file=None, opset=No
             for k in inputs.keys():  # pylint: disable=consider-iterating-dictionary
                 t = sess.graph.get_tensor_by_name(k)
                 dtype = tf.as_dtype(t.dtype).name
-                if type != "float32":
-                    v = inputs[k]
-                    inputs[k] = v.astype(dtype)
+                v = inputs[k]
+                if dtype != v.dtype:
```
pengwa: inputs is provided in yaml, right? If yes, shall we add a warning if yaml giving a type not matching tf type? this should be a error for yaml config I feel.

tf2onnx/tfonnx.py
```diff
@@ -766,6 +766,14 @@ def concat_op(ctx, node, name, args):
...
+    for i, inp in enumerate(node.inputs):
+        if inp.is_const() and inp.get_tensor_value() == []:
+            ctx.replace_all_inputs(ctx.get_nodes(), node.output[0], node.inputs[1-i].output[0])
+            return
```
pengwa: you will only handle on input, return is used correctly?

pengwa: concat might have more than two inputs, what if two are valid, the other one is []?

tf2onnx/tfonnx.py
```diff
@@ -1154,6 +1162,25 @@ def minmax_op(ctx, node, name, args):
...
     # this will create a tensor of zeros of the shape of x1, adds x2 to it (which broadcasts) and use that for min.
+    # support more dtype
+    supported_dtypes = [
+        onnx_pb.TensorProto.FLOAT,
+        onnx_pb.TensorProto.FLOAT16,
+        onnx_pb.TensorProto.DOUBLE
+    ]
+    target_dtype = onnx_pb.TensorProto.FLOAT
+    for inp in node.input:
+        if ctx.get_dtype(inp) not in supported_dtypes:
```
pengwa: check the dtype is None before this?

tf2onnx/tfonnx.py
```diff
@@ -1154,6 +1162,25 @@ def minmax_op(ctx, node, name, args):
...
+            inp_cast = ctx.insert_new_node_on_input(node, "Cast", inp, to=target_dtype)
+            ctx.copy_shape(inp, inp_cast.output[0])
+            ctx.set_dtype(inp_cast.output[0], target_dtype)
+    origin_dtype = ctx.get_dtype(node.output[0])
+    ctx.set_dtype(node.output[0], target_dtype)
+    cast_name = utils.make_name(name)
+    cast_node = ctx.insert_new_node_on_output("Cast", node.output[0], name=cast_name, to=origin_dtype)
```
pengwa: origin_dtype might be NOne?

tf2onnx/tfonnx.py
pengwa: is doing this way safe enough?

pengwa: I mean we should try not hiding any issues to make our debugging hard

tf2onnx/utils.py
```diff
@@ -121,17 +121,20 @@ def tf_to_onnx_tensor(tensor, name=""):
...
+    if dims == [0] and not is_raw and not data:
```
pengwa: use data ==[] instead of "not data"?

tf2onnx/utils.py
```diff
@@ -151,16 +154,13 @@ def get_tf_tensor_data(tensor):
...
-        data = [0]
-    elif tensor.dtype == tf.int64:
-        data = [0]
-    elif tensor.dtype == tf.float32:
-        data = [0.]
-    elif tensor.dtype == tf.float16:
-        data = [0]
     elif tensor.string_val:
         data = tensor.string_val
+    elif tensor.dtype == tf.int32 or \
```
pengwa: can you explain why we don this check here?

tf2onnx/tfonnx.py
lucienwang1009: orz, thanks!

tf2onnx/tfonnx.py
```diff
@@ -1154,6 +1160,29 @@ def minmax_op(ctx, node, name, args):
...
+    supported_dtypes = [
+        onnx_pb.TensorProto.FLOAT,
+        onnx_pb.TensorProto.FLOAT16,
+        onnx_pb.TensorProto.DOUBLE
+    ]
+    target_dtype = onnx_pb.TensorProto.FLOAT
+    for inp in node.input:
+        dtype = ctx.get_dtype(inp)
+        if not dtype:
+            log.warning("dtype of %s is None", inp)
```
nbcsm: print a warning won't help...
dtype SHOULD be here, unless we screwed up somewhere earlier...
how about make_sure?

tf2onnx/tfonnx.py
```diff
@@ -1676,6 +1705,7 @@ def where_op(ctx, node, name, args):
     "BiasAdd": (biasadd_op, []),
     "BiasAddV1": (biasadd_op, []),
     "Cast": (cast_op, []),
+    "CheckNumerics": (identity_op, ["Identity"]),
```
nbcsm: @pengwa , according to tf [spec](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/check-numerics), it will throw if there is Nan or Inf, otherwise, pass the tensor through.
for onnx, seems there is no equivalent like Assert, and we cannot really tell whether there is Nan/Inf at converting stage.
so seems all we can do it is map to Identity.

nbcsm: @lucienwang1009, is this from user script or some TF api call?

tf2onnx/utils.py
nbcsm: agree, for empty tenosr, it should be```data is None```?

tf2onnx/utils.py
```diff
@@ -120,21 +120,38 @@ def split_nodename_and_shape(name):
...
     tdim = tensor.tensor_shape.dim
     dims = [d.size for d in tdim]
-    # FIXME: something is fishy here
-    if dims == [0]:
-        dims = [1]
     is_raw, data = get_tf_tensor_data(tensor)
+    # empty tensor
+    if dims == [0] and not is_raw and not data:
+        np_data = np.array([], dtype=map_onnx_to_numpy_type(new_type))
+        return numpy_helper.from_array(np_data, name=name)
```
nbcsm: data should not be None from here, right?
let's add a make_sure.

tf2onnx/utils.py
lucienwang1009: we return None if we cannot get the tentor data. So here data is exactly None instead of "[]"

tf2onnx/tfonnx.py
lucienwang1009: it's in user script, nevertheless this op won't affect the model output.

tf2onnx/tfonnx.py
lucienwang1009: If input dtype is None while output dtype is given, this conversion can work.
Anyway, make_sure can throw the error early so that we can debug it more easily.

tf2onnx/utils.py
lucienwang1009: hmm, `data = None` here is just a indicator for empty tensor data.

nbcsm: yep, i mean add a check AFTER dealing with empty tensor?

lucienwang1009: oh, yes, it won't be nether None nor []

tf2onnx/tfonnx.py
```diff
@@ -766,6 +766,12 @@ def concat_op(ctx, node, name, args):
 def concatv2_op(ctx, node, name, args):
     # T output = ConcatV2(T values, Tidx axis, @int N, @type Tidx)
     # T concat_result = Concat(T inputs, @INT axis)
+    # if any input is empty, remove this concat op
```
pengwa: the comment is not update to date.

tf2onnx/tfonnx.py
```diff
@@ -766,6 +766,12 @@ def concat_op(ctx, node, name, args):
...
+    # NOTE: workaround for https://github.com/Microsoft/onnxruntime/issues/681
+    for i, inp in enumerate(node.inputs):
+        if inp.is_const() and inp.get_tensor_value() == []:
```
pengwa: so original there are 2 inputs, remove one of them, concat having one input is valid, right? can you confirm on ONNX side and ONNXRuntime side?

tf2onnx/utils.py
pengwa: okay, then use "data is None" for the check?

tf2onnx/utils.py
```diff
@@ -154,16 +171,15 @@ def get_tf_tensor_data(tensor):
...
-    elif tensor.dtype == tf.int64:
-        data = [0]
-    elif tensor.dtype == tf.float32:
-        data = [0.]
-    elif tensor.dtype == tf.float16:
-        data = [0]
     elif tensor.string_val:
         data = tensor.string_val
+    elif tensor.dtype in [
+            tf.int32,
```
pengwa: please have a comprehensive testing for this change locally.

tf2onnx/tfonnx.py
lucienwang1009: yes, both of ONNX and ONNXRuntime accept this kind of input.

tf2onnx/utils.py
lucienwang1009: all tests including internal models are passed.

nbcsm:
@lucienwang1009 , please add a few tests to cover these special cases in ```tf_to_onnx_tensor``` and ```get_tf_tensor_data```, maybe add to ```test_internals```?
either in this pr or a new one.

nbcsm:
@lucienwang1009, and also a test to cover concat input removal case.

tf2onnx/tfonnx.py
```diff
@@ -1154,6 +1160,28 @@ def minmax_op(ctx, node, name, args):
...
+    target_dtype = onnx_pb.TensorProto.FLOAT
+    for inp in node.input:
+        dtype = ctx.get_dtype(inp)
+        utils.make_sure(dtype, "dtype of {} is None".format(inp))
+        if dtype not in supported_dtypes:
+            inp_cast = ctx.insert_new_node_on_input(node, "Cast", inp, to=target_dtype)
+            ctx.copy_shape(inp, inp_cast.output[0])
+            ctx.set_dtype(inp_cast.output[0], target_dtype)
+    origin_dtype = ctx.get_dtype(node.output[0])
+    utils.make_sure(origin_dtype, "dtype of {} is None".format(node.output[0]))
```
pengwa: better to use origin_dtype is not None for a check in the further.

tf2onnx/tfonnx.py
```diff
@@ -769,7 +769,7 @@ def concatv2_op(ctx, node, name, args):
     # if any input is empty, remove the input and concat the others
     # NOTE: workaround for https://github.com/Microsoft/onnxruntime/issues/681
     for i, inp in enumerate(node.inputs):
-        if inp.is_const() and inp.get_tensor_value() == []:
+        if inp.is_const() and inp.get_tensor_value(as_list=False).size == 0:
```
nbcsm: err, what if all inputs are const and empty?
we don't really need to support this case, but better to add a check and throw.

Issue #402, opened 2019-03-22
support ms domain


nbcsm:
enable converter to support multiple domains, e.g. MS domain.

PR #404, opened 2019-03-23
reorganize onnx domain mapping functions

nbcsm:
- move onnx domain mapping functions in _tfonn.py_ to _mappings\onnx\functions.py_
- split _tfonnx.py_ while keep commit history for both files
the first 4 commits did the trick.
- move scripts under _function_ folder to _mappings\onnx_

it is almost pure code cut & paste, no change on existing api or mapping function behavior.


nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/402

PR #406, opened 2019-03-25
fix onnx package version check

nbcsm:
fix onnx package version check issue when opset argument is not specified


nbcsm:
unit_test failed due to accuracy, rerun passed but not showed here.
https://dev.azure.com/tensorflow-onnx/tensorflow-onnx/_build/results?buildId=2144

nbcsm:
thanks for review

PR #407, opened 2019-03-26
fix some transpose optimizer bugs

zhijxu-MS:
fix transpose optimizer bug

node can't be removed directly when graph output comes from it

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -340,8 +340,13 @@ class TransposeOptimizer(object):
             ops = self._g.get_nodes()
             self._g.replace_all_inputs(ops, node.output[0], trans.input[0])
 
+            shape = self._g.get_shape(node.output[0])
+            dtype = self._g.get_dtype(node.output[0])
             self._g.remove_node(trans.name)
             self._g.remove_node(node.name)
+            if node.output[0] in self._g.outputs:
+                self._g.make_node("Identity", [trans.input[0]],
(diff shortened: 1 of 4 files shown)


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -340,8 +340,13 @@ def _transpose_handler(self, trans, node):
             ops = self._g.get_nodes()
             self._g.replace_all_inputs(ops, node.output[0], trans.input[0])
 
+            shape = self._g.get_shape(node.output[0])
+            dtype = self._g.get_dtype(node.output[0])
```
nbcsm: please add tests to cover these cases.

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -380,7 +385,7 @@ def _mul_handler(self, trans, node):
                 multiplier_input_id = i
                 multiplier_input_node = input_node
 
-        if not multiplier_input_node.is_const():
+        if multiplier_input_node is None or not multiplier_input_node.is_const():
```
pengwa: thanks for fixing that. add a comment for better understanding?

zhijxu-MS: done

tf2onnx/optimizer/transpose_optimizer.py
zhijxu-MS: 1 transpose node is graph output has been covered by test_transpose_leaky_relu
2 identity is graph output is covered by new-added test case "test_transpose_with_identity"

PR #408, opened 2019-03-26
fix bug

lucienwang1009:
@guschmue 
1. fetch actual inputs from frozen graph when loading from saved model. 
2. a typo when replacing graph outputs.

### DIFF
--- a/tf2onnx/loader.py
+++ b/tf2onnx/loader.py
@@ -87,6 +87,11 @@ def from_saved_model(model_path, input_names, output_names):
             for _, output_tensor in sorted(outputs_tensor_info.items()):
                 outputs[output_tensor.name] = sess.graph.get_tensor_by_name(output_tensor.name)
         frozen_graph = freeze_session(sess, output_names=list(outputs.keys()))
+        frozen_inputs = []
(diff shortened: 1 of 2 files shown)


nbcsm:
seems the first issue is covered in https://github.com/onnx/tensorflow-onnx/pull/397?

tf2onnx/loader.py
```diff
@@ -87,6 +87,11 @@ def from_saved_model(model_path, input_names, output_names):
             for _, output_tensor in sorted(outputs_tensor_info.items()):
                 outputs[output_tensor.name] = sess.graph.get_tensor_by_name(output_tensor.name)
         frozen_graph = freeze_session(sess, output_names=list(outputs.keys()))
+        frozen_inputs = []
+        # get inputs in frozen graph
+        for n in frozen_graph.node:
+            if n.name in inputs.keys():
+                frozen_inputs.append(n.name)
```
nbcsm: maybe print a warning to print inputs that are not part of frozen graph?

lucienwang1009:
> seems the first issue is covered in #397?

hmm. didn't notice it. will revert the change.

PR #418, opened 2019-03-27
fix transpose bug

lucienwang1009:
when transpose's output is graph output, skip it.
@zhijxu-MS Please have a look after your PR about applying optimization in unittests check in.

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -256,6 +256,9 @@ class TransposeOptimizer(GraphOptimizerBase):
     # if return value is True, then it means Transpose is handled as designed
     # otherwise, it means that we skip handling since it is not in our support set
     def _handle_nhwc_tranpose(self, trans):
+        if trans.output[0] in self._g.outputs:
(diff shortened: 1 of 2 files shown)


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -252,6 +252,9 @@ def _switch_transpose_and_node(self, node, trans):
     # if return value is True, then it means Transpose is handled as designed
     # otherwise, it means that we skip handling since it is not in our support set
     def _handle_nhwc_tranpose(self, trans):
+        if trans.output[0] in self._g.outputs:
+            log.debug("%s connects to graph outputs", trans.output[0])
+            return False
```
nbcsm: add a test to cover this case?

lucienwang1009: need f850cbe15a3a3e928274e8d94400d756846baf7f

nbcsm: yep

nbcsm: @lucienwang1009, a simple case will be enough, tks.

PR #419, opened 2019-03-27
support Select with Where of opset 9

lucienwang1009:


tf2onnx/function/select.py
```diff
@@ -13,6 +13,49 @@
 # pylint: disable=unused-argument,missing-docstring
 
 
+def select_op9(ctx, node, name, args):
+    # T output = Select(bool condition, T x, T y)
+    # T1 output = Where(bool condition, T1 x, T1 y)
+    # NOTE: current onnxruntime only register Where with T1=float
```
nbcsm: we should always follow onnx spec by default.
can rewrite_incomplete_type_support_rs6 do the cast trick below?

nbcsm:
https://github.com/onnx/tensorflow-onnx/pull/419

tf2onnx/function/select.py
lucienwang1009: Sure!

PR #433, opened 2019-04-02
fix bugs found in real model

zhijxu-MS:
1 const node's name should be unique
2 enhance shape_inference: unpack, maximum, concatv2
4  fix bug at add_padding: sometime the output shape can't be get
5 fix bug at gather's shape inference

### DIFF
--- a/tf2onnx/shape_inference.py
+++ b/tf2onnx/shape_inference.py
@@ -151,7 +151,7 @@ def infer_shape_for_node(g, node):
         shape_indices = g.get_shape(node.input[1])
         axis = node.input[2].get_tensor_value()
 
-        shape = shape_params[:axis] + shape_indices + shape_indices[axis + 1:]
+        shape = shape_params[:axis] + shape_indices + shape_params[axis + 1:]
         g.set_shape(node.output[0], shape)
         return True
 
--- a/tf2onnx/shape_inference.py
+++ b/tf2onnx/shape_inference.py
@@ -9,6 +9,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 import logging
+import numpy as np
 from onnx import onnx_pb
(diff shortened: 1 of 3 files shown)


tf2onnx/shape_inference.py
```diff
@@ -213,6 +210,40 @@ def infer_input_shapes(g, node):
...
+        axis = axis_node.get_tensor_value()
+        data_inputs = node.input[:-1]
+        input_shapes = [g.get_shape(node) for node in data_inputs]
+        # find the best shape info, concat node requires all dims equal except the concat dim
+        new_shape = input_shapes[0]
+        for tmp in input_shapes:
+            if tmp is not None:
+                if new_shape is None:
+                    new_shape = tmp
+                elif tmp.count(-1) < new_shape.count(-1):
```
nbcsm: if i understand correctly, what we really want to do is below, right?
```
if new_shape[i] == -1
    new_shape[i] = tmp[i]
```

comparing count of -1 does not seem to be right.

tf2onnx/shape_inference.py
```diff
@@ -403,3 +434,43 @@ def _find_tensorarray_write(graph, node):
...
+        if 1 in [i, j] or -1 in [i, j]:
+            continue
+        if i != j:
+            legal_flag = False
+
+    return legal_flag
+
+
+def _shape_after_broadcasting(shape_a, shape_b):
+    # expand shape data according to broadcasting rule to make them have same length
```
nbcsm: can this method meet your needs?
https://github.com/onnx/tensorflow-onnx/blob/master/tf2onnx/shape_inference.py#L355

nbcsm:
@zhijxu-MS, please resolve the conflicts.

nbcsm:
@zhijxu-MS, is this still needed?

Issue #424, opened 2019-03-27
Add missing ops that were supported in onnx-tensorflow frontend


chinhuang007:
We would like to add the following operators that were supported in onnx-tensorflow frontend

- BatchNorm (not needed; never appears in TF graphs)
- Ceil (done)
- Conv1D (done)
- GRU (already covered by rewriter)
- LogicalXor (not needed; never appears in TF graphs)
- LSTM (already covered by rewriter)
- MaxPoolwithArgmax (done)
- RNN (already covered by rewriter)
- RandomStandardNormal (already covered by rewriter)
- Selu (done)
- Softplus (done)
- Softsign (done)

guschmue:
I was wrong when I said we never call  the onnx GRU and LSTM ops - after checking, we do in some cases. We can maybe look at your test case to make sure we cover that.
RandomStandardNormal we do.
The remaining ones - sure.

pengwa:
@guschmue  I did not remember we called GRU or LSTM, which case do you mean?

pengwa:
@chinhuang007  do you mean TF has GRU/LSTM node that we should map? Can you elaborate about the mapping onnx-tensorflow frontend did for them?

guschmue:
@pengwa  - I thought the same but Nathan keeps telling me we do call it. Found this: 
https://github.com/onnx/tensorflow-onnx/blob/master/tf2onnx/rewriter/lstm_rewriter.py#L275

pengwa:
@guschmue  Yeah, that's the place we map tf control flow based RNN into a single LSTM/GRU op, so using that one is expected.

guschmue:
Yeh, I get it now.

pengwa:
about RandomStadardNormal, @chinhuang007  please kindly check https://github.com/onnx/tensorflow-onnx/blob/b2d3f9771887da56490e009093a3a97701700e27/tf2onnx/tfonnx.py#L1955 because there we rewriter the graph before doing the mapping, and we callied it Pre-Rewriter which might can save you guys some time :) .

chinhuang007:
@pengwa Thanks for your comments. We simply compared with the mapping of ops in tfonnx.py, and have not looked into the rewriters. So I guess the ops like GRU, LSTM, RandomStadardNormal will be handled by some rewriters during the graph pre-processing?

pengwa:
@chinhuang007  yes, it is. :) we have bunches of testing for that.

chinhuang007:
[...]

PR #435, opened 2019-04-02
add direct ops from onnx-tf

chinhuang007:
Add the following direct ops from onnx-tf
1. ceil
2. softplus
3. softsign

This is to partially address issue
https://github.com/onnx/tensorflow-onnx/issues/424

The other ops on the list require individual op methods
and will be covered subsequent PRs.

This is my first PR here so I might have missed something
important, like additional tests or docs. Just let me know.
Thanks.


guschmue:
Thanks Chin!

chinhuang007:
Thanks for the review!

PR #439, opened 2019-04-03
make ut happy for older tf versions

guschmue:


chinhuang007:
Did not know the older version doesn't have tf.math.xx. Thanks for the fix.

PR #443, opened 2019-04-06
reorg handlers in tfonnx.py into category based files

guschmue:
proposal so folks can take a look - don't merge


tjingrant:
Looks great to me! Thanks!

chinhuang007:
Makes sense to group ops in files in a similar way to the folders under onnx/onnx/defs. Very nice!

tf2onnx/handler.py
```diff
@@ -0,0 +1,69 @@
...
+import collections
+import inspect
+
+# pylint: disable=unused-argument,missing-docstring,invalid-name
+
+class tf_op:
+    _OPSETS = collections.OrderedDict()
+    _MAPPING = None
+
+    def __init__(self, name, domain="onnx", **kwargs):
```
nbcsm: do we want to keep it consistent with onnx spec, use an empty string for domain onnx.
https://github.com/onnx/onnx/blob/master/onnx/common/constants.h#L12-L13

guschmue: fair, I will make it constants.ONNX_DOMAIN

tf2onnx/handler.py
```diff
@@ -0,0 +1,79 @@
...
+                m = mapping.get(domain)
+                if m:
+                    if target_opset <= m and op_map:
+                        ops_mapping.update(op_map)
+
+        tf_op._MAPPING = ops_mapping
+        return ops_mapping
+
+    @staticmethod
+    def find_op(name):
```
nbcsm: is this a must? if one knows the tf op name, he can invoke corresponding method directly.

if we decide to keep this method, maybe we should add a ```domain``` argument, to allow to filter op by domain name.

tf2onnx/custom_opsets/__init__.py
```diff
@@ -4,7 +4,3 @@
 
 from . import ms
 from .. import constants
```
nbcsm: constants is not needed?

tf2onnx/handler.py
```diff
@@ -0,0 +1,79 @@
...
+            opset = []
+            tf_op._OPSETS[self.domain] = opset
+        for k, v in inspect.getmembers(func, inspect.ismethod):
+            if k.startswith("version_"):
+                version = int(k.replace("version_", ""))
+                while version >= len(opset):
+                    opset.append({})
+                opset_dict = opset[version]
+                for name in self.name:
+                    opset_dict[name] = (v, self.kwargs)
```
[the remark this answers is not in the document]
guschmue: done

tf2onnx/handler.py
```diff
@@ -0,0 +1,79 @@
...
+                opset.append({})
+            opset_dict = opset[version]
+            opset_dict[self.name[0]] = (func, self.kwargs)
+
+    @staticmethod
+    def get_opsets():
+        return tf_op._OPSETS
+
+    @staticmethod
+    def create_mapping(max_opset, extra_opsets):
```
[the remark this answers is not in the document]
guschmue: done, I call it now max_onnx_opset_version.

tf2onnx/onnx_opset/nn.py
```diff
@@ -0,0 +1,730 @@
...
+        node.set_attr("group", i_c)
+        add_padding(ctx, node, kernel_shape, strides)
+
+        new_kernel_shape = [k_output_channels, 1, k_h, k_w]
+        conv_convert_inputs(ctx, node, with_kernel=True, new_kernel_shape=new_kernel_shape)
+
+
+@tf_op(["AvgPool", "AvgPool3D", "MaxPool", "MaxPoolV2"],
+       type_map={"AvgPool": "AveragePool", "AvgPool3D": "AveragePool",
+                 "MaxPool": "MaxPool", "MaxPoolV2": "MaxPool"})
```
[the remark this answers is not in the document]
guschmue: Yes, it looks much better your way. done.

tf2onnx/onnx_opset/controlflow.py
```diff
@@ -0,0 +1,424 @@
...
+        make_range_non_const(ctx, start, limit, delta, output, scope_name, shape, dtype)
+
+
+@tf_op(["If", "Loop", "Scan"])
+class PassThroughOp:
+    @classmethod
+    def version_7(cls, ctx, node, **kwargs):
+        pass
+
+
```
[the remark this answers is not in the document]
guschmue: not sure where I should put it - Range is more a generator and Select/Where might fit in in control flow ... maybe sit it out ?

tf2onnx/handler.py
guschmue: I renamed it to find_effective_op() - it is suppose to find the op that really goes into the graph. Today we do this a little iffy, finding it with something like handler.tf_op.find_effective_op("Conv2D")  seems to be better way.

tf2onnx/custom_opsets/__init__.py
guschmue: yeah, wonder why pylint did not bark.

PR #446, opened 2019-04-10
fix dropout bug

lucienwang1009:
1. Refine dropout matching pattern.
2. **When delete unused nodes, we will re-add PlaceholderWithDefault to keep the graph complete, but ignore its input const. These const shouldn't be deleted either!!!**

### DIFF
--- a/tests/run_pretrained_models.py
+++ b/tests/run_pretrained_models.py
@@ -242,6 +242,8 @@ class Test(object):
                 inputs[k] = v
 
         graph_def = tf2onnx.tfonnx.tf_optimize(inputs.keys(), self.output_names, graph_def, fold_const)
+        if debug:
+            utils.save_protobuf(os.path.join(TEMP_DIR, name + "_after_tf_optimize.pb"), graph_def)
         shape_override = {}
         g = tf.import_graph_def(graph_def, name='')
         with tf.Session(config=tf.ConfigProto(allow_soft_placement=True), graph=g) as sess:
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -353,6 +353,21 @@ class BackendTests(Tf2OnnxBackendTestBase):
         output_names_with_port = ["output:0"]
         self.run_test_case(feed_dict, input_names_with_port, output_names_with_port)
 
+    def test_nn_dropout(self):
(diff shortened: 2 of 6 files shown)


lucienwang1009:
Remove Dropout with ratio 1.0.

nbcsm:
@mindest for awareness, please sync with @lucienwang1009 for details.

tests/run_pretrained_models.py
```diff
@@ -242,6 +242,7 @@ def run_test(self, name, backend="caffe2", debug=False, onnx_file=None, opset=No
                 inputs[k] = v
 
         graph_def = tf2onnx.tfonnx.tf_optimize(inputs.keys(), self.output_names, graph_def, fold_const)
+        utils.save_protobuf(os.path.join(TEMP_DIR, name + "_after_tf_optimize.pb"), graph_def)
```
nbcsm: save only if it is debug mode?
it would be very slow to save large graph.

tests/test_backend.py
```diff
@@ -353,6 +353,21 @@ def test_dropout(self):
...
+        x = tf.placeholder(tf.float32, shape=x_val.shape, name="input_1")
+        x_ = tf.identity(x)
+
+        fc1 = tf.nn.dropout(x_, keep_prob)
+
+        _ = tf.identity(fc1, name="output")
+        feed_dict = {"input_1:0": x_val}
+        input_names_with_port = ["input_1:0"]
+        output_names_with_port = ["output:0"]
+        self.run_test_case(feed_dict, input_names_with_port, output_names_with_port, constant_fold=False)
```
nbcsm: please add a few comments for constant_fold=False

tf2onnx/tfonnx.py
```diff
@@ -216,6 +216,12 @@ def rewrite_dropout(g, ops):
         for n in set(match.get_nodes()):
             g.remove_node(n.name)
 
+    # remove dropout if its ratio is 1.0
+    for node in g.get_nodes():
```
nbcsm: no need to go through all nodes, we can get all dropout node names easily.

lucienwang1009: don't want to mix these two (actually) rewriter up.

nbcsm: make sense.

PR #450, opened 2019-04-12
remove travis config

nbcsm:


nbcsm:
@guschmue, i assume we don't need travis anymore, right?

Issue #381, opened 2019-03-18
TF 1.13.1.  test_backend.py failed 


pengwa:
ERROR: test_leaky_relu (__main__.BackendTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests\test_backend.py", line 741, in test_leaky_relu
    self._run_test_case([_OUTPUT], {_INPUT: x_val})
  File "tests\test_backend.py", line 95, in _run_test_case
    return self.run_test_case(feed_dict, [], output_names_with_port, **kwargs)
  File "C:\last\tensorflow-onnx\tests\backend_test_base.py", line 142, in run_test_case
    target=self.config.target, **process_args)
  File "C:\Program Files\Python3.6\lib\site-packages\tf2onnx-0.5.0-py3.6.egg\tf2onnx\tfonnx.py", line 2477, in process_tf_graph
    mapped_op, unmapped_op = tensorflow_onnx_mapping(g, continue_on_error, custom_op_handlers)
  File "C:\Program Files\Python3.6\lib\site-packages\tf2onnx-0.5.0-py3.6.egg\tf2onnx\tfonnx.py", line 2289, in tensorflow_onnx_mapping
    raise ValueError("tensorflow op " + op + " is not supported")
ValueError: tensorflow op LeakyRelu is not supported

the original TF graph looks as below, this should be different with the one generated in older tf versions. 

![image](https://user-images.githubusercontent.com/10530022/54513894-ad6af980-4993-11e9-96e4-fc2b1e3703f9.png)

pengwa:
@mindest  is working on the fix.

nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/381

PR #452, opened 2019-04-12
fix ci build errors in tf 1.13

mindest:
Two main errors when upgraded to tf 1.13 are about `LeakyRelu` and `dropout`.

- add tf op `LeakyRelu` and modify corresponding backend test case `test_leaky_relu`
- add backend test case for `nn.dropout` with `rate` given (new feature in tf 1.13)  instead of `keep_prob`
- add `test_graph` test case for `nn.dropout` when tf version is 1.13 or newer
- add `check_tf_max_version` in `tests/common.py`

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -124,11 +124,11 @@ def _append_message(reason, message):
     return reason
 
 
-def check_tf_max_version(max_required_version, message=""):
+def check_tf_max_version(max_accepted_version, message=""):
     """ Skip if tf_version > max_required_version """
     config = get_test_config()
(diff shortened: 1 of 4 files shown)


nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/381

nbcsm:
hmm, seems there is some trouble with tf 1.5 for leakyrelu
https://dev.azure.com/tensorflow-onnx/tensorflow-onnx/_build/results?buildId=2703

@mindest, could you help to fix it please? thanks.

mindest:
working on it, will fix asap @nbcsm

Issue #448, opened 2019-04-11
refine logging


nbcsm:

nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/437

nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/437

PR #453, opened 2019-04-15
refine logging infrastructure

nbcsm:
this PR is for logging infrastructure, refine logging message will be in future PRs.

- unify logging level setup

for converter from command line, default level is INFO. 
for test, default level is WARNING.
user can pass argument ```--verbose``` to override logging level. ```-v``` is VERBOSE, ```-vv``` is DEBUG.

for invocation via script, respect caller's logging level
user can call ```tf2onnx.logging.set_level``` to set log level for tf2onnx explicitly.

- debug mode

introduce debug mode formally.
converter may do extra work under debug mode. e.g. dump intermediate tf/onnx graph, print shape information. 
to avoid pass a debug flag argument down to a deep function call chains, use global flag instead.

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -30,7 +30,7 @@ from tf2onnx.rewriter import *  # pylint: disable=wildcard-import
 from tf2onnx.shape_inference import infer_shape_for_graph
 from tf2onnx.utils import port_name
 
-logging.basicConfig(level=logging.INFO)
+
 log = logging.getLogger("tf2onnx")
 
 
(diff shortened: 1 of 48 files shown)


nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/448

tf2onnx/tfonnx.py
```diff
@@ -670,7 +668,7 @@ def process_tf_graph(tf_graph, continue_on_error=False, verbose=False, target=No
         Args:
             tf_graph: tensorflow graph
             continue_on_error: if an op can't be processed (aka there is no mapping), continue
-            verbose: print summary stats
+            verbose: print summary stats (deprecated)
```
nbcsm: @guschmue, please let me know if you have any concern with this one.

nbcsm:
thanks

PR #454, opened 2019-04-15
add an utility to save pretrained model

lucienwang1009:


tools/save_pretrained_model.py
```diff
@@ -0,0 +1,83 @@
...
+        for inp, value in feeds.items():
+            if isinstance(inp, str):
+                inputs[inp] = value
+            else:
+                inputs[inp.name] = value
+        np.save(inputs_path, inputs)
+        print("Saved inputs to {}".format(inputs_path))
+
+        # save graph and weights
+        from tensorflow.saved_model import simple_save
```
nbcsm: is there any tf version restriction?

tools/save_pretrained_model.py
```diff
@@ -0,0 +1,83 @@
...
+def save_pretrained_model(sess, outputs, feeds, out_dir, model_name="pretrained"):
+    """Save pretrained model and config"""
+    try:
+        import os
+        to_onnx_path = "{}/to_onnx".format(out_dir)
+        if not os.path.isdir(to_onnx_path):
+            os.makedirs(to_onnx_path)
+        saved_model = "{}/saved_model".format(to_onnx_path)
+        inputs_path = "{}/inputs.npy".format(to_onnx_path)
+        pretrained_model_yaml_path = "{}/pretrained.yaml".format(to_onnx_path)
```
nbcsm: let's also collect tf version, os, python distribution (cpython or conda) and version, installed pip packages (dump ```pip freeze --all``` to requirements.txt).

tools/save_pretrained_model.py
```diff
@@ -0,0 +1,83 @@
...
+
+        print("===============Save Frozen Graph========================")
```
nbcsm: it is saved model now. :-)

tools/save_pretrained_model.py
```diff
@@ -0,0 +1,83 @@
...
+    z = tf.matmul(x, y)
+    w = tf.get_variable("weight", [5, 10], dtype=tf.float32)
+    init = tf.global_variables_initializer()
+    outputs = [z + w]
+    feeds = {x: x_val, y: y_val}
+    with tf.Session() as sess:
+        sess.run(init)
+        sess.run(outputs, feeds)
+        # NOTE: Put below snippet after the LAST testing step
+        save_pretrained_model(sess, outputs, feeds, "./tests", model_name="test")
```
nbcsm: what if user call it multiple times?

tools/save_pretrained_model.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def save_pretrained_model(sess, outputs, feeds, out_dir, model_name="pretrained"):
+    """Save pretrained model and config"""
+    try:
+        import os
+        to_onnx_path = "{}/to_onnx".format(out_dir)
+        if not os.path.isdir(to_onnx_path):
+            os.makedirs(to_onnx_path)
+        saved_model = "{}/saved_model".format(to_onnx_path)
```
nbcsm: make sure all save logic can work correctly (overwrite) if a target file already exists.

tools/save_pretrained_model.py
lucienwang1009: at least tf1.9 can work, let me double check it.

tools/save_pretrained_model.py
nbcsm: add an ```epoch="latest"``` argument? so we only save once for each epoch

Issue #447, opened 2019-04-11
 get different inference result with origin tensorflow model


PeterWang1986:
Hi all,
Recently I tried to convert my tensorflow model to onnx format, and take onnxruntime as backend, but I got the different inference result. I'm not sure is there anything I did wrong ?

You can find much more information from here:
1. tensorflow pb file: https://pan.baidu.com/s/1a5itDj6BQFcE9or1wWXY6w

2. convert command:
` python -m tf2onnx.convert --input=CSRNet.pb --inputs=Reshape:0[1,512,512,3] --outputs=backend/density_map/BiasAdd:0 --output=model_v1.onnx --opset=9`
I can see sys info:
>  **using tensorflow=1.13.1, onnx=1.4.1, opset=9, tfonnx=1.5.0/757ed9**

3. using onnxruntime as backend to call this onnx model, get 490.95. But when I use tensorflow to execute pb file, I can get 494.22.

4. I also tried onnx-tensorflow to convert my pb file, and then call it with onnxruntime, I can get the same result with tensorflow. Here is my convert script:
`import tensorflow as tf

from tensorflow.core.framework import graph_pb2

from onnx_tf.frontend import tensorflow_graph_to_onnx_model
from onnx_tf.pb_wrapper import TensorflowGraph

graph_def = graph_pb2.GraphDef()
with open("CSRNet.pb", "rb") as f:  # load tf graph def
    graph_def.ParseFromString(f.read())
    output = TensorflowGraph.get_output_node_names(graph_def) # get output node names

    # convert tf graph to onnx model
    # opset=0 will use latest onnx op version,
    # but currently onnx-tensorflow not support MaxPool under version 10
    model = tensorflow_graph_to_onnx_model(graph_def, output, opset=9)

with open("model_v2.onnx", 'wb') as f:
    f.write(model.SerializeToString())`

If you need much info, please tell me. Thanks for your help.

nbcsm:
thanks for reporting.
it would be helpful to share your onnx models (both onnx-tensorflow and tensorflow-onnx) as well.
we will take a look and get back to you later.

sorry but it probably take a while since we are busy with opset10 currently.
if you find the root cause before that, feel free to send a PR. :-)

PeterWang1986:
hi nbcsm,
here is the onnx models:
model_v1 for tensorflow-onnx: https://pan.baidu.com/s/1JY3T-E0P7Dj--1zlkOTiqw
[...]

PR #457, opened 2019-04-15
fix mobilnetv2

guschmue:


nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/447

PR #458, opened 2019-04-16
fix bug in leaky_relu for tf_version <= 1.5

mindest:
When running UnitTest for tf 1.5, ValueError was raised in `test_leaky_relu` for `x` of int type. The following lines are missing in `leaky_relu` for old tf versions (1.4, 1.5) compared with tf_version >= 1.6

```python
    if features.dtype.is_integer:
      features = math_ops.to_float(features)
```

Therefore, another test case `test_leaky_relu_old` is added for tf_version <= 1.5, while the original test case `test_leaky_relu` now requires tf_version >= 1.6.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -799,7 +799,9 @@ class BackendTests(Tf2OnnxBackendTestBase):
 
     @skip_caffe2_backend("fails on caffe2 with dim issue")
     @check_onnxruntime_incompatibility("Mul")
+    @check_tf_min_version("1.6")
     def test_leaky_relu(self):
+        # starting from tf 1.6, leaky_relu supports `feature` x of int type
         x_types = [np.float32, np.int32, np.int64]
         for x_type in x_types:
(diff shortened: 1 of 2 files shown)


tests/test_backend.py
```diff
@@ -810,6 +812,19 @@ def test_leaky_relu(self):
                 self._run_test_case([_OUTPUT], {_INPUT: x_val})
                 tf.reset_default_graph()
 
+    @skip_caffe2_backend("fails on caffe2 with dim issue")
+    @check_onnxruntime_incompatibility("Mul")
+    @check_tf_max_version("1.5")
+    def test_leaky_relu_old(self):
```
nbcsm: on a second thought, how about have ```test_leaky_relu_int``` and ```test_leaky_relu_float```.
```test_leaky_relu_int``` requires tf > 1.5, while ```test_leaky_relu_float``` has no filter?

mindest: This seems better. I will modify accordingly.

nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/381

PR #459, opened 2019-04-16
refactor optimizer & refine logging

nbcsm:
- refactor optimizer
- refine logging library
- refine some log message, replace print with logging

### DIFF
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -12,8 +12,6 @@ from __future__ import unicode_literals
 import collections
 import copy
 import logging
-import sys
-import traceback
 import six
 import numpy as np
 
@@ -1065,27 +1063,18 @@ class GraphUtil(object):
         """Optimize the model proto, for example: eliminating all useless Transpose pairs.
 
         Returns:
-            model proto after optimization, if optimizer run successfully
-            or None, if exceptions happens
+            model proto after optimization
         """
-        try:
-            kwargs = GraphUtil.get_onnx_model_properties(onnx_model_proto)
-            graph = GraphUtil.create_graph_from_onnx_model(onnx_model_proto)
-            graph = GraphUtil.optimize_graph(graph)
-            model_proto = graph.make_model(onnx_model_proto.graph.doc_string,
-                                           graph_name=onnx_model_proto.graph.name, **kwargs)
+        kwargs = GraphUtil.get_onnx_model_properties(onnx_model_proto)
+        graph = GraphUtil.create_graph_from_onnx_model(onnx_model_proto)
+        graph = GraphUtil.optimize_graph(graph)
+        model_proto = graph.make_model(onnx_model_proto.graph.doc_string,
+                                       graph_name=onnx_model_proto.graph.name, **kwargs)
 
-            if onnx_model_proto.metadata_props:
-                metadata_props = {p.key: p.value for p in onnx_model_proto.metadata_props}
-                helper.set_model_props(model_proto, metadata_props)
-            return model_proto
-        except Exception:
-            # sometimes, onnx shape inference will fail for some reason, in this case,
-            # we just log the error, and skip the transpose optimizer.
-            type_, value_, traceback_ = sys.exc_info()
-            ex_ext = traceback.format_exception(type_, value_, traceback_)
-            print("NON-CRITICAL error in optimizer: ", ex_ext)
-            return None
+        if onnx_model_proto.metadata_props:
+            metadata_props = {p.key: p.value for p in onnx_model_proto.metadata_props}
+            helper.set_model_props(model_proto, metadata_props)
+        return model_proto
 
     @staticmethod
     def get_onnx_model_properties(onnx_model_proto):
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -21,7 +21,6 @@ from tf2onnx.utils import port_name, find_opset
 from tf2onnx import optimizer
 from tf2onnx.schemas import get_schema
 
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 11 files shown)


nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/448

nbcsm:
@zhijxu-MS, @pengwa to take a look at the optimizer part.

tf2onnx/graph.py
```diff
@@ -1065,27 +1063,18 @@ def optimize_model_proto(onnx_model_proto):
         """Optimize the model proto, for example: eliminating all useless Transpose pairs.
 
         Returns:
-            model proto after optimization, if optimizer run successfully
```
pengwa: try-catch needs to be kept otherwise it will break the conversion if onnx infer shape failed.

tf2onnx/optimizer/__init__.py
```diff
@@ -6,43 +6,51 @@
...
+            logger.warning("Failed to apply %s", name, exc_info=1)
+
+    after = graph.dump_node_statistics()
+    diff = copy.deepcopy(after)
+    diff.subtract(before)
+    diff = ["{} {} ({}->{})".format(k, str(v) if v < 0 else '+' + str(v), before.get(k, 0), after.get(k, 0))
+            for k, v in diff.most_common() if v != 0]
+    logger.info("After optimization: %s", ', '.join(diff) if diff else "no change")
+
+    return graph
```
pengwa: call graph.update_proto() before return graph.

tf2onnx/optimizer/__init__.py
```diff
@@ -6,43 +6,51 @@
...
+    """ Optimize graph, return optimized graph. No throw. """
+    logger = logging.getLogger(__name__)
+    logger.info("Optimizing ONNX model")
+
+    before = graph.dump_node_statistics()
+    opts = _get_optimizers()
+    for name, factory in opts.items():
+        try:
+            logger.verbose("Apply %s", name)
+            current = copy.deepcopy(graph)
```
pengwa: this graph can be copied completely including nodes/shapes/contained subgraphs? have your verified that?

tf2onnx/optimizer/__init__.py
```diff
@@ -6,43 +6,51 @@
...
-    # merge_duplicated_nodes should be used after transpose_opt
-    # for transpose_opt may have some trans nodes that can be merge
-    ("merge_duplicated_nodes", MergeDuplicatedNodesOptimizer),
-    ("identity_opt", IdentityOptimizer),
+    ("reduce_transpose", TransposeOptimizer),
+    ("fold_constants", ConstFoldOptimizer),
+    # merge_duplication should be used after reduce_transpose
+    # for reduce_transpose may have some trans nodes that can be merge
+    ("merge_duplication", MergeDuplicatedNodesOptimizer),
+    ("reduce_identity", IdentityOptimizer),
```
pengwa: shall we call it remove_transpose/remove_identity?

zhijxu-MS: "remove" will let others assume we can remove all trans/identity?

tf2onnx/optimizer/__init__.py
```diff
@@ -6,43 +6,51 @@
...
+            current = copy.deepcopy(graph)
+            graph = factory().optimize(current)
+        except Exception:  # pylint: disable=broad-except
+            # if current optimizer fails, continue with other optimizers
+            logger.warning("Failed to apply %s", name, exc_info=1)
+
+    after = graph.dump_node_statistics()
+    diff = copy.deepcopy(after)
+    diff.subtract(before)
+    diff = ["{} {} ({}->{})".format(k, str(v) if v < 0 else '+' + str(v), before.get(k, 0), after.get(k, 0))
```
zhijxu-MS: _print_stat_diff?

tf2onnx/optimizer/__init__.py
nbcsm: seems so.

https://docs.python.org/3/library/copy.html#copy.deepcopy

```
The difference between shallow and deep copying is only relevant for compound objects (objects that contain other objects, like lists or class instances):

A shallow copy constructs a new compound object and then (to the extent possible) inserts references into it to the objects found in the original.
A deep copy constructs a new compound object and then, recursively, inserts copies into it of the objects found in the original.
```

tf2onnx/optimizer/__init__.py
nbcsm: with current design, it is optimizer is responsible to return a valid graph with proto updated.
update_proto in added in GraphOptimizerBase.optimize, and derived class is supposed to override _optimize.

nbcsm: updated TransposeOptimizer for this.

tf2onnx/optimizer/__init__.py
nbcsm: yep, i defined _print_stat_diff in __init__ at the at first, then i realized there is circle import...
considering it is a 3 lines function, i decide to live with it.

tf2onnx/optimizer/__init__.py
nbcsm: optimize_transpose, remove_identity?

tf2onnx/graph.py
nbcsm: done

tf2onnx/optimizer/__init__.py
pengwa: but optimize_graph() is a public interface, it's different path, right?

pengwa: got it.

nbcsm:
thanks for the reivew.

PR #460, opened 2019-04-16
ONNX domain attr cannot be None with onnx package 1.2.3

wenbingl:


wenbingl:
@guschmue , can you take a look?

guschmue:
domain = None and domain = "" do a different thing in the onnx helper library:
https://github.com/onnx/onnx/blob/master/onnx/helper.py#L51
None is the correct value in onnx-1.4 and I think we made a change in the converter that requires >= onnx-1.4.

wenbingl:
onnx 1.4.1 only now? Good to know. 😆

PR #466, opened 2019-04-18
consider half_val when retrieving tensor content

ziyuang:


guschmue:
Thanks for the fix !!

Issue #470, opened 2019-04-19
tf.random_uniform fails to convert if shape is dynamic and does not come from shape operator


guschmue:
For example:
```
        x_val = np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]], dtype=np.float32)
        x = tf.placeholder(x_val.dtype, [None, 3], name=_TFINPUT)
        x_ = tf.shape(_, name="shape")[1:]
        x_ = tf.random_uniform(x_, name="rand", dtype=tf.float32)
        x_ = tf.identity(x_)
        _ = tf.identity(x_, name=_TFOUTPUT)
        self._run_test_case([_OUTPUT], {_INPUT: x_val}, check_value=False, check_shape=True)
```

PR #471, opened 2019-04-19
fix dynamic shape in tf.random_uniform for some cases

guschmue:
Fixes for https://github.com/onnx/tensorflow-onnx/issues/470


tf2onnx/rewriter/random_uniform.py
```diff
@@ -59,25 +60,50 @@ def rewrite_random_uniform_fold_const(g, ops):
...
+            # no real fill op in onnx. In general this is not going to help performance but the tensors
+            # created are expected to be small.
+
+            # tell the caller to not delete the shape node
+            to_delete.remove(shape_node)
+            # create a fill op with the shape of the value of the input tensor
+            zero = g.make_const(utils.make_name("zero"), np.zeros((), dtype=np.float32))
+            fill_node = g.make_node("Fill", inputs=[shape_node.output[0], zero.name],
+                                    shapes=shape, dtypes=[dtype])
+            func, _ = handler.tf_op.find_effective_op("Fill")
```
nbcsm: don't feel quite right about this.
what we want to do is **create an onnx node**, but we have to **create a tf node first and then convert it**...
and what if there is no corresponding "Fill" op in tf?

nbcsm: i think we do need an abstraction of onnx node(s) builder/creator.
so we can reuse common onnx node(s) creation routines among mapping functions and handle different opset versions properly inside builder/creator.

we will demonstrate an idea in Slice-10 PR later.

guschmue: its a little wired because instead of composing onnx it composes tf ops which has the side effect that one doesn't need to worry about different opset versions. At least for that part this works good. I should have checked error but thought its given that it is there.
But yes, a more formal pattern would be good so I'll watch slice-10.

Issue #468, opened 2019-04-19
opset10 support


guschmue:
**Need to implement:**
N/A

**Maybe implement:**
MatMulInteger
ConvInteger
QLinearMatMul
QLinearConv
QuantizeLinear
DequantizeLinear
StringNormalizer

**Done:**
TopK
Slice
IsInf
Upsample
Resize
NonMaxSuppression
ReverseSequence
ThresholdedRelu
MaxPool
AveragePool
Dropout

vuzelac-cadence:
@guschmue, I've took Inception V4 from a checkpoint here: `https://github.com/tensorflow/models/tree/master/research/slim`

Run the necessary scripts to obtain the frozen .pb model. Then used ONNX=1.5.0, tf2onnx 1.4.1, Tensorflow 1.13.1 to convert to ONNX.
Slice operator generated close to the end gets no begin/end/stride inputs.

guschmue:
I think the inception v4 issue has nothing to do with this issue and moved to here https://github.com/onnx/tensorflow-onnx/issues/490

ksaurabh-cadence:
@guschmue, @nbcsm 
I am facing issues with converting Tensorflow NonMaxSuppressionV4 to ONNX NonMaxSuppression. Since Tensorflow NonMaxSuppressionV3 is already supported, can V4 be also supported with minor changes? Thanks!

PR #475, opened 2019-04-24
support topk-10

zhijxu-MS:


nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/468

Issue #412, opened 2019-03-26
Python 3.7 support


prasanthpul:
Add CI to ensure tf2onnx works on Python 3.7

nbcsm:
tf2onnx depends on tensorflow, and tensorflow starts to support python 3.7 since 1.13.
we are enabling tf 1.13 currently, there are some minor compatibility issues to fix.

will enable python3.7 in CI with tf 1.13.

nbcsm:
https://github.com/onnx/tensorflow-onnx/pull/452

nbcsm:
onnx does not have python3.7 wheel either

prasanthpul:
With ONNX 1.5 it will

prasanthpul:
ONNX 1.5 WHLs are now available in test PyPi: https://test.pypi.org/project/onnx/#files (will be on official pypi soon)

nbcsm:
done, close it.

PR #478, opened 2019-04-25
support python 3.7 in ci

nbcsm:


nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/412

nbcsm:
thanks for the review.

PR #482, opened 2019-04-26
support retry for url get

nbcsm:
saw several times that pre-trained model test failed to download model due to temporary network issue.
add retry to mitigate this issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ class create_version(Command):
 
     def finalize_options(self):
         pass
-        
+
     def run(self):
         with open(os.path.join(SRC_DIR, 'version.py'), 'w') as f:
             f.write(dedent('''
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -74,11 +74,11 @@ def run(self):
...
     setup_requires=['pytest-runner'],
-    tests_require=['graphviz', 'requests', 'parameterized', 'pytest', 'pytest-cov', 'pyyaml'],
+    tests_require=['graphviz', 'parameterized', 'pytest', 'pytest-cov', 'pyyaml'],
     cmdclass=cmdclass,
     packages=find_packages(),
     author='onnx@microsoft.com',
     author_email='onnx@microsoft.com',
     url='https://github.com/onnx/tensorflow-onnx',
-    install_requires=['numpy>=1.14.1', 'onnx>=1.4.1', 'six']
+    install_requires=['numpy>=1.14.1', 'onnx>=1.4.1', 'requests', 'six']
```
nbcsm: add requests as a tf2onnx package dependency, i am considering to support an url as converter input as well.

nbcsm: @guschmue, do you have concern with this change?

guschmue: good idea

PR #484, opened 2019-04-28
disable shape inference test for scan in opsets larger than 8

lucienwang1009:


tests/common.py
```diff
@@ -174,6 +190,13 @@ def check_opset_min_version(min_required_version, message=""):
     return unittest.skipIf(config.opset < min_required_version, reason)
 
 
+def check_opset_max_version(max_required_version, message=""):
```
nbcsm: max_required_version => max_accepted_version

PR #488, opened 2019-04-29
re-enable some tests

nbcsm:
re-enable some tests which can pass now.

shrink image channel dim for pooling tests to reduce test time


nbcsm:
@zhijxu-MS to take a look.

PR #489, opened 2019-04-29
fix custom rnn shape bug in opset 10

lucienwang1009:


nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/468

PR #491, opened 2019-04-30
infer shape for tensorflow graph

lucienwang1009:
Infer shape for Tensorflow ops. Tensorflow explicitly sets shape for some ops in python code, such as Switch, Merge and TensorArrayGather. These shapes may be lost after freezing TF graph to graph_def without add_shapes=True.

To bring these shapes back, we implement our own shape inference for these control flow ops based on one assumption: **outputs of Merge op have the same shape (at least the same rank) of its inputs**.
With this assumption, our shape inference can handle:
1. in tf.cond, outputs of two branches have the same rank.
 2. in tf.while_loop, loop variables don't change their rank.


nbcsm:
@lucienwang1009, please also add tests for this change, maybe a new test file.

lucienwang1009:
Reverse our original shape inference for TF1.5, since shape inference for this version is not from cpp API and incomplete.
Add a few unittests, most of them are complex model, such as while_loop, custom_rnn, etc.

tf2onnx/utils.py
```diff
@@ -146,22 +148,40 @@ def get_tf_tensor_data(tensor):
...
             shape = get_tf_node_attr(node, "value").tensor_shape
-            dims = [int(d.size) for d in shape.dim]
+            if not shape.unknown_rank:
+                dims = [int(d.size) for d in shape.dim]
         else:
             shape = get_tf_node_attr(node, "shape")
-            dims = [d.size for d in shape.dim]
+            if not shape.unknown_rank:
+                dims = [d.size for d in shape.dim]
     except:  # pylint: disable=bare-except
```
nbcsm: what exception?

tf2onnx/utils.py
```diff
@@ -146,22 +148,40 @@ def get_tf_tensor_data(tensor):
...
-def get_shape(node):
+def get_tf_const_value(op, as_list=True):
+    make_sure(is_const_op(op), "{} isn't a const op".format(op.name))
+    value = get_tf_tensor_data(op.get_attr("value"))
+    if as_list:
+        value = value.tolist()
+    return value
+
+
+def get_tf_shape(node):
```
nbcsm: maybe it is better to not handle const and shape attr together.
maybe a ```get_tf_shape_attr``` and ```get_tf_const_shape``` (or just get shape from ```get_tf_const_value```)?

tf2onnx/tfonnx.py
```diff
@@ -700,6 +699,14 @@ def process_tf_graph(tf_graph, continue_on_error=False, verbose=False, target=No
                        "please upgrade onnx package to avoid potential conversion issue.",
                        utils.get_onnx_version(), opset)
 
+    if shape_override:
+        for name, shape in shape_override.items():
+            tf_graph.get_tensor_by_name(name).set_shape(shape)
+
+    tf_graph = infer_shape_for_graph(tf_graph)
+    # FIXME: remove it if no long supporting TF 1.5
+    tf_graph = infer_shape_for_graph_supplementary(tf_graph)
```
nbcsm: i would prefer to have a single call here.
maybe ```infer_shape(tf_graph, shape_override)```?

nbcsm: need to reload graph after apply shape_override? even all shapes are there after shape override.
will a wrongly specified shape_override break tf shape inference?

tf2onnx/shape_inference.py
```diff
@@ -9,18 +9,219 @@
...
+        1. in tf.cond, outputs of two branches have the same rank.
+        2. in tf.while_loop, loop variables don't change their rank.
+    """
+    no_shape_updated = True
+    while no_shape_updated:
+        no_shape_updated = False
+        for o in tf_graph.get_operations():
+            updated = infer_shape_for_op(o)
+            if updated:
+                no_shape_updated = True
```
nbcsm: i prefer to invoke tf infer first for whole graph (should skip when tf version > 1.5?)
then if there is still some missing shape, print a log and run legacy shape infer.

tf2onnx/shape_inference.py
```diff
@@ -9,18 +9,219 @@
...
+import tensorflow as tf
 from tf2onnx import utils
 
 # pylint: disable=logging-not-lazy,missing-docstring,consider-swap-variables
 
 
 logger = logging.getLogger(__name__)
 
+
+def tf_cpp_infer_shape(tf_graph):
```
nbcsm: maybe just call it reload_tf_graph?

tf2onnx/shape_inference.py
```diff
@@ -9,18 +9,219 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 import logging
-import numpy as np
-from onnx import onnx_pb
+import os
+from distutils.version import LooseVersion
+
```
nbcsm: revisit make_sure usage, some should throw, some may just swallow (maybe log a warning).

tests/test_loops.py
```diff
@@ -12,14 +12,32 @@
...
-from common import unittest_main
+from common import unittest_main, check_tf_min_version
 
 
 # pylint: disable=missing-docstring,invalid-name,unused-argument,using-constant-test
 
 
 class LoopTests(Tf2OnnxBackendTestBase):
 
+    @check_tf_min_version("1.9")
```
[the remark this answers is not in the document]
lucienwang1009: sess.run hang on TF < 1.9

tf2onnx/utils.py
lucienwang1009: get const shape isn't used, just handle the attr case now.

tf2onnx/utils.py
lucienwang1009: attr is absent

tf2onnx/tfonnx.py
```diff
@@ -89,11 +86,13 @@ def tflist_to_onnx(node_list, shape_override):
                 # out_idx is used by ListDiff
                 attr[a] = utils.map_tf_dtype(utils.get_tf_node_attr(node, a))
             elif a == "shape":
-                attr[a] = utils.get_shape(node)
+                shape = utils.get_tf_shape(node)
+                if shape is not None:
+                    attr[a] = shape
             elif a == "Tperm":
                 pass
             elif a == "_output_shapes":
```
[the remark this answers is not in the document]
lucienwang1009: nop, remove it.

nbcsm:
btw: can we export share for save_pretrained_model.py as well?

lucienwang1009:
> btw: can we export share for save_pretrained_model.py as well?

Sure, implement it in the next PR.

tf2onnx/utils.py
```diff
@@ -146,22 +148,43 @@ def get_tf_tensor_data(tensor):
...
             dims = [int(d.size) for d in shape.dim]
-        else:
-            shape = get_tf_node_attr(node, "shape")
-            dims = [d.size for d in shape.dim]
     except:  # pylint: disable=bare-except
         pass
     return dims
 
 
+def get_tf_tensor_shape(output):
```
nbcsm: output => tensor?

nbcsm:
merge it to unblock some issue, i will fix pylint in another pr.

tf2onnx/shape_inference.py
```diff
@@ -9,18 +9,261 @@
...
+        2. in tf.while_loop, loop variables don't change their rank.
+    """
+    shape_updated = True
+    while shape_updated:
+        shape_updated = False
+        for o in tf_graph.get_operations():
+            updated = infer_shape_for_op(o)
+            if updated:
+                shape_updated = True
+        tf_graph = reload_tf_graph(tf_graph)
```
nbcsm: should be?
```
if shape_updated:
    tf_graph = reload_tf_graph(tf_graph)
```

lucienwang1009: sure, will refine it later.

PR #501, opened 2019-05-06
enchace fold_const_optimizer

zhijxu-MS:
support cast and un-squeeze


zhijxu-MS:
onnx topk has a bug, will wait until it's  fixed.

nbcsm:
@zhijxu-MS, we may skip failing test for onnxruntime 0.4 and merge it.

PR #506, opened 2019-05-07
support multi birnn share the same input

lucienwang1009:
Support the case when multi bidirectional RNNs share the same input sequence. 
Previous birnn_rewriter pairs the forward and backward RNN only by their input. If there are several bidirectional RNNs consume the same input sequence, some of them will be mismatched. To handle this case, the output pattern of fw and bw RNNs are also considered in this PR.
Other changes:
1. put utilities concerned birnn_rewriter into rnn_utils.
2. remove script removing reverse node on inputs, it can be done by delete_unused_node later.

@pengwa @zhijxu-MS please take a look.


nbcsm:
@pengwa, @zhijxu-MS could you help to take a look please?

tf2onnx/rewriter/rnn_utils.py
```diff
@@ -263,3 +266,185 @@ def get_weights_from_const_node(g, node):
...
+            logger.debug("find fw rnn %s", input_id)
+            fw_rnns[input_id].append(n)
+
+    # when fw_rnn has same input as bw_rnn, then it may be a bi rnn
+    birnn_input = list(set(fw_rnns.keys()).intersection(bw_rnns.keys()))
+    bi_rnns = []
+    for inp in birnn_input:
+        fw_rnn = fw_rnns[inp]
+        bw_rnn = bw_rnns[inp]
+        # when bi-rnn 1st output are unused, only single birnn is support
```
pengwa: don't get what you mean

tf2onnx/rewriter/rnn_utils.py
```diff
@@ -263,3 +266,185 @@ def get_weights_from_const_node(g, node):
...
+
+def belong_to_birnn(g, fw_rnn, bw_rnn):
+    """check fw_rnn, bw_rnn are part of the same birnn"""
+    if not g.find_output_consumers(fw_rnn.output[0]) or \
+            not g.find_output_consumers(bw_rnn.output[0]):
+        logger.warning("cannot support birnn with unused 1st output, please check them by yourself.")
+        return False
+
+    is_bw_for_fw, fw_output_id = check_birnn_output_after_y(g, fw_rnn)
+    if is_bw_for_fw:
```
pengwa: why return false here?

lucienwang1009:
After discussing with @pengwa, there are three usages of outputs of bi-RNN, assuming the bi-RNN contains RNN1 and RNN2:
1. RNN1's and RNN2's outputs are both used.
2. Either RNN1 or RNN2 has the outputs.
3. Neither RNN1's or RNN2's outputs are used.
Just in case 1 and 2, we should check the output pattern: Reverse --> Transpose --> Pack/Concat to match two RNNs.

Bi-RNN usually have several outputs, e.g., LSTM has Y, Y_h, Y_c if state_is_tuple is set True. Apart from the first output, the other outputs' pattern also should be checked.

lucienwang1009:
> After discussing with @pengwa, there are three usages of outputs of bi-RNN, assuming the bi-RNN contains RNN1 and RNN2:
> 
> 1. RNN1's and RNN2's outputs are both used.
> 2. Either RNN1 or RNN2 has the outputs.
> 3. Neither RNN1's or RNN2's outputs are used.
> 
> Just in case 1 and 2, we should check the output pattern: Reverse --> Transpose --> Pack/Concat to match two RNNs.
> 
> Bi-RNN usually have several outputs, e.g., LSTM has Y, Y_h, Y_c if state_is_tuple is set True. Apart from the first output, the other outputs' pattern also should be checked.

On second thought, it's complex to consider the output pattern, as there are plenty of usage cases for bidirectional RNN. 
For instance, in the graph below, either output (y) or hidden state (y_h, y_c) of forward and backward GRU are consumed. In this case, there are even no connection between the output of FW and BW GRUs.

image

Instead, FW and BW RNN must share the same prefix in their name scope. So here we just leverage the name scope pattern to match the pair of bidirectional RNNs.

nbcsm:
@zhijxu-MS to take a look.

tf2onnx/rewriter/rnn_utils.py
```diff
@@ -263,3 +267,189 @@ def get_weights_from_const_node(g, node):
...
+            bw_rnns[input_id].append(n)
+        else:
+            logger.debug("find fw rnn %s", input_id)
+            fw_rnns[input_id].append(n)
+
+    # fw_rnn and bw_rnn must share the same input
+    birnn_input = list(set(fw_rnns.keys()).intersection(bw_rnns.keys()))
+    bi_rnns = []
+    for inp in birnn_input:
+        fw_rnn = fw_rnns[inp]
```
zhijxu-MS: could we add a comment here: if fw and bw share same input and have same name scope, then they can be combine into one birnn? or modify the name "belong_to_birnn" into "is_same_rnn_name_scope"

PR #508, opened 2019-05-09
Fix mobilenet_v2_1.4_224 failure on tf < 1.8

nbcsm:
seems there is shape inference issue for Relu6 when tf < 1.8, fix it from converter side.
add more debug logging messages.


nbcsm:
@lucienwang1009

lucienwang1009:
> @lucienwang1009

strangely, on TF 1.5, with new shape inference, no error appears.

nbcsm:
thanks for the review.

PR #511, opened 2019-05-10
Add Selu operator support

chinhuang007:
Add Selu operator support to partially address issue:
https://github.com/onnx/tensorflow-onnx/issues/424

Currently thinking a directOp handler should work. But 
please let me know if a rewriter is needed.

### DIFF
--- a/tf2onnx/onnx_opset/math.py
+++ b/tf2onnx/onnx_opset/math.py
@@ -399,3 +399,9 @@ class FloorMod:
         ctx.remove_node(node.name)
         ctx.make_node(op_type="Sub", inputs=[node.input[0], mul.output[0]],
                       name=node.name, outputs=node.output, shapes=shapes, dtypes=dtypes)
+
+@tf_op("Selu")
+class Selu:
+    @classmethod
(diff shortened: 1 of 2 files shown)


tf2onnx/onnx_opset/math.py
```diff
@@ -399,3 +399,9 @@ def version_7(cls, ctx, node, **kwargs):
         ctx.remove_node(node.name)
         ctx.make_node(op_type="Sub", inputs=[node.input[0], mul.output[0]],
                       name=node.name, outputs=node.output, shapes=shapes, dtypes=dtypes)
+
+@tf_op("Selu")
+class Selu:
+    @classmethod
+    def version_10(cls, ctx, node, **kwargs):
```
guschmue: Selu is there since opset 1:
https://github.com/onnx/onnx/blob/master/docs/Operators.md#selu
want to enable it for all opsets ?

chinhuang007: Changed to version_4, which seems to be the earliest so far.

guschmue: sold.

PR #512, opened 2019-05-13
Optimize sparse softmax cross entropy with logits

zhijxu-MS:
if label is one-hot, some optimization can be done to avoid un-necessary computation.

### DIFF
--- a/tf2onnx/onnx_opset/nn.py
+++ b/tf2onnx/onnx_opset/nn.py
@@ -707,22 +707,30 @@ class SoftmaxCrossEntropyWithLogits:
 
 
 def _make_sparse_softmax_cross_entropy_with_logits(ctx, label, logit, tf_ori_node):
-    label_dtype = ctx.get_dtype(label.output[0])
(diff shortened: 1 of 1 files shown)


tf2onnx/onnx_opset/nn.py
```diff
@@ -706,6 +706,33 @@ def version_7(cls, ctx, node, **kwargs):
         _make_softmax_cross_entropy_with_logits(ctx, labels, logits, node)
 
 
+def _make_sparse_softmax_cross_entropy_with_logits(ctx, label, logit, tf_ori_node):
```
nbcsm: does this work for SoftmaxCrossEntropyWithLogits?

zhijxu-MS: not.  SoftmaxCrossEntropyWithLogits's input "label" **not needs to be one-hot** in tensorflow then the math equation is not equal.

https://www.tensorflow.org/api_docs/python/tf/nn/softmax_cross_entropy_with_logits_v2?hl=en

PR #521, opened 2019-05-15
fix bug found in "bert"

zhijxu-MS:
1 when perm is dynamic, we can't assume it's a reversed transpose

2 can't remove node directly if it's used by others

3 dtype of slice's input should be same


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -60,16 +60,17 @@ def pre_optimize_action(self):
...
             target_shape = reshape_op.inputs[1].get_tensor_value(as_list=False)
             new_data = np.reshape(target_t, tuple(target_shape))
-            const_name = utils.port_name(utils.make_name("Const"))
+            const_name = reshape_op.output[0]
+            self._g.remove_node(reshape_op.name)
+            self._g.make_const(const_name, new_data)
 
             # point all children nodes inputs to the new node
             for output_name in reshape_op.output:
                 for child in ops:
```
guschmue: self._g.replace_all_inputs() ?

Issue #480, opened 2019-04-26
yolov2-voc conversion and Transposes


ayermolo:
When graph is converted Transpose is not inserted for MaxPool.
Output graph is:
Input --> Transpose --> Conv -->Transpose --> Add (constant folded fusedBatchNorm) --> LeakyRelu --> MaxPool --> Transpose -->Conv

ayermolo:
So I looked in to it some more. Issue seems to be triggered by rewrite_conv2d_with_pad, and some assumptions in conv_convert_inputs. More precisely it looks like it assumes that none constant inputs are already converted. Which is not true when rewrite_conv2d_with_pad happens. The conv_convert_inputs generates transpose for that input in to Conv, and then blindly sets parent, MaxPool, to NCHW. This parent.data_format = "NCHW" looks suspicious. It makes sense for when input is constant in if case, line 373, but I am not sure I understand the logic for always setting it at the end.

```
        for idx in input_indices:
            parent = node.inputs[idx]
            if node.inputs[idx].is_const():
                # if input is a constant, transpose that one
                if not parent.data_format:
                    val = parent.get_tensor_value(as_list=False)
                    parent.set_tensor_value(val.transpose(NHWC_TO_NCHW))
            else:
                # if input comes from a op, insert transpose op
                input_name = node.input[idx]
                transpose = ctx.insert_new_node_on_input(node, "Transpose", input_name)
                transpose.set_attr("perm", NHWC_TO_NCHW)
                transpose.skip_conversion = True
                shape = ctx.get_shape(input_name)
                new_shape = spatial_map(shape, NHWC_TO_NCHW)
                ctx.set_shape(transpose.output[0], new_shape)
            **parent.data_format = "NCHW"**
```

nbcsm:
@ayermolo, the bug is fixed, and your model is working fine on my devbox, please give another try.
i will close this issue.

ayermolo:
@nbcsm  Can confirm it works in my environment also. Thanks for fixing this!

PR #522, opened 2019-05-15
fix conv_convert_inputs bug

nbcsm:
for this pattern:
image

```rewrite_conv2d_with_pad``` calls ```conv_convert_inputs``` internally, while https://github.com/onnx/tensorflow-onnx/blob/e59757ba903c42fe0816b764bad22de33ae1e343/tf2onnx/onnx_opset/nn.py#L54-L73 L73 will set MaxPool data_format to NCHW, which prevents ```conv_convert_inputs``` to transpose input when dealing with MaxPool at mapping stage.


nbcsm:
https://github.com/onnx/tensorflow-onnx/issues/480

PR #532, opened 2019-05-22
if seq_len node doesn't exist, set sequence_lens of RNN empty

lucienwang1009:
if seq_len node doesn't exist, set sequence_lens of RNN empty since it's a optional input.


lucienwang1009:
@pengwa please kick off the unittest and have a look at the PR.

pengwa:
@lucienwang1009  you have the permission for that I guess. please have a try.

PR #535, opened 2019-05-23
fix bug in maxpool

zhijxu-MS:
ksize and strides are not always NHWC, actual their formats are aligned
with input data's format

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -12,6 +12,7 @@ from collections import defaultdict
 from distutils.version import LooseVersion
 from parameterized import parameterized
 import numpy as np
+import tensorflow as tf
 from tf2onnx import constants, logging, utils
 
 __all__ = [
@@ -22,6 +23,7 @@ __all__ = [
     "check_tf_min_version",
     "check_tf_max_version",
     "skip_tf_versions",
+    "skip_if_tf_cpu",
     "check_onnxruntime_min_version",
     "check_opset_min_version",
     "check_opset_max_version",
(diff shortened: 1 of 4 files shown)


zhijxu-MS:
MaxPoolWithArgmax may share same bug, will take a look in next PR

zhijxu-MS:
> MaxPoolWithArgmax may share same bug, will take a look in next PR

zhijxu-MS:
also SpaceToDepth may share same bug.

tf2onnx/onnx_opset/nn.py
```diff
@@ -301,25 +301,28 @@ def _convert(cls, ctx, node, **kwargs):
...
             ctx.remove_input(node, node.input[1])
 
+        if node.is_nhwc():
+            kernel_shape_hw = kernel_shape_tf[1:3]
+            strides_hw = strides_tf[1:3]
+        elif node.is_nchw():
+            kernel_shape_hw = kernel_shape_tf[2:4]
+            strides_hw = strides_tf[2:4]
+        else:
+            logger.warning("unexpected data format, please check it")
```
nbcsm: throw
and better error message: tf op name, attribute name, expected values, actual value.

tests/common.py
```diff
@@ -22,6 +23,7 @@
     "check_tf_min_version",
     "check_tf_max_version",
     "skip_tf_versions",
+    "skip_if_tf_cpu",
```
nbcsm: skip_tf_cpu? to align with existing names

tf2onnx/graph.py
```diff
@@ -136,9 +136,13 @@ def data_format(self, val):
         self.set_attr("data_format", val)
 
     def is_nhwc(self):
-        """Return True if node is in NCHW format."""
+        """Return True if node is in NHWC format."""
         return self.data_format == "NHWC"
 
+    def is_nchw(self):
```
nbcsm: not used? remove?

zhijxu-MS: done.

Issue #77, opened 2018-07-18
import frozen graph with error "Input 0 of node X was passed float from Y:0 incompatible with expected float_ref."


pengwa:
**Note: create this issue for anybody who might come across the similar issue in the future.** 

when I tried to convert frozen DCGAN inference model (trained with https://github.com/carpedm20/DCGAN-tensorflow), the error was thrown as below:

 ```
------------------- start handling dcgan.pb ----------------------------
change working directory to /home/pengwang/community/tensorflow
------ summarize the frozen graph, to get the inputs and outputs name
bazel-bin/tensorflow/tools/graph_transforms/summarize_graph --in_graph=/tmp/frozen/dcgan.pb
Found 2 possible inputs: (name=y, type=float(1), shape=[64,10]) (name=z, type=float(1), shape=[?,100])
No variables spotted.
Found 1 possible outputs: (name=generator/Sigmoid, op=Sigmoid)
Found 7080115 (7.08M) const parameters, 0 (0) variable parameters, and 6 control_edges
Op types used: 50 Const, 23 Identity, 8 Mul, 8 Reshape, 6 AssignSub, 6 Sub, 4 ConcatV2, 3 FusedBatchNorm, 3 Relu, 2 Add, 2 BiasAdd, 2 Conv2DBackpropInput, 2 Fill, 2 MatMul, 2 Placeholder, 1 Sigmoid
To use with tensorflow/tools/benchmark:benchmark_model try these arguments:
bazel run tensorflow/tools/benchmark:benchmark_model -- --graph=/tmp/frozen/dcgan.pb --show_flops --input_layer=y,z --input_layer_type=float,float --input_layer_shape=64,10:-1,100 --output_layer=generator/Sigmoid
------ update the inputs and outputs name to format like input_name:index
python3 /home/pengwang/community/learning/onnx/update_name_with_index.py y,z
updated input names is y:0,z:1, output names is generator/Sigmoid:0
------ start convertion, tensorflow usage require caller program must not in tensorflow root folder, so switch to current user directory with cd
using tensorflow=1.9.0-rc0, onnx=1.2.1
2018-07-17 16:10:59.166646: I tensorflow/tools/graph_transforms/transform_graph.cc:318] Applying fold_batch_norms
2018-07-17 16:10:59.194705: I tensorflow/tools/graph_transforms/transform_graph.cc:318] Applying fold_old_batch_norms
Traceback (most recent call last):
[...]

PR #536, opened 2019-05-23
substitute variable-related ops

lucienwang1009:
#77


lucienwang1009:
This PR only handles the case where variable-related ops' value depends on the referenced op:
image
In the graph above, Assign update Variable's value after the value is read. So the Assign doesn't affect the final result.
However, there are a lot of more complex cases this PR cannot cover. For instance, if Assign is independent on the referenced op (see the graph below), the Variable's value should be set before flowing down into Add op. In this case, substituting Assign op with Identity will definitely lead to a wrong result.
image

nbcsm:
if a variable is UPDATED in the inference part of a TF graph, then there is NO guarantee the frozen TF graph will behave exactly the same as the original TF graph. (variable is converted to const, not stateful anymore)

and even we can parse a TF training graph directly, there is still nothing we can do, because there is no variable concept in onnx, tensors are stateless...

so i think we should:
- **always let the converter fail on such case by default**, give meaningful message to explain why and let user to modify his graph to avoid such variable usage.
we may do this check at very early stage.

- provide an option to enable substitute_variable_related_ops
but need to print **VERY conspicuous warning message** to let user know the graph may not behave like what they want.
but i am wondering do we really need to support this?

@guschmue for comments.

nbcsm:
We may need to take a look at the usage.
If it is created explicitly by the user script, we do support it.
If it is from some common tf api, we can evaluate whether it is safe to update the graph.

nbcsm:
after discussion, we decided to NOT support this feature for now.
user should avoid to use variables in their script. 
close this pr.

PR #543, opened 2019-05-27
Enhance trans opt

zhijxu-MS:
enhance logic to replace transpose with reshape if possible

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -74,40 +74,40 @@ class TransposeOptimizer(GraphOptimizerBase):
             self._g.topological_sort(self._g.get_nodes())
 
     def post_optimize_action(self):
+        def _calculate_new_shape(graph, op):
+            input_shape = graph.get_shape(op.input[0])
+            if input_shape.count(-1) <= 1:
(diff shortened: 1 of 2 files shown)


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -74,40 +74,40 @@ def pre_optimize_action(self):
...
-                        op_name = utils.make_name("reshape")
-                        shape_name = utils.make_name(op_name)
-                        self._g.make_const(shape_name, np.array(new_shape, dtype=np.int64))
-                        self._g.remove_node(op.name)
-                        self._g.make_node("Reshape", inputs=[op.input[0], shape_name], outputs=op.output,
-                                          name=op_name)
-                    else:
-                        self._remove_useless_tranpose(op)
-        self._g.topological_sort(self._g.get_nodes())
+                if (is_nchw_transpose(op) and (input_shape[3] == 1 or (input_shape[1:3] == [1, 1])))\
```
lucienwang1009: check length of input_shape first?

zhijxu-MS: only if is_nchw_transpose(op) is true, the input_shape[3] will be executed.

PR #544, opened 2019-05-27
support strided_slice in opset10

lucienwang1009:
Slice of opset 10 moves begin, end and strides from attributes to inputs. Now we can support more cases of tf.strided_slice.

### DIFF
--- a/tf2onnx/onnx_opset/math.py
+++ b/tf2onnx/onnx_opset/math.py
@@ -54,8 +54,8 @@ class TrigOpSinceOpset9:
         pass
 
 
-def make_minmax_op(ctx, op_type, inputs, outputs,
-                   output_shapes=None, output_dtypes=None):
+def make_min_or_max_op(ctx, op_type, inputs, outputs,
+                       output_shapes=None, output_dtypes=None):
     # support more dtype
     supported_dtypes = [
         onnx_pb.TensorProto.FLOAT,
@@ -123,7 +123,7 @@ class MinMaxOp:
         shapes = node.output_shapes
         dtypes = node.output_dtypes
         ctx.remove_node(node.name)
-        make_minmax_op(ctx, node.type, node.input, node.output, shapes, dtypes)
+        make_min_or_max_op(ctx, node.type, node.input, node.output, shapes, dtypes)
 
 
 @tf_op("Softmax")
--- a/tf2onnx/onnx_opset/math.py
+++ b/tf2onnx/onnx_opset/math.py
@@ -54,62 +54,76 @@ class TrigOpSinceOpset9:
         pass
 
 
+def make_minmax_op(ctx, op_type, inputs, outputs,
+                   output_shapes=None, output_dtypes=None):
+    # support more dtype
+    supported_dtypes = [
+        onnx_pb.TensorProto.FLOAT,
+        onnx_pb.TensorProto.FLOAT16,
+        onnx_pb.TensorProto.DOUBLE
+    ]
+    target_dtype = onnx_pb.TensorProto.FLOAT
+    need_cast = False
+    cast_inputs = []
+    for inp in inputs:
+        dtype = ctx.get_dtype(inp)
+        utils.make_sure(dtype is not None, "dtype of {} is None".format(inp))
+        if dtype not in supported_dtypes:
+            cast_inp = ctx.make_node("Cast", [inp], attr={"to": target_dtype})
+            cast_inputs.append(cast_inp.output[0])
+            need_cast = True
+        else:
+            cast_inputs.append(inp)
+    node = ctx.make_node(op_type, cast_inputs, shapes=output_shapes)
+    actual_outputs = node.output
+    if need_cast:
+        origin_dtype = ctx.get_dtype(inputs[0])
+        if output_dtypes is not None:
+            origin_dtype = output_dtypes[0]
+        ctx.set_dtype(node.output[0], target_dtype)
+        cast_name = utils.make_name(node.name)
+        cast_node = ctx.insert_new_node_on_output("Cast", node.output[0], name=cast_name, to=origin_dtype)
+        ctx.set_dtype(cast_node.output[0], origin_dtype)
+        ctx.copy_shape(node.output[0], cast_node.output[0])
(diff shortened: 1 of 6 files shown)


lucienwang1009:
@zhijxu-MS please have a look at graph_builder's change.

tf2onnx/utils.py
```diff
@@ -460,6 +460,10 @@ def get_max_value(np_dtype):
     return np.iinfo(np_dtype).max
 
 
+def get_min_value(np_dtype):
```
zhijxu-MS: looks like this function is not used in this PR?

tf2onnx/onnx_opset/math.py
```diff
@@ -54,62 +54,76 @@ def version_9(cls, ctx, node, **kwargs):
         pass
 
 
+def make_minmax_op(ctx, op_type, inputs, outputs,
+                   output_shapes=None, output_dtypes=None):
+    # support more dtype
+    supported_dtypes = [
```
zhijxu-MS: why only support these dtype, could we add some comment in case when needing to add more dtype supported?

tf2onnx/onnx_opset/tensor.py
```diff
@@ -637,6 +637,161 @@ def version_4(cls, ctx, node, **kwargs):
...
+    def version_10(cls, ctx, node, **kwargs):
+        # T output = Slice(T input, Index begin, Index end, Index strides
+        #                 @int begin_mask, @int end_mask, @int ellipsis_mask
+        #                 @int shrink_axis_mask, @int new_axis_mask)
+        # T output = Slice(T input, Tind starts, Tind ends, Tind axes, Tind steps)
+        # "ends" are exclusive, "axes" and "steps" are optional, their default val are [0, ...] and 1
+        begin = node.inputs[1]
+        end = node.inputs[2]
+        strides = node.inputs[3]
+        strides_val = strides.get_tensor_value()
```
zhijxu-MS: strides is not always const?    the code below has statement "if strides.is_const():"

tf2onnx/onnx_opset/tensor.py
```diff
@@ -637,6 +637,161 @@ def version_4(cls, ctx, node, **kwargs):
...
+        if begin.is_const() and end.is_const() and strides.is_const() \
+                and np.all(strides_val == 1):
+            cls.version_4(ctx, node, **kwargs)
+
+        not_supported_attr = ["new_axis_mask"]
+        for attr_name in not_supported_attr:
+            attr = node.get_attr(attr_name)
+            if attr is not None and attr.i != 0:
+                raise ValueError("StridedSlice: attribute " + attr_name + " not supported")
+        onnx_dtype = ctx.get_dtype(node.input[1])
```
zhijxu-MS: put these two lines close to the line(line 716) where it used, so we don't need to jump too much to look at its meaning?

tf2onnx/onnx_opset/math.py
```diff
@@ -54,62 +54,76 @@ def version_9(cls, ctx, node, **kwargs):
...
+        onnx_pb.TensorProto.DOUBLE
+    ]
+    target_dtype = onnx_pb.TensorProto.FLOAT
+    need_cast = False
+    cast_inputs = []
+    for inp in inputs:
+        dtype = ctx.get_dtype(inp)
+        utils.make_sure(dtype is not None, "dtype of {} is None".format(inp))
+        if dtype not in supported_dtypes:
+            cast_inp = ctx.make_node("Cast", [inp], attr={"to": target_dtype})
```
zhijxu-MS: will it be legal in tf while not in onnx that:  dtype of begin, end and stride is double, double, float? I have seen this is not supported in onnx slice.

tf2onnx/onnx_opset/math.py
```diff
@@ -54,62 +54,76 @@ def version_9(cls, ctx, node, **kwargs):
         pass
 
 
+def make_minmax_op(ctx, op_type, inputs, outputs,
```
zhijxu-MS: the name looks confusing, maybe "make_min_or_max_op" or add a comment?

tf2onnx/onnx_opset/tensor.py
```diff
@@ -637,6 +637,161 @@ def version_4(cls, ctx, node, **kwargs):
...
+
+        inputs_map = {
+            "data": node.input[0],
+            "starts": begin.output[0],
+            "ends": end_output,
+            "steps": strides_output,
+            "axes": axes_output
+        }
+        kwargs = {**inputs_map, "outputs": node.output}
+        node = GraphBuilder(ctx).make_slice(kwargs, name=node.name, dtypes=out_dtypes, shapes=out_shapes)
```
zhijxu-MS: if needs_squeeze is true, are the out_shapes correct here?

tf2onnx/utils.py
lucienwang1009: Ideally, we should use this function, but it will overflow when casting to low precision.
So in this PR, 1e9 and -1e9 are used for max and min value for index input (begin, end and strides).
Will re-enable it after we don't need to cater for the type restriction of Max op.

tf2onnx/onnx_opset/math.py
lucienwang1009: sure

tf2onnx/onnx_opset/math.py
lucienwang1009: it's supported dytypes of ONNX op not convertor supported dtype

tf2onnx/onnx_opset/tensor.py
lucienwang1009: sure

tf2onnx/onnx_opset/tensor.py
lucienwang1009: of course, we always respect tf shape.

tf2onnx/onnx_opset/math.py
lucienwang1009: nope, it's also illegal in tf.

jeffsaremi:
This solved a problem I was encountering. thanks very much

PR #547, opened 2019-05-28
support CudnnCompatibleGRUCell

lucienwang1009:
CudnnCompatibleGRUCell only differs from GRUCell on memory gate `(h'_t = tanh(x_t * W_h + (r_t .* h_t-1) * R_h + b_{Wh})`, which can be covered by linear_before_reset attribute of GRU: https://github.com/onnx/onnx/blob/master/docs/Operators.md#GRU.

### DIFF
--- a/tests/test_cudnn_compatible_gru.py
+++ b/tests/test_cudnn_compatible_gru.py
@@ -1,0 +1,481 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT license.
+
+"""Unit Tests for gru."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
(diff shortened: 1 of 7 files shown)


tests/test_cudnn_compatible_gru.py
```diff
@@ -0,0 +1,481 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
zhijxu-MS: these test functions are same with gru's except the cell type, will it be better to merge with them? otherwise, if we want to add more test cases, them we have to add 3 function.

lucienwang1009: it's doable like contrib op test, but I suggest to keep it separated in this PR.

zhijxu-MS: got it.

Issue #490, opened 2019-04-29
Inception V4 broken


vuzelac-cadence:
I've took Inception V4 from a checkpoint here: https://github.com/tensorflow/models/tree/master/research/slim

Close to the end, there is Flattening layer implemented using `StridedSlice`. TF2ONNX 0.3.1 will properly recognize flattening and will insert `Flatten `layer.
Current TF2ONNX generates:
```
Shape->Gather->Cast->Slice->Squeeze->Cast->Unsqueeze->Concat->Cast
                                           Unsqueeze-^
```
I am not evaluating this complex translation for correctness, however, `Squeeze `above removes axis=0 from 1D data which breaks the model.

guschmue:
This is the same model that we use in our CI pipeline:
https://github.com/onnx/tensorflow-onnx/blob/master/tests/run_pretrained_models.yaml#L169 
As far I can tell its fine. If I run it manual:
```
python tests/run_pretrained_models.py --tests googlenet_v4_slim --opset 8
TensorFlow OK
Using tensorflow=1.13.1, onnx=1.4.1, tf2onnx=1.5.0/82f805
Using opset <onnx, 8>
Optimizing ONNX model
After optimization: Identity -2 (2->0), Transpose -86 (336->250)
To_ONNX, OK
Run_ONNX OK
Results: OK
RESULT: 0 failed of 1, backend=onnxruntime
```
results look good. (the rtol in the yaml file is 0.1 but I changed it to 0.01 in my tree and will send a PR for that).
We can take a look at the graph.

vuzelac-cadence:
Thanks for a quick response.
Maybe I am missing something, but your script is downloading officially non-existent model: 
`https://github.com/tensorflow/models/tree/master/research/slim` 

This README lists the checkpoint files only, right ?

When freezing from a checkpoint using provided instructions (using TF1.13.1) the frozen model looks very different. I assume the model you use was frozen 2 years back using Tensorfflow version < 1.0.0.

guschmue:
Ah, we use the frozen.pb from tensorflow team which is pretty old. 
I can try tf-1.13, see how that works.

vuzelac-cadence:
Hi @guschmue , please let me know if I can help with anything. We have some customers blocked on Inception V4, and would like to enable them ASAP.

zhijxu-MS:
hi @vuzelac-cadence could you please share with me the frozen pb file in your env?

vuzelac-cadence:
[...]

PR #554, opened 2019-05-30
fix flatten rewriter bug

lucienwang1009:
address #490


lucienwang1009:
@guschmue can you take a look, please? it's a minor change.

PR #557, opened 2019-05-31
remove const from graph inputs when IR>=4

lucienwang1009:
As we discussed before, remove const from graph inputs to conform to ONNX new IR version.


tf2onnx/graph.py
```diff
@@ -866,7 +867,11 @@ def make_graph(self, doc, graph_name="tf2onnx"):
             initializers.append(tensor)
 
         # create input_tensor_values
-        input_ids = [op.output[0] for op in placeholder_ops + const_ops]
+        input_ids = [op.output[0] for op in placeholder_ops]
+        # onnx below IR 4 requires initializer should be in inputs
```
guschmue: Checking the IR version is correct, BUT since the onnx package is 1.5 it will have that IR version of 4 burned in it would give you the wrong behavior if you use say opset 8 since opset 8 come with IR version 3 (something iffy with the onnx versioning here).

The IR version was updated with opset9 (I think), maybe checking for opset would be better.

guschmue: Example: you create a model with opset 8 (that came with IR version 3). The model say opset8/IR4 since we use onnx-1.5. This actually only works because runtime is checking on the IR version. 
But since opset8 came with IR3 the runtime may expect the initializers in the input.

lucienwang1009: From https://github.com/onnx/onnx/blob/master/docs/Versioning.md, it seems IR version is largely independent on opset version. It only defines the abstract for graphs and operators and the concrete format that represents them, which doesn't involve the specification of operations.
And all tests are green, indicating that runtime is also happy with the different combinations of IR version and opset version, even if they aren't from the same onnx package.

guschmue: You take out the initializes from inputs for all graphs since onnx-1.5 has it set to 4. That is not what you'd want for opset8 and while the unit test on onnxruntime is ok - that is the version that supports this but you have no idea if older versions or other runtimes work so change this please,.

lucienwang1009: OK, changed. Thank you for comments.

guschmue: super, thanks!

PR #565, opened 2019-06-04
combine and report unittest results

lucienwang1009:
Merge all unittests cross different platform, different TF version, different python version, etc.
You can find the unittest coverage report in `combine_and_report_coverage` job of unittest build.
Report of html is also available as a build artifact.

### DIFF
--- a/ci_build/azure_pipelines/onnxruntime_nightly_test.yml
+++ b/ci_build/azure_pipelines/onnxruntime_nightly_test.yml
@@ -12,3 +12,6 @@ jobs:
       - template: 'unit_test.yml'
         parameters:
             onnx_opsets: ['10', '9', '8', '7']
+    report_coverage: 'True'
+
+- template: 'templates/combine_test_coverage.yml'
--- a/ci_build/azure_pipelines/templates/job_generator.yml
+++ b/ci_build/azure_pipelines/templates/job_generator.yml
@@ -9,73 +9,83 @@ parameters:
   onnx_backends: {onnxruntime: ['0.4.0']}
   job: {}
   run_setup: 'True'
+  report_coverage: 'False'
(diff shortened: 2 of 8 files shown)


nbcsm:
seems something is wrong with the test run name?

image

lucienwang1009:
@nbcsm azure pipelines will automatically add job_name ahead of its title. 
We didn't set it before so the title was what we set in config.
Now we need these unittest jobs to run before the report job and thus assign it explicitly.

nbcsm:
can we leverage stage?
https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#stage

lucienwang1009:
@nbcsm look like fitting to our scenario, I'll give a try.

ci_build/azure_pipelines/onnxruntime_nightly_test.yml
```diff
@@ -12,3 +12,8 @@ jobs:
       - template: 'unit_test.yml'
         parameters:
             onnx_opsets: ['10', '9', '8', '7']
+    report_unittest: 'True'
+
+- template: 'templates/combine_test_coverage.yml'
+  parameters:
+    combined_jobs: ['nightly']
```
nbcsm: combined_jobs?

ci_build/azure_pipelines/templates/job_generator.yml
```diff
@@ -9,73 +9,83 @@ parameters:
   onnx_backends: {onnxruntime: ['0.4.0']}
   job: {}
   run_setup: 'True'
+  report_unittest: 'False'
```
nbcsm: rename to report_coverage?

PR #569, opened 2019-06-05
Adding a flag to disable onnx runtime optimization

jeffsaremi:


jeffsaremi:
You're right. I should introduce a command line option just for that instead of setting under --debug
I'll close this for now.

PR #570, opened 2019-06-05
Merge consts

jeffsaremi:
This change will unify all constants that have the same value for IR's >=4 
Collapsing Const nodes as such will help with the duplication-removal optimization


zhijxu-MS:
we already have an optimizer "merge duplicated nodes", it supposes to merge const with same value.
Is the opt not working in your model?

jeffsaremi:
@zhijxu-MS The current optimizer did not accomplish what this PR is doing. If there are special flags that I need to supply to make it work please let me know. Otherwise, as is it does not merge constant nodes with the same values.
@nbcsm has the details.

jeffsaremi:
NOTE: it looks like this PR has picked up other changes inadvertently. I will be re-submitting it with only the required files (which should be 2 or 3)

PR #584, opened 2019-06-10
enhance optimizer

zhijxu-MS:
1 Add loop opt
2 enhance opt logic to rerun opt if graph is changed during opt


tf2onnx/optimizer/__init__.py
```diff
@@ -37,14 +39,21 @@ def optimize_graph(graph):
...
+    while continue_flag:
+        continue_flag = False
+        for name, factory in opts.items():
+            try:
+                logger.verbose("Apply %s", name)
+                current = copy.deepcopy(graph)
+                opt = factory()
+                graph = opt.optimize(current)
+                if not continue_flag:
+                    continue_flag = opt.graph_been_opt
```
lucienwang1009: continue_flag |= opt.graph_been_opt?
also put graph_been_opt as a attribute function.

zhijxu-MS: done.

PR #585, opened 2019-06-10
A fix for Merge Duplicates optimizer to allow merging consts

jeffsaremi:
Currently the logic for Const merging inside MergeDuplicatedNodesOptimizer is based on comparing `attr` of each pair of nodes. This comparison method does not work in the case of Tensors (or Graphs) which are created individually however carry the same value.
A unit test is introduces to illustrate the issue
A `_are_attr_equal()` method is introduced to replace `n1.attr == n2.attr`
For now the Graph comparison defaults back to `==` until a proper, recursive Node comparator has been written.


tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
```diff
@@ -57,7 +59,7 @@ def _del_nodes_if_duplicated(self, nodes_group, graph):
             unprocessed_node = []
             nodes_to_process = [nodes_group[0]]
             for node in nodes_group[1:]:
-                if node.attr == nodes_to_process[0].attr:
+                if self._are_attr_equal(node.attr, nodes_to_process[0].attr):
```
zhijxu-MS: could you give some comments here why "node.attr == nodes_to_process[0].attr" is not work? thanks.

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
```diff
@@ -90,3 +92,41 @@ def _skip_node_type(node):
...
+        if len(a) != len(b):
+            return False
+        for k, v in a.items():
+            if v.type == onnx_pb.AttributeProto.FLOAT or \
+                    v.type == onnx_pb.AttributeProto.INT or \
+                    v.type == onnx_pb.AttributeProto.STRING or \
+                    v.type == onnx_pb.AttributeProto.FLOATS or \
+                    v.type == onnx_pb.AttributeProto.INTS or \
+                    v.type == onnx_pb.AttributeProto.STRINGS:
+                if v != b[k]:
```
zhijxu-MS: the onnx_pb.AttributeProto.FLOATS doesn't have  "raw_data" field?

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
```diff
@@ -90,3 +92,41 @@ def _skip_node_type(node):
...
+                    return False
+            elif v.type == onnx_pb.AttributeProto.TENSOR:
+                t1 = v.t
+                t2 = b[k].t
+                if not (t1.dims == t2.dims and t1.data_type == t2.data_type and t1.raw_data == t2.raw_data):
```
zhijxu-MS: what if t1 stores data in data_type while t2 stores data in raw_data

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
```diff
@@ -90,3 +92,41 @@ def _skip_node_type(node):
...
+                # TODO provide implementation which skips doc_string and name
+                # TODO We need an equals() method for each Node and every Proto object
+                # TODO Using that graphs (and their nodes) could be compared recursively
+                if not v == b[k]:
+                    return False
+            elif v.type == onnx_pb.AttributeProto.GRAPHS:
+                # TODO provide implementation which skips doc_string and name
+                if not v == b[k]:
+                    return False
+        return True
```
zhijxu-MS: reture false may be better? for we are not sure we have cover all attribute type.

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
```diff
@@ -90,3 +92,41 @@ def _skip_node_type(node):
...
+                    return False
+            elif v.type == onnx_pb.AttributeProto.TENSORS:
+                if len(v.ts) != len(b[k].ts):
+                    return False
+                for i in range(len(v.ts)):
+                    t1 = v.ts[i]
+                    t2 = b[k].ts[i]
+                    if not (t1.dims == t2.dims and t1.data_type == t2.data_type and t1.raw_data == t2.raw_data):
+                        return False
+            elif v.type == onnx_pb.AttributeProto.GRAPH:
```
zhijxu-MS: if "node.attr == nodes_to_process[0].attr" is not work, can we compare graph directly?

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
jeffsaremi: A Tensor attribute has more fields than just the values of it
If you use `==` to compare two tensors their names get compared as well. (the same is true if the attribute is a graph). And that's not what we want because you can create multiple tensors from the same static values. 
What we need is to compare only the tensors shape, type and raw data

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
jeffsaremi: I think the plain old `==` works fine for all types except tensor and graph

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
jeffsaremi: raw_data must almost always have the data regardless of its type. Please let me know if I'm wrong in understanding this.

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
jeffsaremi: for the time being I am comparing graphs using the `==` since that's what we have had in the code since when this optimizer was written. However a more solid solution would be first to skip comparing graph's `name` ans `string_doc` fields and then compare nodes
I might take another stab at this

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
zhijxu-MS: you mean if tensor's shape & type & raw data are same, then we treat them as duplicated even other attributes are different

zhijxu-MS: could you please give out a attribute that we can ignore? thanks a lot!

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
zhijxu-MS: i'm afrad not, tensor can decide their data in data_type field or raw_data.

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
zhijxu-MS: according to my understanding, **the plain old == is too much strict** so the optimizer don't merge node that can be merged actually even they have some different attributes, am i right?

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
zhijxu-MS: i am thinking of modifying "if node.attr == nodes_to_process[0].attr:" to "if node.attr == nodes_to_process[0].attr **or is_same_const_node()**".

any thought, jeff  and @nbcsm?

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
lucienwang1009: `make_tensor` has the option `raw`.

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
jeffsaremi: I think we should not restrict this comparison to const nodes only. If we are determining the equality of nodes based on their behavior (type, data, inputs, and outputs) then other informational fields should be stripped from this comparison

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
jeffsaremi: I don't know. What I'm saying is that we have been using `==` with no issues other than the one I described here so we should be able to continue using it when we don't have a better way of comparing them

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
jeffsaremi: I'm going to investigate this a little further and make necessary changes

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
jeffsaremi: I'm going to rewrite this section so it's a little more robust

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
jeffsaremi: @zhijxu-MS I elaborated on the solution. and moved the routines to utils.py for potential reusability. Please review when you get a chance.
Important changes are not relying on raw_data, making the overall flow a little more stable, and also adding comparison for graph attributes

tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
lucienwang1009: The current comparison looks fine to me.
It's too difficult to compare raw data with the plain number, as we barely know which form the raw data comes from. It can be from list, numpy, string, etc..

nbcsm:
@jeffsaremi, thanks for the PR.
but after reconsideration and discussion with team, it might not be suitable to merge constants by current approach.

the problem is initializer. user may provide value as session input to override existing initializer value in graph, so we cannot merge initializers even they have same value.

however, currently, due to Constant-1 type limitation, all TF Const are converted to ONNX initializer. So in optimizer, we cannot tell whether an initializer is from a TF constant (which we can merge) or from a PlaceholderWithDefault (which cannot be merged).

we will have to modify current make graph logic to make things work.
the plan is:
- for IR4 (opset >= 9), convert TF const to ONNX Constant-9, convert TF PlaceholderWithDefault to ONNX initializer.
- introduce a separated const merge optimizer to merge Constant nodes.

nbcsm:
we will make the change.
will copy your test case and make it pass.

PR #588, opened 2019-06-11
fix bug of tranpose opt with broadcasting op

zhijxu-MS:


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -315,7 +323,16 @@ def _create_transpose_pairs_before_node(self, node):
 
         # add Transpose(0, 3, 1, 2) and Transpose(0, 2, 3, 1) before each non_nhwc_trans_consumers
         for input_id, n in non_nhwc_trans_inputs:
-            nchw_node = self._g.make_node("Transpose", [input_id], attr={"perm": [0, 3, 1, 2]})
+            shape = self._g.get_shape(n.output[0])
+            if len(shape) == 4:
```
jiafatom: shape is None in my case, current this if cannot handle.
I try to use `shape = n.inputs[0].inputs[0].op.attribute[0].t.dims[:]` to extract the actual shape, this is case by case.
When I try this, it can convert to an onnx model, but running via onnxruntime actually breaks.
The reason is that:
I feel that we cannot simply add `new_shape = [1]*(4-ori_rank) + ori_shape` and then do `Transpose`. This process is not enough. This seems depending on the size of the other input of `Add` .

zhijxu-MS: broadcasting op will broadcast inputs to make sure they share same rank, and the broadcast will expand the rank as my code.

zhijxu-MS: now the code can accept input with shape info is None

jiafatom:
Another issue I want to bring: According to the logic, for 1d tensor case, after optimization, we will have a `Reshape` and `Transpose`, and the initializer is in `Reshape`. During inference, this `Transpose` is done runtime. If shape is known, can we merge these two ops? We change the initializer to the value that is transposed.

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -315,7 +325,28 @@ def _create_transpose_pairs_before_node(self, node):
...
+            shape = self._g.get_shape(input_id)
+            # if rank of n is not 4, then we need to insert a reshape op before inserting a transpose
+            # for example shape of n is [x, y], then output shape of reshape will be [1, 1, x, y]
+            if shape is None:
+                const_4 = self._g.make_const(utils.make_name("const_4"), np.array([4], np.int64)).output[0]
+                tensor_1 = onnx.helper.make_tensor("value", onnx.TensorProto.INT64, [1], [1])
+                shape_node = self._g.make_node("Shape", [input_id]).output[0]
+                rank_node = self._g.make_node("Shape", [shape_node]).output[0]
+                expand_rank = self._g.make_node("Sub", [const_4, rank_node]).output[0]
+                array_fill_1 = self._g.make_node("ConstantOfShape", [expand_rank], attr={"value": tensor_1}).output[0]
```
guschmue: ConstantOfShape is only supported with opset9 or better

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -306,6 +307,15 @@ def _create_transpose_pairs_after_node(self, node):
             self._g.replace_input(consumer, node.output[0], nhwc_node.output[0])
 
     def _create_transpose_pairs_before_node(self, node):
+        def shape_after_expand(ori_shape):
+            # according to broadcasting rule to expand shape to 4D while not tile the tensor here
+            # still count on the broadcasting op to tile the tensor
+            utils.make_sure(ori_shape.count(-1) <= 1, "shape can contain one -1 at most")
+            ori_rank = len(ori_shape)
+            utils.make_sure(ori_rank <= 4, "ONNX only supports 4D data")
```
guschmue: this error message does not describe the problem

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -315,7 +325,28 @@ def _create_transpose_pairs_before_node(self, node):
...
+            # for example shape of n is [x, y], then output shape of reshape will be [1, 1, x, y]
+            if shape is None:
+                const_4 = self._g.make_const(utils.make_name("const_4"), np.array([4], np.int64)).output[0]
+                tensor_1 = onnx.helper.make_tensor("value", onnx.TensorProto.INT64, [1], [1])
+                shape_node = self._g.make_node("Shape", [input_id]).output[0]
+                rank_node = self._g.make_node("Shape", [shape_node]).output[0]
+                expand_rank = self._g.make_node("Sub", [const_4, rank_node]).output[0]
+                array_fill_1 = self._g.make_node("ConstantOfShape", [expand_rank], attr={"value": tensor_1}).output[0]
+                new_shape = self._g.make_node("Concat", [array_fill_1, shape_node], attr={"axis": 0}).output[0]
+                reshape = self._g.make_node("Reshape", [input_id, new_shape]).output[0]
```
guschmue: This form of reshape exists since opset5. We should check opset before issuing it.

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -315,7 +325,28 @@ def _create_transpose_pairs_before_node(self, node):
...
+                input_of_new_trans = reshape
+            elif len(shape) == 4:
+                input_of_new_trans = input_id
+            else:
+                shape_4d = shape_after_expand(shape)
+                const = self._g.make_const(utils.make_name("reshape_shape"), np.array(shape_4d, np.int64)).output[0]
+                reshape = self._g.make_node("Reshape", [input_id, const]).output[0]
+                input_of_new_trans = reshape
+
+            nchw_node = self._g.make_node("Transpose", [input_of_new_trans], attr={"perm": [0, 3, 1, 2]})
```
guschmue: maybe we should use the constants defined here:
https://github.com/onnx/tensorflow-onnx/blob/master/tf2onnx/constants.py#L34

zhijxu-MS: many code lines in transpose optimizer hasn't used the predefined constants. while to avoid making this PR too complex, I will refactor the code in next PR

tf2onnx/optimizer/transpose_optimizer.py
zhijxu-MS: yeah, will add a checker.

tf2onnx/optimizer/transpose_optimizer.py
zhijxu-MS: broadcasting is supported since opset7, so it the code path executed, it means graph's opset is higher than 7

tf2onnx/optimizer/transpose_optimizer.py
zhijxu-MS: this checker seems useless here since insert transpose means it's not larger than 4D, will remove.

tf2onnx/optimizer/transpose_optimizer.py
zhijxu-MS: add support broadcasting from opset 7, so if this code path executed, it means graph's opset is larger than 5

guschmue: k

tf2onnx/optimizer/transpose_optimizer.py
guschmue: k

tf2onnx/optimizer/transpose_optimizer.py
zhijxu-MS: the checker has been added before entering the branch since there may be a case where some input shapes are known while others are not.

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -306,6 +308,15 @@ def _create_transpose_pairs_after_node(self, node):
             self._g.replace_input(consumer, node.output[0], nhwc_node.output[0])
 
     def _create_transpose_pairs_before_node(self, node):
+        def shape_after_expand(ori_shape):
+            # according to broadcasting rule to expand shape to 4D while not tile the tensor here
+            # still count on the broadcasting op to tile the tensor
+            utils.make_sure(ori_shape.count(-1) <= 1,
+                            "shape can contain one -1 at most, otherwise reshape op can't work")
```
lucienwang1009: return False?

zhijxu-MS: done

PR #590, opened 2019-06-11
Enhance BatchToSpaceND to support 3D input data

lei-Qiao:
If the input is in 3d, add an unsqueeze op to convert input into 4d, and add a reshape op to convert output into 3d.
If the input is in 4d, nothing changes.

### DIFF
--- a/tf2onnx/onnx_opset/tensor.py
+++ b/tf2onnx/onnx_opset/tensor.py
@@ -1083,13 +1083,11 @@ class BatchToSpace:
         shapes = ctx.get_shape(node.output[0])
 
         if len(ctx.get_shape(input_tensor.output[0])) == 3:
-            # add a reshape op to convert output into 3d
+            # add a squeeze op to convert output into 3d
             kwargs = {**inputs_map}
             ctx.remove_node(node.name)
             slice1 = GraphBuilder(ctx).make_slice(kwargs)
-            shape_name = utils.make_name(node.name)
-            ctx.make_const(shape_name, np.array(shapes, dtype=np.int64))
-            ctx.make_node("Reshape", [slice1, shape_name], outputs=node.output, name=node.name, dtypes=dtypes)
+            ctx.make_node("Squeeze", [slice1], {"axes": [3]}, outputs=node.output, name=node.name, dtypes=dtypes)
         else:
             kwargs = {**inputs_map, "outputs": node.output}
             ctx.remove_node(node.name)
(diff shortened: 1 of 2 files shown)


tf2onnx/onnx_opset/tensor.py
```diff
@@ -1018,43 +1018,75 @@ def version_4(cls, ctx, node, **kwargs):
...
-            else:
-                ends.append(np.iinfo(np.int32).max)
+        # insert automatically an Unsqueeze op if the input is 3d
+        if len(ctx.get_shape(input_tensor.output[0])) == 3:
+            unsqz1 = ctx.make_node("Unsqueeze", input_tensor.output, {"axes": [3]})
+            trans1 = ctx.make_node("Transpose", unsqz1.output, {"perm": [3, 0, 1, 2]})
+            reorganize_node = ctx.make_node(node.type, trans1.output, attr={"blocksize": blocksize[0]})
+            trans2 = ctx.make_node("Transpose", reorganize_node.output, {"perm": [1, 2, 3, 0]})
+
+            # implement crop logic, the data format is NHWC
```
zhijxu-MS: code to implement crop logic are same into the two branch?

if so, could we refactor the code to move this logic after the if..else; or we extract this logic to be a function?

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1072,11 +1079,21 @@ def version_1(cls, ctx, node, **kwargs):
...
+        shapes = ctx.get_shape(node.output[0])
+
+        if len(ctx.get_shape(input_tensor.output[0])) == 3:
+            # add a reshape op to convert output into 3d
+            kwargs = {**inputs_map}
+            ctx.remove_node(node.name)
+            slice1 = GraphBuilder(ctx).make_slice(kwargs)
+            shape_name = utils.make_name(node.name)
+            ctx.make_const(shape_name, np.array(shapes, dtype=np.int64))
+            ctx.make_node("Reshape", [slice1, shape_name], outputs=node.output, name=node.name, dtypes=dtypes)
```
zhijxu-MS: if squeeze used, do we need to make_const?

zhijxu-MS: for sometimes, the shapes will be [N, -1, -1, -1], and reshape's input can accept one -1 at most

lucienwang1009: shape_name isn't the output name. 
use
```
shape_const = ctx.make_const(...)
ctx.make_node("Reshape", [slice1, shape_const.output[0]], ...)
```

PR #594, opened 2019-06-14
Op ReverseV2

fuzzyBatman:
Added support for the tf ReverseV2 op.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1996,6 +1996,14 @@ class BackendTests(Tf2OnnxBackendTestBase):
         x_ = tf.reverse_v2(x, axis=[0, 1, 2])
         _ = tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: x_val})
+        tf.reset_default_graph()
+
+        x_val_shape = [5, 5, 9, 7, 8, 9]
+        x_val = np.random.randint(0, 100, x_val_shape).astype(np.float32)
+        x = tf.placeholder(tf.float32, [5, 5, 9, 7, 8, 9], name=_TFINPUT)
+        x_ = tf.reverse_v2(x, axis=[0, 1, 4, 3, 5])
+        _ = tf.identity(x_, name=_TFOUTPUT)
+        self._run_test_case([_OUTPUT], {_INPUT: x_val})
 
     @check_opset_min_version(8, "where")
     def test_where(self):
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1959,6 +1959,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
 
     @skip_opset(9, "ReverseV2")
     def test_reverse_v2(self):
+        # Tests for constant axis.
         x_val = np.array([[[[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11]],
                             [[12, 13, 14, 15], [16, 17, 18, 19], [20, 21, 22, 23]]]],
                             dtype=np.float32)
@@ -1973,6 +1974,25 @@ class BackendTests(Tf2OnnxBackendTestBase):
                             dtype=np.float32)
         x = tf.placeholder(tf.float32, [2, 3, 4], name=_TFINPUT)
         x_ = tf.reverse_v2(x, axis=[2])
+        _ = tf.identity(x_, name=_TFOUTPUT)
+        self._run_test_case([_OUTPUT], {_INPUT: x_val})
+        tf.reset_default_graph()
+
+        # Tests for len(axis) > 1
+        x_val = np.array([[[[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11]],
+                            [[12, 13, 14, 15], [16, 17, 18, 19], [20, 21, 22, 23]]]],
+                            dtype=np.float32)
+        x = tf.placeholder(tf.float32, [1, 2, 3, 4], name=_TFINPUT)
+        x_ = tf.reverse_v2(x, axis=[0, 1, 2, 3])
+        _ = tf.identity(x_, name=_TFOUTPUT)
+        self._run_test_case([_OUTPUT], {_INPUT: x_val})
+        tf.reset_default_graph()
+
+        x_val = np.array([[[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11]],
+                            [[12, 13, 14, 15], [16, 17, 18, 19], [20, 21, 22, 23]]],
+                            dtype=np.float32)
+        x = tf.placeholder(tf.float32, [2, 3, 4], name=_TFINPUT)
+        x_ = tf.reverse_v2(x, axis=[0, 1, 2])
         _ = tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: x_val})
 
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1963,10 +1963,18 @@ class BackendTests(Tf2OnnxBackendTestBase):
                             [[12, 13, 14, 15], [16, 17, 18, 19], [20, 21, 22, 23]]]],
                             dtype=np.float32)
         x = tf.placeholder(tf.float32, [1, 2, 3, 4], name=_TFINPUT)
-        x_ = tf.reverse_v2(x, axis=[0])
+        x_ = tf.reverse_v2(x, axis=[3])
         _ = tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: x_val})
-        # tf.reset_default_graph()
+        tf.reset_default_graph()
+
+        x_val = np.array([[[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11]],
+                            [[12, 13, 14, 15], [16, 17, 18, 19], [20, 21, 22, 23]]],
+                            dtype=np.float32)
+        x = tf.placeholder(tf.float32, [2, 3, 4], name=_TFINPUT)
+        x_ = tf.reverse_v2(x, axis=[2])
+        _ = tf.identity(x_, name=_TFOUTPUT)
+        self._run_test_case([_OUTPUT], {_INPUT: x_val})
 
     @check_opset_min_version(8, "where")
     def test_where(self):
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1984,7 +1984,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
                             [[12, 13, 14, 15], [16, 17, 18, 19], [20, 21, 22, 23]]]],
                             dtype=np.float32)
         x = tf.placeholder(tf.float32, [1, 2, 3, 4], name=_TFINPUT)
-        x_ = tf.reverse_v2(x, axis=[0, 1, 2, 3])
+        x_ = tf.reverse_v2(x, axis=[0, -3, 2, 3])
         _ = tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: x_val})
         tf.reset_default_graph()
@@ -1993,7 +1993,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
                             [[12, 13, 14, 15], [16, 17, 18, 19], [20, 21, 22, 23]]],
                             dtype=np.float32)
         x = tf.placeholder(tf.float32, [2, 3, 4], name=_TFINPUT)
-        x_ = tf.reverse_v2(x, axis=[0, 1, 2])
+        x_ = tf.reverse_v2(x, axis=[-3, 1, 2])
         _ = tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: x_val})
         tf.reset_default_graph()
@@ -2001,7 +2001,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
         x_val_shape = [5, 5, 9, 7, 8, 9]
         x_val = np.random.randint(0, 100, x_val_shape).astype(np.float32)
         x = tf.placeholder(tf.float32, [5, 5, 9, 7, 8, 9], name=_TFINPUT)
-        x_ = tf.reverse_v2(x, axis=[0, 1, 4, 3, 5])
+        x_ = tf.reverse_v2(x, axis=[0, 1, -2, 3, 5])
         _ = tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: x_val})
 
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1963,7 +1963,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
                             [[12, 13, 14, 15], [16, 17, 18, 19], [20, 21, 22, 23]]]],
                             dtype=np.float32)
         x = tf.placeholder(tf.float32, [1, 2, 3, 4], name=_TFINPUT)
-        x_ = tf.reverse_v2(x, axis=[1])
+        x_ = tf.reverse_v2(x, axis=[0])
         _ = tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: x_val})
         # tf.reset_default_graph()
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1957,42 +1957,40 @@ class BackendTests(Tf2OnnxBackendTestBase):
         _ = tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: x_val})
 
+
     @skip_opset(9, "ReverseV2")
-    def test_reverse_v2(self):
+    def test_reversev2_constant_axis(self):
         # Tests for constant axis.
-        x_val = np.array([[[[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11]],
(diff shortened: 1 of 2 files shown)


fuzzyBatman:
This does not support unknown Tensor shapes like `[1, None, 256, 256]`

tests/test_backend.py
```diff
@@ -1957,6 +1957,76 @@ def test_reverse_sequence_time_major(self):
...
+                            [[12, 13, 14, 15], [16, 17, 18, 19], [20, 21, 22, 23]]]],
+                            dtype=np.float32)
+        x = tf.placeholder(tf.float32, [1, 2, 3, 4], name=_TFINPUT)
+        x_ = tf.reverse_v2(x, axis=[3])
+        _ = tf.identity(x_, name=_TFOUTPUT)
+        self._run_test_case([_OUTPUT], {_INPUT: x_val})
+        tf.reset_default_graph()
+
+        # Empty axis vector.
+        x_val = np.array([[[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11]],
```
zhijxu-MS: not need to assign a same value again, since it's same as line 1963

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+    @classmethod
+    def version_9(cls, ctx, node, **kwargs):
+        raise NotImplementedError("ReverseV2 is not supported to convert in OPSET 9,"
+                                  " if possible please try using OPSET>=10 instead.")
+
+    @classmethod
+    def version_10(cls, ctx, node, **kwargs):
+
+        axes_node = node.inputs[1]
+        axes = axes_node.get_tensor_value() # Is a Python list.
```
zhijxu-MS: axes is always constant in tf?

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+@tf_op("ReverseV2")
+class ReverseV2:
+    @classmethod
+    def version_9(cls, ctx, node, **kwargs):
+        raise NotImplementedError("ReverseV2 is not supported to convert in OPSET 9,"
+                                  " if possible please try using OPSET>=10 instead.")
+
+    @classmethod
+    def version_10(cls, ctx, node, **kwargs):
+
```
zhijxu-MS: could you please add some comments about your idea to convert this op?

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+                # If axis is a constant other than zero, then
+                # the previous permutation list is used.
+                # 
+                # Else compute the required permutation list.
+                if len_axes != 1:
+                    for i, ax in enumerate(axes[::-1][1:]):
+                        curr_perm[0], curr_perm[ax] = \
+                            curr_perm[ax], curr_perm[0]
+                
+                print()
```
zhijxu-MS: remove useless line?

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+
+        axes_node = node.inputs[1]
+        axes = axes_node.get_tensor_value() # Is a Python list.
+        len_axes = len(axes)
+
+        # Store input and output parameters of the ReverseV2 node.
+        rv2_in_names = [node.input[0]]
+        rv2_out_name = node.output
+        output_shape = ctx.get_shape(node.output[0])
+        len_shape = len(output_shape)
```
zhijxu-MS: will it be better to use "output_rank" instead of len_shape?

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+    @classmethod
+    def version_10(cls, ctx, node, **kwargs):
+
+        axes_node = node.inputs[1]
+        axes = axes_node.get_tensor_value() # Is a Python list.
+        len_axes = len(axes)
+
+        # Store input and output parameters of the ReverseV2 node.
+        rv2_in_names = [node.input[0]]
+        rv2_out_name = node.output
```
zhijxu-MS: why not use "node.output" in your code directly since the following code only reading rv2_out_name's value?

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+                    "ReverseSequence",
+                    inputs=inputs[-1],
+                    op_name_scope=rv2_node_name,
+                    outputs=rs_out_name,
+                    attr={"batch_axis": 1, "time_axis": 0}
+                )
+
+                # Ensure that the dtype of seq_len is INT64.
+                seq_len_dtype = ctx.get_dtype(new_node.input[1])
+                utils.make_sure(seq_len_dtype is not None, "dtype of {} is None".format(node.input[1]))
```
zhijxu-MS: could you put the checker before the real conversion logic?

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+                curr_perm[axis], curr_perm[0] = curr_perm[0], curr_perm[axis]
+
+                # Add a Transpose node if the axis != 0 (finish first due to sort).
+                if axis != 0:
+                    # Permutation indices for the transpose node relative to IN tensor shape.
+                    new_node = ctx.make_node(
+                        "Transpose",
+                        inputs=inputs[-1],
+                        op_name_scope=rv2_node_name,
+                        outputs=None,
```
zhijxu-MS: default value is none, no need to specify it again.

tests/test_backend.py
zhijxu-MS: could you split these tests into several test functions? otherwise it's hard for us to debug if something went wrong.

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
         target_dtype = TensorProto.INT64
         if seq_len_dtype != target_dtype:
             ctx.insert_new_node_on_input(node, "Cast", node.input[1], to=target_dtype)
+
+
+@tf_op("ReverseV2")
+class ReverseV2:
+    @classmethod
+    def version_9(cls, ctx, node, **kwargs):
```
zhijxu-MS: no need to add this function, tf2onnx will find out this op is not supported below 10.

tf2onnx/onnx_opset/tensor.py
fuzzyBatman: I think so. Is it not?

tests/test_backend.py
fuzzyBatman: Alright.

zhijxu-MS: and you could use np.random to generate input data, not need to specify it.

tf2onnx/onnx_opset/tensor.py
zhijxu-MS: in tf, the op can accept tensor as input.
but it's fine the conversion function only support const, while it will be better you put a checker in the beginning of function, so the reader know the limitation of your function quickly.

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+                    inputs.append([new_node.output[0]])
+
+                # Add a Constant node (seq_len) for ReverseSequence.
+
+                # Index 1 for the shape should not return 0
+                # since the input must have rank >= 2.
+                rs_batch_size = output_shape[1] if new_node is None \
+                    else ctx.get_shape(new_node.output[0])[1]
+
+                seq_list = [output_shape[axis]] * rs_batch_size
```
lucienwang1009: need to check rs_batch_size != -1 and output_shape[axis] != -1, which mean unknown dim in our converter.

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+                seq_array = np.asarray(seq_list)
+
+                const_seq_name = rv2_node_name + '_Const' + str(i)
```
lucienwang1009: please use utils.make_name to make sure the name is unique.

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+                    outputs=rs_out_name,
+                    attr={"batch_axis": 1, "time_axis": 0}
+                )
+
+                # Ensure that the dtype of seq_len is INT64.
+                seq_len_dtype = ctx.get_dtype(new_node.input[1])
+                utils.make_sure(seq_len_dtype is not None, "dtype of {} is None".format(node.input[1]))
+                target_dtype = TensorProto.INT64
+                if seq_len_dtype != target_dtype:
+                    ctx.insert_new_node_on_input(new_node, "Cast", new_node.input[1], to=target_dtype)
```
lucienwang1009: no need to check and insert cast here, since seq_len node is created by us so we can control its dtype by
`seq_array = np.asarray(seq_list, dtype=np.int64)`

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+
+                inputs.append([new_node.output[0]])
+
+            # Additional transpose block is required if the largest
+            # value in axis is greater than zero (0).
+            if axes[-1] > 0:
```
lucienwang1009: a little confusing, just use `curr_perm == orig_perm`?

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+
+                # If axis is a constant other than zero, then
+                # the previous permutation list is used.
+                # 
+                # Else compute the required permutation list.
+                if len_axes != 1:
+                    for i, ax in enumerate(axes[::-1][1:]):
+                        curr_perm[0], curr_perm[ax] = \
+                            curr_perm[ax], curr_perm[0]
```
lucienwang1009: you can use `np.argsort(curr_perm)` and then no need to check len_axes.

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+        rv2_out_name = node.output
+        output_shape = ctx.get_shape(node.output[0])
+        len_shape = len(output_shape)
+        rv2_node_name = node.name
+
+        # Remove ReverseV2 node from graph.
+        ctx.remove_node(rv2_node_name)
+
+        # List to keep track of input names.
+        inputs = [rv2_in_names]
```
lucienwang1009: really need a list to track all inputs? It seems the last inputs is enough for us to connect the subgraph.

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,142 @@ def version_10(cls, ctx, node, **kwargs):
...
+                    )
+
+                    inputs.append([new_node.output[0]])
+
+                # Add a Constant node (seq_len) for ReverseSequence.
+
+                # Index 1 for the shape should not return 0
+                # since the input must have rank >= 2.
+                rs_batch_size = output_shape[1] if new_node is None \
+                    else ctx.get_shape(new_node.output[0])[1]
```
lucienwang1009: just use `rs_batch_size = ctx.get_shape(inputs[-1][-1])[1]`? then no need to sort axes.

tf2onnx/onnx_opset/tensor.py
fuzzyBatman: Ah yes. You mean !=

tf2onnx/onnx_opset/tensor.py
fuzzyBatman: Is there no way to account for this? Will we have to use Shape nodes?

tf2onnx/onnx_opset/tensor.py
fuzzyBatman: Ah. How do I obtain that?

fuzzyBatman: Alright. The variable name.

tf2onnx/onnx_opset/tensor.py
fuzzyBatman: I am unable to make this suggestion work. There certainly are other changes required to make this work or my implementation logic does not support this. In my current implementation, when processing for the next axis, the permutation information for the previous axis is lost. Calling `np.argsort(curr_perm)` computes the sort order for the last axis alone.

tf2onnx/onnx_opset/tensor.py
fuzzyBatman: This is confusing for me. Axes can be a tensor?

fuzzyBatman: K nvm. I just read the documentation. Missed it before. Thanks for pointing that out.

fuzzyBatman:
Updated and committed all changes except for using `np.argsort()` as suggested by @lucienwang1009

tf2onnx/onnx_opset/tensor.py
lucienwang1009: Yeah, Shape definitely works for this requirement, but there will be a lot of efforts to create the seq_len input. We can get batch size and sequence length by Shape -> Slice, and the seq_len input by Tiling sequence length by batch size times.

tests/test_backend.py
```diff
@@ -1957,6 +1957,75 @@ def test_reverse_sequence_time_major(self):
...
+
+        ## For tensors with 1 dimension and axis = [0]
+        ## RESULT: 'input' must have rank >= 2
+
+        # x_val = np.array([1, 2, 3, 4],
+        #                     dtype=np.float32)
+        # x = tf.placeholder(tf.float32, [4], name=_TFINPUT)
+        # x_ = tf.reverse_v2(x, axis=[0])
+        # _ = tf.identity(x_, name=_TFOUTPUT)
+        # self._run_test_case([_OUTPUT], {_INPUT: x_val})
```
lucienwang1009: Please remove useless code.

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1248,3 +1248,156 @@ def version_10(cls, ctx, node, **kwargs):
...
+@tf_op("ReverseV2")
+class ReverseV2:
+    @classmethod
+    def version_10(cls, ctx, node, **kwargs):
+        # T output = ReverseV2(T input, int32|int64 seq_lengths, @int seq_dim, @int batch_dim)
+        # 
+        # Implement tensorflow ReverseV2 op using multiple ReverseSequence (for each axis)
+        # and Transpose ops. We sort the axis vector at the start. Each axis can be reversed only
+        # once (in tf) and so we can compute the transpose for each axis (other than 0), feed the
+        # tensor to a ReverseSequence node and finally transpose again to get back the original shape.
```
lucienwang1009: hm.. seem you forgot to sort axes.

tests/test_backend.py
```diff
@@ -1957,6 +1957,75 @@ def test_reverse_sequence_time_major(self):
         _ = tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: x_val})
 
+
+    @skip_opset(9, "ReverseV2")
```
lucienwang1009: hm.. it doesn't work, please use `@check_opset_min_version(10, "ReverseSequence")`.

tf2onnx/onnx_opset/tensor.py
lucienwang1009: I believe the current implementation could solve most of the cases. Let's leave this as an enhancement.

lucienwang1009:
@fuzzyBatman Thanks. There are still some new minor requests. Please also make sure the pylint is pass.

tf2onnx/onnx_opset/tensor.py
fuzzyBatman: Alright

tests/test_backend.py
fuzzyBatman: Replaced `@skip_opset(9, "ReverseV2")` with `@check_opset_min_version(10, "ReverseSequence")` on all 3 test functions.

fuzzyBatman:
@lucienwang1009 Done

tf2onnx/onnx_opset/tensor.py
lucienwang1009: what I meant is you need to sort the axes in order to handle the 0th axis first.

lucienwang1009: @fuzzeBatman kindly ping.

fuzzyBatman: Ah! Sorry. I missed this notification.

fuzzyBatman: @lucienwang1009 Sorting does not affect the result. Is this for using `np.argsort` ?

lucienwang1009: Sorting will make us deal with the axis 0 first. see line 1326 comment.

fuzzyBatman: I lost track of my own code. Thanks

fuzzyBatman: Pushed the changes. Runs sort on the axis vector.

fuzzyBatman:
Nice! Thanks for the advice along the way.

PR #599, opened 2019-06-14
go through all mapping functions regardless of unsupported op

lucienwang1009:
Go through all mapping functions despite some ops aren't supported and something goes wrong during the mapping process. For continue_on_error option:
- If continue_on_error=True, ignore all exceptions, log the error message and raise nothing.
- If continue_on_error=False, ignore all exceptions, log the error message and raise the first exception caused by mapping function.

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -529,10 +529,11 @@ def rewrite_conv2d_with_pad(g, ops):
     return ops
 
 
-def tensorflow_onnx_mapping(g, continue_on_error, ops_mapping):
+def tensorflow_onnx_mapping(g, ops_mapping):
     logger.verbose("Mapping TF node to ONNX node(s)")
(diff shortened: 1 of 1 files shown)


tf2onnx/tfonnx.py
```diff
@@ -545,40 +546,39 @@ def tensorflow_onnx_mapping(g, continue_on_error, ops_mapping):
         op = node.type
         map_info = ops_mapping.get(op)
         if map_info is None:
-            if continue_on_error:
-                unmapped_op[op] += 1
-                continue
-            else:
-                raise ValueError("tensorflow op " + op + " is not supported")
+            unmapped_op[op] += 1
+            logger.error("Tensorflow op %s is not supported", op)
```
nbcsm: better to add node name?

lucienwang1009: sure

Issue #531, opened 2019-05-21
error in inferencing onnx model of bvlc_googlenet


Ason93:
1) convert pb model of bvlc_googlenet  to onnx:
python3 -m tf2onnx.convert --input ~/code/models/bvlc_googlenet/bvlc_googlenet.pb --inputs Placeholder:0[1,224,224,3] --outputs prob:0 --output ~/code/models/bvlc_googlenet/bvlc_googlenet.onnx --verbose

2)    create an inference session to begin working with onnx model and run the inference session with my selected outputs and inputs to get the predicted value
    m = rt.InferenceSession(model_path)
    input_name = m.get_inputs()[0].name
    X = np.random.random((1, 224, 224, 3)).astype(np.float32)
    results = m.run(outputs, {input_name: X})

But while inferencing onnx, there is an error:

Traceback (most recent call last):
  File "pb2onnx.py", line 180, in <module>
    onnx_results = run_onnxruntime(name=None, model_proto=None, model_path=output)
  File "pb2onnx.py", line 22, in run_onnxruntime
    m = rt.InferenceSession(model_path)
  File "/home/cxjin/.local/lib/python3.6/site-packages/onnxruntime/capi/session.py", line 29, in __init__
    self._sess.load_model(path_or_bytes)
RuntimeError: [ONNXRuntimeError] : 1 : GENERAL ERROR : Exception during initialization: /onnxruntime_src/onnxruntime/core/providers/cpu/nn/lrn.h:22 onnxruntime::LRN<T>::LRN(const onnxruntime::OpKernelInfo&) [with T = float] size_ % 2 == 1 was false.

zhijxu-MS:
could you please share with me the link to download the model pb?

Ason93:
Hello! This is my test model, bvlc_googlenet.pb. Link:https://pan.baidu.com/s/1SDb0QJViZEoy6fTnlNXsFQ, password: 7nu5









At 2019-06-13 17:08:22, "zhijxu-MS" <notifications@github.com> wrote:


could you please share with me the link to download the model pb?

lei-Qiao:
Hello, the problem caused by lrn during inference session is sovled by PR #602 . However, a new problem occurs during inference session:

[...]

PR #602, opened 2019-06-17
fix math lrn op numerical problem

lei-Qiao:
reset the attributes size and alpha, and add transpose op to convert data format to fix numerical problem
issue #531

### DIFF
--- a/tf2onnx/onnx_opset/math.py
+++ b/tf2onnx/onnx_opset/math.py
@@ -262,7 +262,7 @@ class LRN:
 
         ctx.insert_new_node_on_input(node, "Transpose", node.input[0], perm=[0, 3, 1, 2])
         op_name = utils.make_name(node.name)
-        ctx.insert_new_node_on_output("Transpose", node.output[0], perm=[0, 2, 3, 1],name=op_name)
+        ctx.insert_new_node_on_output("Transpose", node.output[0], perm=[0, 2, 3, 1], name=op_name)
 
 
 @tf_op(["MatMul", "BatchMatMul"])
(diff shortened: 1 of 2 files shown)


lei-Qiao:
This can also solve the lrn problem of issue #531 : onnxruntime::LRN::LRN(constonnxruntime::OpKernelInfo&) [with T = float] size % 2 == 1 was false.

And then a new problem of issue #531 occurs during creating an inference session,  but it is not because of lrn op.

lucienwang1009:
@lei-Qiao please also comment on that issue.

tf2onnx/onnx_opset/math.py
```diff
@@ -253,10 +253,16 @@ def version_1(cls, ctx, node, **kwargs):
...
-            size = depth_radius.i
+            size = depth_radius.i * 2 + 1
         else:
-            size = 5
+            size = 5 * 2 + 1
+
         node.set_attr("size", size)
+        node.set_attr("alpha", size * node.get_attr("alpha").f)
+
+        ctx.insert_new_node_on_input(node, "Transpose", node.input[0], perm=[0, 3, 1, 2])
```
zhijxu-MS: instead of [0, 3, 1, 2], you could use constants defined in https://github.com/onnx/tensorflow-onnx/blob/7363ef6e0fa3c587c5337035c1bd26f07a9d34f4/tf2onnx/constants.py#L33

tf2onnx/onnx_opset/math.py
```diff
@@ -253,10 +253,16 @@ def version_1(cls, ctx, node, **kwargs):
         #     output = input / (bias + alpha * sqr_sum) ** beta
         depth_radius = node.get_attr("depth_radius")
         if depth_radius:
-            size = depth_radius.i
+            size = depth_radius.i * 2 + 1
         else:
-            size = 5
+            size = 5 * 2 + 1
```
zhijxu-MS: add a comment about  why 5 is used here?

Issue #323, opened 2019-02-21
ConvTranpose depends on shape to set attribute output_shape


pengwa:
for some model, when we enable --fold_const, (e.g. enable TensorFlow constant folding), the ConvTranspose's output shape would contain -1. 

so **following logic** will get -1:

```
def convtranspose_op(ctx, node, name, args):
    # T output = Conv2DBackpropInput(int32 input_sizes, T filter, T out_backprop,
    #    @list(int) strides, @bool use_cudnn_on_gpu, @string padding, @string data_format, @list(int) dilations)
    # T Y = ConvTranspose(T X, T W, T B, @STRING auto_pad, @INTS dilations,
    #    @INT group, @INTS kernel_shape, @INTS output_shape, @INTS pads, @INTS strides)

    # Note: inputs are reversed from what one would expect.
    kernel_shape = conv_kernel_shape(ctx, node, 1)

    # ouput_shape is explicitly specified here, in this case pads values are auto generated/calculated.
    **output_shape** = ctx.get_shape(node.output[0])
    if node.is_nhwc():
        new_output_shape = [output_shape[1], output_shape[2]]
    else:
        new_output_shape = [output_shape[2], output_shape[3]]
    node.set_attr("output_shape", new_output_shape)
```
So when we get the model converted, it is not runnable in runtime like OnnxRuntime.

![image](https://user-images.githubusercontent.com/10530022/53185548-32375180-363a-11e9-90dd-61b756d19b46.png)

Running in onnxruntime, we will get error like this:

`run_onnx FAIL Method run failed due to: [ONNXRuntimeError] : 1 : GENERAL ERROR : c:\onnxruntime\onnxruntime\core\providers\cpu\nn\conv_transpose.cc:184 onnxruntime::ConvTransposeBase::ComputePadsAndOutputShape output_height >= H was false. Output height cannot be smaller than input height.`

because output_height = -1 shown as above.

zhijxu-MS:
will add an checker when setting attribute output_shape.

zhijxu-MS:
fixed in #603.

PR #603, opened 2019-06-17
Add checker when setting convtranspose’s attribute

lei-Qiao:
add checker to make sure the output shape is known, and the output h and w is not smaller then input h and w
issue #323

### DIFF
--- a/tf2onnx/onnx_opset/nn.py
+++ b/tf2onnx/onnx_opset/nn.py
@@ -231,7 +231,7 @@ class ConvTranspose:
             new_output_shape = [output_shape[2], output_shape[3]]
             input_hw = [input_shape[2], input_shape[3]]
 
-        utils.make_sure(all(i != -1 for i in new_output_shape), "output h and w need to be known")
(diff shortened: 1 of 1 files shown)


tf2onnx/onnx_opset/nn.py
```diff
@@ -220,13 +220,21 @@ def version_1(cls, ctx, node, **kwargs):
...
         # ouput_shape is explicitly specified here, in this case pads values are auto generated/calculated.
         output_shape = ctx.get_shape(node.output[0])
         if node.is_nhwc():
             new_output_shape = [output_shape[1], output_shape[2]]
+            input_hw = [input_shape[1], input_shape[2]]
         else:
             new_output_shape = [output_shape[2], output_shape[3]]
+            input_hw = [input_shape[2], input_shape[3]]
+
+        utils.make_sure(all(i != -1 for i in new_output_shape), "output h and w need to be known")
```
zhijxu-MS: utils.make_sure(new_output_shape.count(-1) <= 0, "......")

lucienwang1009:
Please add a unit test to check the converter will throw exceptions, with something like assertRaises.

PR #607, opened 2019-06-18
refatcor and fix bug

zhijxu-MS:
1 use NCHW_TO_NHWC instead of [0, 2, 3, 1]
2 use NHWC_TO_NCHW instead of [0, 3, 1, 2]
3 fix bug in trans opt with concat

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -24,7 +24,7 @@ def is_nhwc_transpose(transpose_node):
 
 def is_nchw_transpose(transpose_node):
     perm_attr = transpose_node.get_attr('perm')
-    return transpose_node.type == "Transpose" and perm_attr and perm_attr.ints == [0, 3, 1, 2]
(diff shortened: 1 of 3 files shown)


zhijxu-MS:
@guschmue I replace [0, 2, 3, 1] into NCHW_TO_NHWC and [0, 3, 1, 2] into NHWC_TO_NCHW in this PR.

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -425,7 +426,14 @@ def _identity_handler(self, trans, node):
 
     def _concat_handler(self, trans, node):
         if self._handle_node_having_branches(node):
-            node.set_attr("axis", 1)
+            perm = trans.get_attr("perm").ints
+            axis_attr = node.get_attr("axis")
```
lucienwang1009: node.get_attr_value?

zhijxu-MS: done.

PR #609, opened 2019-06-18
Transpose optimization with Squeeze/Unsqueeze/Reshape - Do Not Merge Yet

jeffsaremi:


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -101,42 +102,50 @@ def post_optimize_action(self):
...
+                if op.has_attr('perm'):
+                    perm = list(op.get_attr('perm').ints)
+                if not input_shape:
+                    continue
+                if check_transpose_replaceable_with_reshape(input_shape, perm):
+                    new_shape = _calculate_new_shape(self._g, op)
+                    # replace transpose with reshape
+                    self._g.remove_node(op.name)
+                    self._g.make_node("Reshape", [op.input[0], new_shape], name=op.name, outputs=op.output)
+                    self._g.topological_sort(self._g.get_nodes())
```
guschmue: wonder if this needs to be in the inner loop since we loop over nodes. 
I guess it's not going to happen a lot, maybe ok.

PR #610, opened 2019-06-19
math lrn op set shape after transpose

lei-Qiao:
It needs override=True to set shape. Issue #531 is fixed, the model can run.

### DIFF
--- a/tf2onnx/onnx_opset/math.py
+++ b/tf2onnx/onnx_opset/math.py
@@ -262,6 +262,7 @@ class LRN:
         node.set_attr("alpha", size * node.get_attr("alpha").f)
 
         ctx.insert_new_node_on_input(node, "Transpose", node.input[0], perm=constants.NHWC_TO_NCHW)
+        ctx.update_node_shape_dtype(node, override=True)
         op_name = utils.make_name(node.name)
         ctx.insert_new_node_on_output("Transpose", node.output[0], perm=constants.NCHW_TO_NHWC, name=op_name)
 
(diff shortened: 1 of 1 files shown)


tf2onnx/onnx_opset/math.py
```diff
@@ -262,6 +262,7 @@ def version_1(cls, ctx, node, **kwargs):
         node.set_attr("alpha", size * node.get_attr("alpha").f)
 
         ctx.insert_new_node_on_input(node, "Transpose", node.input[0], perm=constants.NHWC_TO_NCHW)
+        ctx.update_node_shape_dtype(node, override=True)
         op_name = utils.make_name(node.name)
         ctx.insert_new_node_on_output("Transpose", node.output[0], perm=constants.NCHW_TO_NHWC, name=op_name)
```
lucienwang1009: also set the shape for transpose node in case onnx shape inference doesn't work.

lei-Qiao: I add `ctx.set_shape(node.output[0], new_shape)` before 
`ctx.update_node_shape_dtype(node, override=True)`. 
new_shape is NHWC_TO_NCHW(input_shape).
This can still work when I delete ctx.update_node_shape_dtype

lucienwang1009: hm... I mean when we do the mapping from one node to a set of nodes, we should retain their input and output information, including dtypes and shapes. In this case, the output of the last transpose node must share the same shape of the origin node's output. So we need to set its shape explicitly to be the origin output's.

lei-Qiao: I set shape and dtype of second transpose node:
`ctx.insert_new_node_on_output("Transpose", node.output[0], perm=constants.NCHW_TO_NHWC, name=op_name, shapes=shapes, dtypes=dtypes)` ,
and 
` shapes = node.output_shapes[0]`
 `dtypes = node.output_dtypes[0]`

PR #612, opened 2019-06-19
gen doc and sample support status

tjingrant:


tjingrant:
@guschmue here you go.
To use, `python gen_doc.py [path to the doc file you want generated]`.

guschmue:
cool, thanks! Let me see if we can call this from CI.

guschmue:
Can you make pylint happy ?
```
tools/gen_doc.py:35:0: C0305: Trailing newlines (trailing-newlines)
tools/gen_doc.py:1:0: C0111: Missing module docstring (missing-docstring)
tools/gen_doc.py:3:0: C0411: standard import "from collections import OrderedDict" should be placed before "from tf2onnx.handler import tf_op" (wrong-import-order)
```

lucienwang1009:
Can we ignore the PassThroughOp? They are not TF but ONNX ops, just a workaround for our rewriters.

tjingrant:
@lucienwang1009 can you explain what `PassThroughOp` mean and how to identify them?

lucienwang1009:
We rewrite some tf subgraphs with particular pattern, such as RandomNormal, to onnx ops before performing tf-to-onnx mapping. Those rewrited onnx ops are expected to keep intact during the late mapping, so we bring in the **PassThroughOp** as a workaround.
@guschmue is it possible to use `skip_conversion` instead of the `PassThroughOp` to get around the issue? If so, we can remove all `PassThroughOp` ops, right?

guschmue:
I think PassThroughOp  doesn't show up in the list because it is not registered as a handler.
But in general, Tian's tool dumps the handlers and longer term we need to find a way to also include the rewriters. For that we'd need something like a registration with some meta data to describe what that rewriter can do. I'd deal with those in a separate step.

Issue #614, opened 2019-06-20
A problem with the function: match_ops


JiayingGaoo:
I wrote a Gemm rewriter. According to the definition of Gemm, totally four candidate patterns are considered. Each of pattern is defined according to calculation rule, listed as below: 
pattern 0: alpha*(A*B) + beta*(C);
pattern 1: (A*B) + beta*(C);
pattern 2: (A*B) + beta*(C);
pattern 3: (A*B) + C
Where, A, B and C are the three inputs, alpha and beta are the two attributes.

There is a issue in case of pattern 0. When allow_reorder=True, the graph in the figure 1 below cannot be matched.
![cannot match pattern0](https://user-images.githubusercontent.com/32807966/59824075-b5f01900-9362-11e9-890f-d33d94d471c9.jpg)
Figure 1

If I change the sequence of the two Mul ops, then the graph will be successfully matched for rewriter, refers to figure 2:
![Successful rewriter](https://user-images.githubusercontent.com/32807966/59823840-0ca92300-9362-11e9-97a8-73b26cf252cd.jpg)
Figure 2

Therefore, I suspect that may be something wrong with the function match_ops when allow_reorder is set as true?

692307:
In my gemm_rewriter.py file, the 4 patterns are defined as:

    # pattern0: alpha*A*B + beta*C
    pattern0 = \
        OpTypePattern('Add', name='add', inputs=[
            OpTypePattern('Mul', name='mul1', inputs=[
                OpTypePattern('Const', name='alpha'),
                OpTypePattern('MatMul', name='matmul', inputs=[
                    OpTypePattern('*', name='A'),
                    OpTypePattern('*', name='B'),
                ]),
            ]),
            OpTypePattern('Mul', name='mul2', inputs=[
                OpTypePattern('Const', name='beta'),
                OpTypePattern('*', name='C'),
            ])
        ])

    # pattern1: alpha*A*B + C
    pattern1 = \
        OpTypePattern('Add', name='add', inputs=[
            OpTypePattern('Mul', name='mul1', inputs=[
                OpTypePattern('MatMul', name='matmul', inputs=[
                    OpTypePattern('*', name='A'),
                    OpTypePattern('*', name='B'),
                ]),
                OpTypePattern('Const', name='alpha')
            ]),
            OpTypePattern('*', name='C'),
        ])

    # pattern2: A*B + beta*C
    pattern2 = \
        OpTypePattern('Add', name='add', inputs=[
[...]

PR #615, opened 2019-06-20
Fix bug and refactor graph matcher

zhijxu-MS:
1 refactor graph_matcher: let readed know when op is matched with pattern more easier

2 refactor _match_pattern: make _match_pattern, the recursive function, not to interact with sub call to _match_pattern by data member _match_result which is hard to debug.

3 **fix bug** in allow_reorder related logic: fix bug mentioned in https://github.com/onnx/tensorflow-onnx/issues/614


zhijxu-MS:
@guschmue @nbcsm @lucienwang1009  these functions used frequently, please help to review it.

nbcsm:
please add a test to cover such case.

zhijxu-MS:
> please add a test to cover such case.

test is added in test_backend now.

PR #616, opened 2019-06-21
map Const to Constant instead of initializer for opset>=9

lucienwang1009:
After opset 9, onnx's Constant op supports more dtypes. No longer need to use initializers as a workaround.

### DIFF
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -141,7 +141,7 @@ class Node(object):
 
     def is_const(self):
         """Return True if node is a constant."""
-        return self.type in ["Const", "ConstV2"]
+        return self.type in ["Const", "ConstV2", "Constant"]
 
     def is_graph_input(self):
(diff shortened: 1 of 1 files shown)


nbcsm:
@guschmue to take a look.
the background is:
in a real model, we found there are several TF const nodes with different names but have same values.
we want to merge these consts node into a single one, then further merge duplicate nodes optimizer can merge subsequent sub-graphs into a single one, which can improve performance for that model.

but currently, all tf const nodes are mapped to onnx initializers.
so at onnx optimizer stage, there is no way to tell whether an initializer is a real initializer (from tf placeholderwithdefault, which cannot merge) or from a regular tf const (which should merge).

this PR aims to distinguish the two.

please refer to https://github.com/onnx/tensorflow-onnx/pull/585#issuecomment-503955931 for more information.

nbcsm:
@lucienwang1009, let's make sure internal models can pass too.

lucienwang1009:
@nbcsm internal tests are green.

PR #624, opened 2019-06-25
make convert.py has same procedure with run_pretrained_models.py

zhijxu-MS:


tf2onnx/convert.py
```diff
@@ -145,10 +145,10 @@ def main():
                              output_names=outputs,
                              inputs_as_nchw=args.inputs_as_nchw)
 
-    model_proto = g.make_model("converted from {}".format(model_path))
-
-    logger.info("")
-    model_proto = GraphUtil.optimize_model_proto(model_proto)
+    onnx_graph = optimizer.optimize_graph(g)
+    if not onnx_graph:
```
nbcsm: optimizer.optimize_graph should always return a valid graph, no need to check None here.

you may want to add a None check at https://github.com/onnx/tensorflow-onnx/blob/d239d1cc5f73d45be93b2f2160d408253538871d/tf2onnx/optimizer/__init__.py#L50
something like ```graph = opt.optimize(current) or graph```

and we can skip None check here.

zhijxu-MS: done.

tf2onnx/convert.py
```diff
@@ -14,7 +14,7 @@
 
 import tensorflow as tf
 
-from tf2onnx.graph import GraphUtil
+from tf2onnx import optimizer
```
nbcsm: merge with ```from tf2onnx import constants, loader, logging, utils```?

zhijxu-MS: done

PR #625, opened 2019-06-25
fix rewriter bugs

lucienwang1009:
1. Retain nodes with other dependencies in every rewriter.
2. Check graph integrity after rewriting in verbose mode.

### DIFF
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -1183,16 +1183,12 @@ class Graph(object):
 
     def delete_nodes_without_dependency(self, to_delete):
         """Delete nodes in `to_delete` without third-party dependency."""
+        delete_set = set(to_delete)
         for n in to_delete:
-            can_delete = True
+            out_consumers = set()
             for out in n.output:
-                if not can_delete:
-                    break
-                for consumer in self.find_output_consumers(out):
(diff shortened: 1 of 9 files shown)


tf2onnx/graph.py
```diff
@@ -1169,6 +1181,20 @@ def delete_unused_nodes(self, outputs_name):
                     body_graph.delete_unused_nodes(body_graph.outputs)
         self.reset_nodes(related_nodes)
 
+    def delete_nodes_without_dependency(self, to_delete):
+        """Delete nodes in `to_delete` without third-party dependency."""
+        for n in to_delete:
+            can_delete = True
```
zhijxu-MS: optional:
find all consumers into a set, and use issubset, so the for ... for.... is not need.

tf2onnx/graph.py
```diff
@@ -591,7 +603,7 @@ def update_node_shape_dtype(self, node, override=False):
         initializers = []
         for i, inp in enumerate(node.inputs):
             if inp is None:
-                if logger.isEnabledFor(logging.INFO):
+                if logger.isEnabledFor(constants.VERBOSE):
```
nbcsm: logger.isEnabledFor(logging.VERBOSE) does not work?

PR #631, opened 2019-06-27
[Optimization] merge constant with same value

lucienwang1009:


tf2onnx/optimizer/merge_duplicated_nodes_optimizer.py
```diff
@@ -59,14 +61,25 @@ def _del_nodes_if_duplicated(self, nodes_group, graph):
...
                 else:
                     unprocessed_node.append(node)
 
             self._merge_nodes_that_are_duplicated(nodes_to_process, graph)
             nodes_group = unprocessed_node
 
+    def _have_equal_attr(self, node_1, node_2):
+        if node_1.attr == node_2.attr:
+            return True
+        if node_1.is_const() and node_2.is_const():
```
nbcsm: do not use is_const, just check ONNX op type.

lucienwang1009: GraphUtils will convert Constant node to Const type.
https://github.com/onnx/tensorflow-onnx/blob/c6d531c82c8d27d9312259881344f09dc293c044/tf2onnx/graph.py#L1289-L1290

PR #632, opened 2019-06-27
enhance check_integrity

zhijxu-MS:


tf2onnx/graph.py
```diff
@@ -575,6 +575,11 @@ def reset_nodes(self, ops):
         self._dtypes = remained_dtypes
         self._output_shapes = remained_shapes
 
+    def is_empty_input(self, name):
+        # in ONNX, operation may have optional input and an empty string may be used
+        # in the place of an actual argument's name to indicate a missing argument
+        return name == ""
```
lucienwang1009: utils.ONNX_EMPTY_INPUT?

zhijxu-MS: done

PR #633, opened 2019-06-27
Transpose optimizer

lei-Qiao:
1. make add handler works when trans.inputs[0].type = Conv
2. change maxmin handler like sub handler
3. add test cases for maxmin, sub, add handlers


tests/test_optimizers.py
```diff
@@ -394,6 +420,148 @@ def test_trans_with_sub(self):
...
+            "transpose-add-test-input-const",
+            [helper.make_tensor_value_info("X", TensorProto.FLOAT, (1, 1, 3, 3))],
+            [helper.make_tensor_value_info("res", TensorProto.FLOAT, (1, 1, 3, 3))],
+        )
+
+        model_proto = helper.make_model(graph, producer_name="onnx-tests")
+        self.run_transpose_compare(["res"], {"X": np.random.randn(1, 1, 3, 3).astype(np.float32)},
+                                   model_proto, remaining_transpose_num=0)
+
+    def test_transpose_add_with_conv_1(self):
```
zhijxu-MS: what is the difference between test_transpose_add_with_conv_1 and test_transpose_add_with_conv_2?

please put a comment, so we know the case where we fail.

tests/test_optimizers.py
```diff
@@ -394,6 +420,148 @@ def test_trans_with_sub(self):
...
+        )
+
+        model_proto = helper.make_model(graph, producer_name="onnx-tests")
+        self.run_transpose_compare(["res"], {"X": np.random.randn(1, 1, 3, 3).astype(np.float32),
+                                             "A": np.random.randn(1, 3, 3, 1).astype(np.float32)},
+                                   model_proto, remaining_transpose_num=0)
+
+    def test_transpose_add_with_input_const(self):
+        const_1_val = np.random.randn(1, 3, 3, 1).astype(np.float32).reshape(9).tolist()
+        const_1 = helper.make_tensor("const_1", TensorProto.FLOAT, (1, 3, 3, 1), const_1_val)
```
zhijxu-MS: why reshape(9) and then you want to create a const with shape (1, 3, 3, 1)?

tests/test_optimizers.py
lei-Qiao: In test_transpose_add_with_conv_1, the bias dim is correct and can be merged into Conv.

In test_transpose_add_with_conv_2, the bias can not be merged into Conv, and the add handler just remove the transposes around Add node

I put a comment in the test case

tests/test_optimizers.py
lei-Qiao: Because the const_val needs to be flatten for make_tensor, and wants to show the dim information.

Change the code like this:

const_b_val = np.random.randn(1, 3, 3, 1).astype(np.float32)
const_b = helper.make_tensor("const_b", TensorProto.FLOAT, (1, 3, 3, 1), const_b_val.flatten())

PR #634, opened 2019-07-01
fix dropout bug on TF 1.14 and enable more unittests.

lucienwang1009:


nbcsm:
@lucienwang1009, could you help to update readme.md for supported tf versions as well?

PR #661, opened 2019-07-23
optimize transpose only in current graph

lucienwang1009:
Keep transpose optimization in the current graph. Avoid them from running into body graphs.


lucienwang1009:
optimized transpose is from -2732 to 2733 on internal models.

lucienwang1009:
discussed offline with @zhijxu-MS, we should have to investigate more cases before making this change.

PR #667, opened 2019-07-30
fix transpose opt bug in opset 10

lucienwang1009:


snnn:
Does it have effect on any of the mlperf models?

guschmue:
I have tested them on opset-10 earlier this week and they are not impacted.

lucienwang1009:
Thanks @guschmue. We ought to include these mlperf models in our tests.

guschmue:
resnet50, mobilenet and ssd-mobilenet are in the tests already I think.

PR #673, opened 2019-08-20
using ConvTransposeWithDynamicPads to support tf conv2d_transpose

zhijxu-MS:


guschmue:
I assume the build issue is fixed in master.

PR #681, opened 2019-08-30
Added support for AddV2 op

djb15:


djb15:
Tried to sign the CLA multiple times over the weekend and it says I've successfully signed it, although it doesn't reflect here.  Let me know when/if I can try again

guschmue:
I think the cla tool goes through all commits and wants those users agree.
There is one commit from dirkbrink that is different than your github id.

djb15:
Ah that sorted it, thanks!

guschmue:
thanks !!!

PR #689, opened 2019-09-10
map tensorflow op to onnxruntime/pyop (purely for debugging)

guschmue:


guschmue:
kwargs not supported yet ... putting it on ice for now.

PR #696, opened 2019-09-23
Md dev

mxdilln:
It appears that batch iterators for training are not supported by onnx yet. This is a script that attempts to train a simple tf model using a batch iterator with dummy data, freeze the model and then convert the trained model to onnx. It currently fails at the conversion stage.

As most models will be designed to train in batches, this would be a great feature to include in tf2onnx!


guschmue:
closing because of CLA.

PR #701, opened 2019-10-04
cast topk index output back to int32

guschmue:


tf2onnx/onnx_opset/tensor.py
```diff
@@ -892,10 +892,15 @@ def version_1(cls, ctx, node, **kwargs):
     def version_10(cls, ctx, node, **kwargs):
         # onnx only supports input K as a 1D tesor with dtype int64
         # while in tf, K is a 0D tensor with dtype int32
+        dtypes = node.output_dtypes
         k_0d = node.input[1]
         cast = ctx.make_node("Cast", [k_0d], attr={"to": onnx_pb.TensorProto.INT64})
         k_1d = ctx.make_node("Unsqueeze", cast.output, attr={"axes": [0]})
         ctx.replace_input(node, k_0d, k_1d.output[0])
+        # cast the index output to int32
+        cast_out = ctx.insert_new_node_on_output("Cast", node.output[1], name=utils.make_name(node.name), to=dtypes[1])
```
jignparm: Should we only add a cast if required (sometimes node.output[1] will be equal to dtypes[1] already)?

One benefit of always adding a cast is that the generated pattern of TopK is always the same (and the optimizer can eliminated back-to-back casts).

guschmue: onnx op always has int64. tensorflow op always has int32.

jignparm: Ok -- looks like it's always required.

PR #713, opened 2019-10-19
Add optimizer to remove back-to-back nodes (e.g. Cast and Transpose)

jignparm:
Remove consecutive 'cast' or 'transpose' nodes, if they are not used by other operators.

### DIFF
--- a/tf2onnx/optimizer/back_to_back_optimizer.py
+++ b/tf2onnx/optimizer/back_to_back_optimizer.py
@@ -59,21 +59,33 @@ class BackToBackOptimizer(GraphOptimizerBase):
                         continue
                     # update downstream nodes, delete this one
                     for node2 in downstream_nodes:
-                        handler(node, node2)
+                        handler(g, node, node2)
                         # add node2 to q, in case it has downstream nodes
                         q.append(node2.output[0])
                     g.remove_node(node.name)
+
         return g
 
     @staticmethod
     @_register_func("Cast")
-    def _fold_cast(node1, node2):
+    def _fold_cast(g, node1, node2):
         node2.input[0] = node1.input[0]
 
     @staticmethod
     @_register_func("Transpose")
-    def _fold_cast2(node1, node2):
+    def _fold_cast2(g, node1, node2):
         node2.input[0] = node1.input[0]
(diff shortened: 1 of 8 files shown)


tf2onnx/optimizer/back_to_back_optimizer.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def _optimize(self, graph):
+        return self._apply_optimization(graph, self._optimize_at_current_graph_level)
+
+    def _optimize_at_current_graph_level(self, g):
+        for optype, handler in _func_map.items():
+            nodes = [n for n in g.get_nodes() if n.type == optype]
+            # find consumer list for each cast node
+            has_dependencies = set()
+            consumer_node_ids = {n.output[0]: [] for n in nodes}
```
guschmue: There is a method on the Graph class called find_consumers() that we use instead of doing that manual.

tf2onnx/optimizer/back_to_back_optimizer.py
```diff
@@ -0,0 +1,79 @@
...
+            for n in nodes:
+                if n.input[0] in consumer_node_ids:
+                    consumer_node_ids[n.input[0]].extend([n])
+                    has_dependencies.add(n.output[0])
+
+            # nodes with no dependencies
+            q = list(set(consumer_node_ids.keys()) - has_dependencies)
```
guschmue: what if a node is a model output ?

tf2onnx/optimizer/back_to_back_optimizer.py
jignparm: I noticed the method you  mentioned. The `find_output_consumers()` method find all nodes that consume a particular output, however the consumers can be different op_type (i.e. doesn't not filter by op-type,). Also, it makes a pass over all the nodes. It seemed a bit more efficient this way. Let me know thoughts.

tf2onnx/optimizer/back_to_back_optimizer.py
jignparm: Good point. I'll update.

jignparm: Updated. please review.

PR #721, opened 2019-10-28
support for opset 11 and tf-1.15

guschmue:


tf2onnx/graph.py
```diff
@@ -362,6 +362,27 @@ def _graph_check(self):
...
+    def maybe_cast_input(self, supported, type_map):
+        """.maybe_cast_input
+        Args:
+            supported: list of supported types for inputs
+            type_map: dict type to supported type mapping
+        """
+        did_cast = False
+        for i, name in enumerate(self.input):
+            dtype = self.graph.get_dtype(name)
+            if dtype not in supported[i]:
```
jignparm: Did you mean this instead?

```python
if dtype not in supported:
```

guschmue: no, supported is a list of list

ci_build/azure_pipelines/unit_test.yml
```diff
@@ -15,7 +15,7 @@ stages:
     
     - template: 'templates/job_generator.yml'
       parameters:
-        tf_versions: ['1.13.1', '1.12', '1.11', '1.10', '1.9', '1.8', '1.7', '1.5']
+        tf_versions: ['1.14', '1.12', '1.11', '1.10', '1.9', '1.8', '1.7']
```
jignparm: Title says tf-1.15 -- but only 1.14 is listed -- typo?

guschmue: the branch is 15 but there are more issues. I fell back to 14 so we can at least test opset-11

guschmue: btw, its passing so at least opset-11 is now good. This is temporary, just 1 continues build on ort nightly + opset-11. Once onnxruntime-1.0 is out we'll have a permanent opset-11 build.

PR #722, opened 2019-10-29
Fix Clip Op for LSTM for opset11

jignparm:
Fix Clip operator for LSTM opset11. Attributes have changed to inputs, do the LSTM/RNN logic needs to be updated.


tf2onnx/onnx_opset/rnn.py
```diff
@@ -139,8 +139,8 @@ def make_sigmoid(i, w, b):
             else:
                 dtype = utils.map_onnx_to_numpy_type(ctx.get_dtype(cs))
                 min_const = ctx.make_const(utils.make_name(f'{node.name}_min'), np.array(-cell_clip, dtype=dtype))
```
guschmue: Ah, our README file says we support python 3.5 - lets not use f'

Issue #716, opened 2019-10-22
Nodes in Loop subgraph are not topologically sorted


jiafatom:
**Describe the bug**
When we convert a keras model, we are using tf2onnx for custom layers, the model can be converted. But onnxruntime crashes because nodes in the subgraph of Loop node are not topologically sorted. Can we fix this? Thanks.

**Urgency**
This is a critical bug.

jignparm:
@jiafatom, what's the onnxruntime error message? Also, is the model small enough to attach to this issue?

PR #724, opened 2019-10-31
top sort subgraphs

guschmue:
Fix for https://github.com/onnx/tensorflow-onnx/issues/716

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -345,6 +345,8 @@ def tensorflow_onnx_mapping(g, ops_mapping):
                 m_ops, unm_ops, body_exceptions = tensorflow_onnx_mapping(b_g, ops_mapping)
                 mapped_op += m_ops
                 unmapped_op += unm_ops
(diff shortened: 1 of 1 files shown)


tf2onnx/tfonnx.py
```diff
@@ -345,6 +345,8 @@ def tensorflow_onnx_mapping(g, ops_mapping):
                 m_ops, unm_ops, body_exceptions = tensorflow_onnx_mapping(b_g, ops_mapping)
                 mapped_op += m_ops
                 unmapped_op += unm_ops
+                # topological_sort on the body in case processing has changed the order
+                b_g.topological_sort(b_g.get_nodes())
                 exceptions.extend(body_exceptions)
```
jignparm: Sorting is done twice (once at line 341) -- are both required?

guschmue: The 2nd sorting is **after** processing which might change the order.

guschmue:
thanks man!

PR #744, opened 2019-11-20
Remove unnecessary ops when converting BERT model

jignparm:
Added optimizations to remove unnecessary Cast, Squeeze and Unsqueeze operators, which make the model more readable visually.

* removed unnecessary  "Squeeze + Unsqueeze" operators between Slice and Concat
* remove  too many extra "Cast" operators when converting StridedSlice op to an ONNX Slice op.


guschmue:
just need to fix the unit test :)

Issue #740, opened 2019-11-15
KeyError: tf.float32_ref


I-CANT-CODE:
Hi I am getting this issue when running the command:

python3 -m tf2onnx.convert  --input my_model.pb --inputs X:0 --outputs summaries/up_convolution_out/activations:0 --output model.onnx --verbose

ISSUE:

WARNING:tensorflow:From /usr/local/lib/python3.5/dist-packages/tf2onnx/verbose_logging.py:72: The name tf.logging.set_verbosity is deprecated. Please use tf.compat.v1.logging.set_verbosity instead.

2019-11-15 12:17:48,167 - WARNING - tensorflow: From /usr/local/lib/python3.5/dist-packages/tf2onnx/verbose_logging.py:72: The name tf.logging.set_verbosity is deprecated. Please use tf.compat.v1.logging.set_verbosity instead.

WARNING:tensorflow:From /usr/local/lib/python3.5/dist-packages/tf2onnx/loader.py:58: The name tf.reset_default_graph is deprecated. Please use tf.compat.v1.reset_default_graph instead.

2019-11-15 12:17:48,167 - WARNING - tensorflow: From /usr/local/lib/python3.5/dist-packages/tf2onnx/loader.py:58: The name tf.reset_default_graph is deprecated. Please use tf.compat.v1.reset_default_graph instead.

WARNING:tensorflow:From /usr/local/lib/python3.5/dist-packages/tf2onnx/loader.py:59: The name tf.Session is deprecated. Please use tf.compat.v1.Session instead.

2019-11-15 12:17:48,167 - WARNING - tensorflow: From /usr/local/lib/python3.5/dist-packages/tf2onnx/loader.py:59: The name tf.Session is deprecated. Please use tf.compat.v1.Session instead.

2019-11-15 12:17:48.168165: W tensorflow/stream_executor/platform/default/dso_loader.cc:55] Could not load dynamic library 'libcuda.so.1'; dlerror: libcuda.so.1: cannot open shared object file: No such file or directory
2019-11-15 12:17:48.168185: E tensorflow/stream_executor/cuda/cuda_driver.cc:318] failed call to cuInit: UNKNOWN ERROR (303)
2019-11-15 12:17:48.168212: I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:156] kernel driver does not appear to be running on this host (henry): /proc/driver/nvidia/version does not exist
[...]

PR #749, opened 2019-11-24
ConvBackpropInput with dynamic hw

RandySheriffH:
When input size is non-const, slice output to match expected shape.


guschmue:
This will remove the need for the workaround if HW is unknown in
https://github.com/onnx/tensorflow-onnx/issues/740

Issue #765, opened 2020-01-07
ConvTranspose missing auto_pad flag.


axeldavy:
**Describe the bug**
Models using ConvTranspose operations in ONNX a stride > 1 and output_shape have a different output running with tensorflow or with tensorrt using onnx.
Running the onnx with tensorflow gives the error:

> tensorflow.python.framework.errors_impl.InvalidArgumentError: 2 root error(s) found.
>   (0) Invalid argument: Conv2DSlowBackpropInput: Size of out_backprop doesn't match computed: actual = 40, computed = 39 spatial_dim: 2 input: 80 filter: 4 output: 40 stride: 2 dilation: 1
>          [[node conv2d_transpose (defined at /home/axel/.local/lib/python3.6/site-packages/tensorflow_core/python/framework/ops.py:1748) ]]
>   (1) Invalid argument: Conv2DSlowBackpropInput: Size of out_backprop doesn't match computed: actual = 40, computed = 39 spatial_dim: 2 input: 80 filter: 4 output: 40 stride: 2 dilation: 1
>          [[node conv2d_transpose (defined at /home/axel/.local/lib/python3.6/site-packages/tensorflow_core/python/framework/ops.py:1748) ]]
>          [[model_2/conv2d_2/Sigmoid/_11]]

**Urgency**
None

**System information**
- OS Platform and Distribution: Ubuntu 18.04
- Tensorflow Version: 1.15.0
- Python version: 3.6

My analysis of the issue is that in this particular scenario there is a missing auto_pad flag for the ConvTranpose operation. Setting it properly fixed the onnx for both tensorrt and tensorflow.

I will soon submit a PR with a proposed fix, but I decided to open this bug report for tracking.

jignparm:
@axeldavy , can you post something similar to [this short example](https://github.com/onnx/tensorflow-onnx/blob/master/tests/test_backend.py#L2914) with the appropriate values, to illustrate when the failure occurs?

axeldavy:
I don't understand how these tests work.

I see the only two tests of tf.nn.conv2d_transpose are with 'VALID'.
Maybe the bug would trigger with identical tests but with 'SAME'.

jignparm:
[...]

PR #766, opened 2020-01-07
Add missing auto_pad attribute for ConvTranspose

axeldavy:
auto_pad affects how the layer is executed.
For tensorflow compatibility, it should be set to
SAME_UPPER.

Fixes: https://github.com/onnx/tensorflow-onnx/issues/765

### DIFF
--- a/tf2onnx/onnx_opset/nn.py
+++ b/tf2onnx/onnx_opset/nn.py
@@ -240,6 +240,8 @@ class ConvTranspose:
             utils.make_sure(new_output_shape.count(-1) <= 0, "output h and w need to be known")
             utils.make_sure(new_output_shape[0] >= input_hw[0] and new_output_shape[1] >= input_hw[1],
                             "output h and w cannot be smaller than input h and w.")
+            # This flag affects the output when stride > 1 and is required for compatibility
+            node.set_attr("auto_pad", "SAME_UPPER")
             node.set_attr("output_shape", new_output_shape)
         else:
             input_shape = ctx.make_node("Cast", [node.input[0]], attr={'to': TensorProto.INT64})


jignparm:
The unit test below is failing with the following message. Maybe the attribute value should only be set under certain conditions?

```
self = <test_backend.BackendTests testMethod=test_Conv2DBackpropInput_const_strided>

    @check_opset_min_version(10, "Conv2DBackpropInput")
    def test_Conv2DBackpropInput_const_strided(self):
        input_sizes_val = np.array([1, 10, 10, 3], dtype=np.int32)
        filter_val = np.random.randint(low=0, high=256, size=[3, 3, 3, 5]).astype(np.float32)
        out_backprop_val = np.random.randint(low=0, high=256, size=[1, 5, 5, 5]).astype(np.float32)
        _ = tf.nn.conv2d_backprop_input(input_sizes=input_sizes_val, filter=filter_val, out_backprop=out_backprop_val,
                                        strides=[1, 2, 2, 1], padding='SAME', name=_TFOUTPUT)
>       self._run_test_case([_OUTPUT], {})

tests\test_backend.py:2920: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests\test_backend.py:111: in _run_test_case
    return self.run_test_case(feed_dict, [], output_names_with_port, **kwargs)
tests\backend_test_base.py:131: in run_test_case
    self.assertAllClose(expected_val, actual_val, rtol=rtol, atol=atol)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

expected = array([[[[111087.,  95576., 113458.],
         [ 46745.,  99963., 104671.],
         [202821., 163341., 136903.],
    ...5466.,  55948.,  40606.],
         [ 86682., 150351., 198641.],
         [ 86307.,  59883.,  47439.]]]], dtype=float32)
actual = array([[[[ 93894.,  39163.,  34824.],
         [105692., 119105., 166645.],
         [ 58051.,  21856.,  26100.],
    ...9971., 152360., 152864.],
         [ 39489.,  94722.,  66264.],
         [ 89379.,  96516.,  85563.]]]], dtype=float32)
kwargs = {'atol': 1e-05, 'rtol': 1e-07}

```

guschmue:
I'm not sure if auto_pad will work since we set the output_shape already. Maybe best would be to add a unit test for strides>1 and take a closer look if its numerical not the same.

axeldavy:
It seems Keras uses SAME_LOWER for transpose convolutions:
[https://github.com/onnx/keras-onnx/blob/master/keras2onnx/ke2onnx/conv.py#L147](https://github.com/onnx/keras-onnx/blob/master/keras2onnx/ke2onnx/conv.py#L147)

According to this bug report:
[https://github.com/onnx/onnx/issues/1609](https://github.com/onnx/onnx/issues/1609)
There is a mismatch in the spec between SAME_LOWER and SAME_UPPER for this definition.
I don't know if that really can be fixed in the spec after so many months.

Thus possibly the correct auto_pad is SAME_LOWER. I'll check if #765 is also happy with this flag.

axeldavy:
Testing with #765 the onnx fails to execute with SAME_LOWER under tensorflow (it indicates ConvTranspose with auto_pad SAME_LOWER" is not supported at the moment).
With tensorrt, the output is visually similar to with SAME_UPPER. I could do a numerical comparison to check which is the closest to the mathematical output of the original model if needed.

Additionnaly, even if this won't help the unit test, the flag should probably only be set for "padding == 'SAME'" (not setting the flag seems similar to 'VALID' according to onnx-tensorflow's onnx_tf/handlers/backend/conv_mixin.py).

My intuition is either the spec is considered wrong, and in that case SAME_UPPER would be the correct flag, or the spec should be followed, and in that case SAME_LOWER would be the correct flag, and in that case onnx-tensorflow is probably incorrect (mixing the two flags for tranpose convolutions).

axeldavy:
Given the discussion above, I think my patch is quite incorrect (should only be activated for SAME padding, fails a unit test, probably is unrelated to stride > 1 and should affect stride = 1 too), and there are possible required changes outside of tensorflow-onnx (onnx spec, and/or onnx-tensorflow).

It seems to need a harder fix than I would have thought, and I think it is more appropriate that I leave this issue to a more experienced tensorflow/onnx developper.

PR #776, opened 2020-01-21
Use the 'sizes' parameters of Resize op instead of 'scales'

vdel:
I  recently tried to convert a TF model via tensorflow-onnx and run it using onnx-tensorrt (https://github.com/onnx/onnx-tensorrt). 

I had some upscaling (ResizeBilinear) operations with a dynamic target size parameter (i.e. dynamic computation of the resize shape). The constructed graph for this ResizeBilinear TF op went through the "else" branch [here](https://github.com/onnx/tensorflow-onnx/blob/84cfa03d35846ddaf524f42890ffc1576ee95134/tf2onnx/onnx_opset/nn.py#L686) because of this dynamic shape. The `scales` parameter was thus the output of the `Concat` op [here](https://github.com/onnx/tensorflow-onnx/blob/84cfa03d35846ddaf524f42890ffc1576ee95134/tf2onnx/onnx_opset/nn.py#L699).

It then failed in Tensort runtime [at this line](https://github.com/onnx/onnx-tensorrt/blob/457f2c34918e12a4a951c7d320ef7c6cf096f6ee/builtin_op_importers.cpp#L2466) because `scales.is_weights()` is false as `scales` if the output of the `Concat` op as discussed above and not a parameter of the model.

Fortunately, ONNX 11 opset introduced the `sizes` parameter for the Resize op, which is well handled by TensorRT (see [those lines](https://github.com/onnx/onnx-tensorrt/blob/457f2c34918e12a4a951c7d320ef7c6cf096f6ee/builtin_op_importers.cpp#L2454-L2460)) and which is actually already used by tf2onnx to handle the CropAndResize TF op. 

This PR:
- uses this `sizes` parameter for the ResizeBilinear/ResizeNearestNeighbor TF ops
- factorize the introduced code with the CropAndResize op

Questions:
- I renamed `roi_required` into `use_target_size` because it seemed to not be used elsewhere and because I felt the naming was more relevant. But it could be great to confirm this from people who know the codebase.

### DIFF
--- a/tf2onnx/onnx_opset/nn.py
+++ b/tf2onnx/onnx_opset/nn.py
@@ -695,7 +695,13 @@ class Resize:
             final_target_size = build_dynamic_target_size(ctx, input_nchw, node.inputs[1])
             roi = ctx.make_const(utils.make_name("roi"), np.array([]).astype(np.float32))
             const_empty_float = ctx.make_const(utils.make_name("const_empty_float"), np.array([], dtype=np.float32))
-            upsample = ctx.make_node("Resize", [input_nchw.output[0], roi.output[0], const_empty_float.output[0], final_target_size.output[0]],
+            resize_inputs = [
+                input_nchw.output[0],
+                roi.output[0],
+                const_empty_float.output[0],
+                final_target_size.output[0]
+            ]
+            upsample = ctx.make_node("Resize", resize_inputs,
                                      attr={"mode": mode, "nearest_mode": "floor",
                                            "coordinate_transformation_mode": "asymmetric"})
         else:
@@ -707,7 +713,8 @@ class Resize:
                 n, h, w, c = shape
                 nh, nw = target_shape
                 # scales is nchw
-                # the reason not storing data at raw field is because of the bug: https://github.com/onnx/onnx/issues/1852
+                # the reason not storing data at raw field is because of the bug:
+                # https://github.com/onnx/onnx/issues/1852
                 scale_val = np.array([1.0, 1.0, float(nh) / h, float(nw) / w]).astype(np.float32)
                 scales = ctx.make_const(utils.make_name("scales"), scale_val, raw=False)
             else:
--- a/tf2onnx/onnx_opset/nn.py
+++ b/tf2onnx/onnx_opset/nn.py
@@ -192,6 +192,28 @@ def conv_kernel_shape(ctx, node, input_idx, spatial=2):
     return kernel_shape
 
 
+def build_dynamic_target_size(ctx, transposed_intput, target_hw):
+    """
+    Build the target tensor shape for the Resize op.
+
+    Args:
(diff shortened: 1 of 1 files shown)


jignparm:
Thanks for the contribution!

Renaming `roi_required' to  'use_target_size' looks fine from a nomenclature standpoint. 

Was this breaking in a way that was not caught previously by the unit tests here?:
https://github.com/onnx/tensorflow-onnx/blob/master/tests/test_backend.py#L2507
https://github.com/onnx/tensorflow-onnx/blob/master/tests/test_backend.py#L1894

Also, would be great to have the following:
* Update the unit test so it catches the previous error (and passes in the updated implementation)
* Fix the PyLint warnings

vdel:
> Was this breaking in a way that was not caught previously by the unit tests here?:

Actually no. What was done by tf2onnx before this PR was correct from ONXX point of view. See [documentation](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Resize): you can either use `scales` or `sizes`. 

It's just that TensorRT does not support non-const `scales` input (I opened an issue here: https://github.com/onnx/onnx-tensorrt/issues/389), and it seems much more complicated to fix in TensorRT runtime. TensorRT runtime hopefully supports the `sizes` parameter (even when its value is dynamic) and I took the initiative to harmonize its use across `CropAndResize` and `resize_bilinear`. 

This PR does not change the behavior of `CropAndResize`, it just factorize the code that creates the `sizes` node and use this code in `resize_bilinear` as well.

So I am not sure if we really need new use-cases or how we can check that the `sizes` parameters is used for opset 11 instead of `scales`. 

Also, the pylint build is not accessible anymore, is there a command to trigger a new build to fix linting issues ?

jignparm:
>Also, the pylint build is not accessible anymore, is there a command to trigger a new build to fix linting issues ?

A couple of options regarding the PyLint error. Submit a small change to trigger a new build, or else run PyLint locally using the command below.

```
pylint --rcfile=tools/pylintrc --ignore=version.py --disable=cyclic-import tf2onnx tests/backend_test_base.py tests/common.py tests/conftest.py tests/run_pretrained_models.py tests/test_backend.py tests/test_cond.py tests/test_const_fold.py tests/test_convert.py tests/test_cudnn_compatible_gru.py tests/test_custom_rnncell.py tests/test_graph.py tests/test_grublock.py tests/test_gru.py tests/test_internals.py tests/test_loops.py tests/test_lstmblock.py tests/test_lstm.py tests/test_onnx_shape_inference.py tests/test_optimizers.py tests/test_seq2seq.py tests/test_tf_shape_inference.py tools -j 0
```

vdel:
@jignparm takes for the command, I've corrected the pylint errors.

jignparm:
@vdel, thanks for the PR. Merged.

PR #781, opened 2020-01-30
support for tf.math.unosrted_segmentX for the special case of num_segments=1

guschmue:


tests/test_backend.py
```diff
@@ -1171,6 +1171,27 @@ def test_reducesum(self):
...
+            self._run_test_case([_OUTPUT], {_INPUT: x_val})
+
+            # TODO: enable once we have support for this
+            # num_segments > 1
+            # tf.reset_default_graph()
+            # x_val = np.array([[1, 2, 3, 4], [5, 6, 7, 8], [4, 3, 2, 1]]).astype(np.float32)
+            # x0 = tf.placeholder(tf.float32, x_val.shape, name=_TFINPUT)
+            # x_ = tf.math.unsorted_segment_sum(x0, tf.constant([0, 1, 0]), num_segments=2)
+            # _ = tf.identity(x_, name=_TFOUTPUT)
+            # self._run_test_case([_OUTPUT], {_INPUT: x_val})
```
jignparm: Should probably uncomment num_segments > 1   before merging with master.

guschmue: but that would not pass unit tests right now.

PR #786, opened 2020-02-04
Convert GNMT

RandySheriffH:
- Replace iterator with placeholders;
- Add lookup support.

### DIFF
--- a/tf2onnx/custom_opsets/__init__.py
+++ b/tf2onnx/custom_opsets/__init__.py
@@ -3,3 +3,4 @@
 """ custom tf2onnx mapping functions. """
 
 from . import ms
+from . import onnx_ml
--- a/tf2onnx/custom_opsets/onnx_ml.py
+++ b/tf2onnx/custom_opsets/onnx_ml.py
@@ -4,32 +4,35 @@
 import numpy as np
 
 from onnx import onnx_pb
-from onnx.onnx_pb import TensorProto
-from tf2onnx import constants, utils
+from tf2onnx import constants
 from tf2onnx.handler import tf_op
 
 @tf_op("HashTableV2")
 class HashTable:
     @classmethod
     def version_11(cls, ctx, node, **kwargs):
+        """ HashTable will be removed """
         pass
 
 @tf_op("LookupTableFindV2")
 class LookupTableFind:
     @classmethod
     def version_11(cls, ctx, node, **kwargs):
+        """ convert lookup to category mapper """
         table_node = node.inputs[0]
         file_path = table_node.get_attr_value("shared_name")[11:-6]
         cats_int64s = []
         cats_strings = []
         with open(file_path, 'r') as f:
-            for i,s in enumerate(f.readlines()):
+            for i, s in enumerate(f.readlines()):
                 cats_int64s.append(i)
                 cats_strings.append(s.strip())
         node_inputs = node.input
         node_outputs = node.output
         ctx.remove_node(node.name)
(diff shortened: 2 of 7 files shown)


tf2onnx/custom_opsets/__init__.py
```diff
@@ -3,3 +3,4 @@
 """ custom tf2onnx mapping functions. """
 
 from . import ms
+from . import onnx_ml
```
jignparm: Does this require additional python package?

jignparm: Looks like it's not a different package.

tf2onnx/custom_opsets/onnx_ml.py
```diff
@@ -0,0 +1,39 @@
...
+@tf_op("HashTableV2")
+class HashTable:
+    @classmethod
+    def version_11(cls, ctx, node, **kwargs):
+        """ HashTable will be removed """
+        pass
+
+
+@tf_op("LookupTableFindV2")
+class LookupTableFind:
```
jignparm: Add unit test?

RandySheriffH: Added.

tf2onnx/tfonnx.py
```diff
@@ -99,6 +99,12 @@ def tflist_to_onnx(node_list, shape_override):
                 continue
             elif a in ignored_attr:
                 continue
+            elif a in ["key_dtype", "value_dtype", "Tin", "Tout"]:
+                attr[a] = utils.map_tf_dtype(utils.get_tf_node_attr(node, a))
+            elif a == "output_types":
```
guschmue: isn't this already stored in graph._output_types ?

tf2onnx/tfonnx.py
```diff
@@ -99,6 +99,12 @@ def tflist_to_onnx(node_list, shape_override):
...
+                attr[a] = [utils.map_tf_dtype(v) for v in utils.get_tf_node_attr(node, a)]
+            elif a == "output_shapes":
+                attr[a] = utils.get_tf_output_shapes_attr(node)
```
guschmue: isn't this already stored in graph._output_shapes ?

tf2onnx/tfonnx.py
```diff
@@ -99,6 +99,12 @@ def tflist_to_onnx(node_list, shape_override):
                 continue
             elif a in ignored_attr:
                 continue
+            elif a in ["key_dtype", "value_dtype", "Tin", "Tout"]:
```
guschmue: is this needed ? Should this not be the same as the dtype if that specific input ?

tf2onnx/custom_opsets/onnx_ml.py
```diff
@@ -0,0 +1,39 @@
...
+from tf2onnx import constants
+from tf2onnx.handler import tf_op
+
+
+# pylint: disable=unused-argument,missing-docstring,unnecessary-pass
+
+@tf_op("HashTableV2")
+class HashTable:
+    @classmethod
+    def version_11(cls, ctx, node, **kwargs):
```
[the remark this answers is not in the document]
RandySheriffH: Switched to version 8.

tf2onnx/tfonnx.py
RandySheriffH: Without the line - we get exception from onnx complaining about unknown iterable r type.

guschmue: So onnx is expecting an attribute output_shapes or output_types ? In that case you should set this in the handler for this op and not pass it 1:1 from tensorflow because all other ops don't need this.

RandySheriffH: removed.

tf2onnx/tfonnx.py
RandySheriffH: removed.

PR #792, opened 2020-02-07
trim ci pipeline to tf-1.12 or better and replace python 3.5 with 3.8

guschmue:


ci_build/azure_pipelines/pretrained_model_test-matrix.yml
```diff
@@ -4,17 +4,17 @@ jobs:
 - template: 'templates/job_generator.yml'
   parameters:
     platforms: ['linux', 'windows']
-    python_versions: ['3.6', '3.5']
-    tf_versions: ['1.13.1', '1.12', '1.11', '1.10', '1.9', '1.8', '1.7', '1.6', '1.5']
```
jignparm: Looks like this is removing a lot of versions -- are they all failing due to the same error, or different errors (when upgrading to TF 2.0 branch)? If they are due to a small number of (sames) errors, might be worth salvaging the versions above. Thoughts?

guschmue: they are currently not failing but are costing a lot of time in the ci pipeline so we want to stop supporting releases older than tf-1.12 (from Sep 2018).
Since we still support importing graphs from older releases it should be ok to ask people to install a decent version of tensorflow.
```tf-1.12``` introduced the tf.compat namespace and if we can depend on that being there we can get rid of a bunch of warning we see today.

jignparm: If they are not failing, can we keep '1.5' (and remove the others to speed up the pipeline). That way one 'old' version is still being tested (even if not officially supported). This will keep some back-wards compatibility checks in the CI pipeline.

guschmue: They are not failing right now but I want to take a dependency on the tf.compat namespace to get rid of 'depreciated' warnings.
tf-1.5 came out in Jan-2018 - doubt that tensorflow team will support such an old version and I don't see any reason why we should. The api in tf-1.x was mostly stable - there is no real impact for the user when upgrading to a newer version (in my view). Its different for the graph itself (like if you have an old tf-1.1 model as example) but that we can still read with newer tensorflow versions.

jignparm: Ok, the  tf.compat namespace dependency makes sense.

Issue #720, opened 2019-10-26
add tf-1.15 support


guschmue:

guschmue:
Fixed with https://github.com/onnx/tensorflow-onnx/pull/803

Issue #691, opened 2019-09-12
tf-2.0 support


guschmue:

srkm009:
@guschmue: By when do you see the support for TF2.0 complete?

guschmue:
basic unit tests are working. Still some issues with tf.while and tensor lists. Maybe another week.

Dragme:
What is the status for this?

edwinZeng:
I have trained a network using keras from Tensorflow 2.1, but i cannot save the model as onnx file format with keras2onnx 1.6.0, throwing the error as "tf2onnx loading failed with tensorflow package", probably due to compatible issue with the library and the tensorflow 2.0. May I know any patch to this or can daily build is able to resolve this? Thank you

tgolsson:
@guschmue Maybe you can make a list of what needs to be done/related issues to bring this forward? I could probably contribute a bunch of hours towards this, and I'm sure others could as well if we knew what needs to be done.

guschmue:
I have most of it work - not fully done with tf.while but I'm pretty close to be done with it.

birdmw:
Is support for tf2 in ?

lesters:
@guschmue, echoing @tgolsson here. Is there anything we can do to assist? I think there are quite a few people waiting for this, maybe event with available time on their hands.

guschmue:
let me cleanup and merge what I have to master. A few minor left when using tensor array reads in tf.while but other than that things look good.

birdmw:
If you want to use TF2 you just need to make a python driver file, and then host it on flask. Put all that into a Dockerfile. Put the docker container into K8. In fact, that's exactly what Azure is doing for you. Literally, they are using flask... So you aren't really saving anything by going to ONNX

guschmue:
The branch gs/tf20 should be in decent shape for tf2.x. 
I'm in the process of getting this to master, just need to fight a little with CI and get it reviewed.

guschmue:
this was merged to master some time ago. Still some issue with lstm but other than that it should be fine.

chriselion:
The tf2onnx python library still can't be used with tensorflow 2.x installed:

```shell
$ pip freeze | grep -i 'tf\|tensorflow'
tensorflow==2.0.0
tensorflow-estimator==2.0.1
tf2onnx==1.5.6

$ python -c "import tf2onnx"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
[...]

PR #803, opened 2020-02-13
tf-2.x support

guschmue:
This adds initial support for tf-2.x and makes tf-1.15 work as well: https://github.com/onnx/tensorflow-onnx/issues/691 https://github.com/onnx/tensorflow-onnx/issues/720

This change supports tf-1.x (including tf-1.15) and tf-2.x.

For now I have not turned on tf-2.x in the CI so we can get master working perfectly good with tf-1.x. Once reviewed I'll turn on tf-2.x in the CI.

About 80% of the unit tests are passing on tf-2.x but I think I can get the remaining working pass in the next few days.

The change is mostly:
- touch every unit test to be able to run them in eager mode. We do this by isolating the model
   itself to a python function and can either run this directly in eager or compose a graph if we
  are using tf-1.x
- changes to deal with incompatibilities between tf-1.x and tf-2.x
- changes needed for tf-2.x
- restructure of all code that interacts with tf directly into tf_loader.py and tf_utils.py
- support for V2 controlflow 
- support for grappler


tests/backend_test_base.py
```diff
@@ -83,47 +93,74 @@ def run_test_case(self, feed_dict, input_names_with_port, output_names_with_port
...
-        if convert_var_to_const:
-            with tf.Session() as sess:
-                tf.tables_initializer().run()
-                variables_lib.global_variables_initializer().run()
-                output_name_without_port = [n.split(':')[0] for n in output_names_with_port]
-                graph_def = tf.graph_util.convert_variables_to_constants(sess, sess.graph_def,
-                                                                         output_name_without_port)
 
-            tf.reset_default_graph()
+        np.random.seed(1)  # Make it reproducible.
```
jignparm: Is this only for certain test cases, or needed for tf 2.0?

tests/backend_test_base.py
```diff
@@ -83,47 +93,74 @@ def run_test_case(self, feed_dict, input_names_with_port, output_names_with_port
...
-            utils.save_protobuf(model_path, sess.graph_def)
-            self.logger.debug("created file %s", model_path)
-
-        graph_def = tf_optimize(input_names_with_port, output_names_with_port,
-                                sess.graph_def, constant_fold)
-
-        if self.config.is_debug_mode:
-            model_path = os.path.join(self.test_data_directory, self._testMethodName + "_after_tf_optimize.pb")
-            utils.save_protobuf(model_path, graph_def)
-            self.logger.debug("created file  %s", model_path)
```
jignparm: This might be useful to keep around.

guschmue: yeap, it is still there but moved a few lines down.

tests/run_pretrained_models.py
```diff
@@ -206,22 +215,21 @@ def run_test(self, name, backend="caffe2", onnx_file=None, opset=None, extra_ops
...
+            graph_def, input_names, outputs = tf_loader.from_saved_model(model_path, input_names, outputs)
         else:
-            graph_def, input_names, outputs = loader.from_graphdef(model_path, input_names, outputs)
+            graph_def, input_names, outputs = tf_loader.from_graphdef(model_path, input_names, outputs)
 
-        # remove unused input names
-        input_names = list(set(input_names).intersection(self.input_names.keys()))
-        graph_def = tf2onnx.tfonnx.tf_optimize(input_names, self.output_names, graph_def, fold_const)
         if utils.is_debug_mode():
             utils.save_protobuf(os.path.join(TEMP_DIR, name + "_after_tf_optimize.pb"), graph_def)
```
jignparm: looks like the  tf_optimize() function is removed above, but the file name is still _after_tf_optimize.pb. The tf_optimize() was removed accidentally?

tests/backend_test_base.py
guschmue: We already initialize the  tf.set_random_seed before calling func() (both for tf-1.x and tf-2.x) and I thought its good to do the same to numpy in case one wants to use np.random.* in func.

tests/run_pretrained_models.py
guschmue: tf_optimize is a few lines above inside the tf_session()

Issue #770, opened 2020-01-14
Support MatrixDiagPart op


chinhuang007:
The Tensorflow MatrixDiagPart  op is currently not supported. We tried to convert a tf model in the env:

Ubuntu 18.04
Tensorflow: 1.13.1
Python: 3.6.8
ONNX: 1.5.0

Command:
python -m tf2onnx.convert --saved-model ./saved_model --output test_model.onnx --opset 10 --fold_const

The onnx file is created with an error message.
ERROR - Tensorflow op [MatrixDiagPart: MatrixDiagPart] is not supported
ERROR - Unsupported ops: Counter({'MatrixDiagPart': 1})

RandySheriffH:
@chinhuang007: you could try the fix with "--opset 11".

PR #807, opened 2020-02-19
Rashuai/matrix diag part

RandySheriffH:
Support tf.MatrixDiagPart to fix https://github.com/onnx/tensorflow-onnx/issues/770.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -3017,6 +3017,20 @@ class BackendTests(Tf2OnnxBackendTestBase):
         input_holder = tf.placeholder(tf.int64, input_val.shape, name=_TFINPUT)
         _ = tf.matrix_diag_part(input_holder, name=_TFOUTPUT)
         self._run_test_case([_OUTPUT], {_INPUT: input_val})
-
+        tf.reset_default_graph()
+        input_val = np.array([[[1, 2, 3], [4, 5, 6], [7, 8, 9], [10, 11, 12]]]).astype(
+            np.int64)
+        input_holder = tf.placeholder(tf.int64, input_val.shape, name=_TFINPUT)
+        _ = tf.matrix_diag_part(input_holder, name=_TFOUTPUT)
+        self._run_test_case([_OUTPUT], {_INPUT: input_val})
+        tf.reset_default_graph()
+        input_val = np.array([[[[1, 2, 3], [4, 5, 6], [7, 8, 9], [10, 11, 12]]],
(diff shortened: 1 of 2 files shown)


tests/test_backend.py
```diff
@@ -3010,6 +3010,12 @@ def test_hashtable_lookup(self):
         self._run_test_case([lookup_results.name], {_INPUT: query})
         os.remove(filnm)
 
+    @check_opset_min_version(11)
+    def test_MatrixDiagPart(self):
+        input_val = np.array([[[1, 2, 3], [4, 5, 6]]], dtype=np.int64)
```
jignparm: Can this also include  tests for matrices of slightly higher ranks (e.g.  rank =3 and rank = 4)?

tests/test_backend.py
```diff
@@ -3010,6 +3010,13 @@ def test_hashtable_lookup(self):
         self._run_test_case([lookup_results.name], {_INPUT: query})
         os.remove(filnm)
 
+    @check_opset_min_version(11)
+    def test_MatrixDiagPart(self):
+        input_val = np.array([[[1, 2, 3, 4, 5], [6, 7, 8, 9, 10], [11, 12, 13, 14, 15], [16, 17, 18, 19, 20]]]).astype(
+            np.int64)
```
jignparm: Is it possible to test out higher dimensional tensors as well? The current tensor seems 2Dimensional ( 4x4).

Maybe we can try random values in shapes of  [ (n1, n2, n3),   (n1, n2, n3, n4),  (n1, n2, n3, n4, n5)] in a loop.

Also in the above it might be good to to ensure that  n5 is **smaller** than n4 above -- since there's a slight variation in the output shape based on which of the last 2 dimensions is larger.


```
https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/matrix-diag-part

Output: The extracted diagonal(s) having shape diagonal.shape = input.shape[:-2] + [min(input.shape[-2:])].
```

RandySheriffH: Good point, let's add them.

tests/test_backend.py
```diff
@@ -3010,6 +3010,26 @@ def test_hashtable_lookup(self):
         self._run_test_case([lookup_results.name], {_INPUT: query})
         os.remove(filnm)
 
-
+    @check_opset_min_version(11)
+    def test_MatrixDiagPart(self):
```
jignparm: nit -- can we rename the test to "test_matrix_diag_part", to make it follow the convention of the tests? Some other tests are also misspelled, but would be good to keep convention.

RandySheriffH: Sure why not.

Issue #787, opened 2020-02-05
Support tf.keras.layers.CudnnGRU


THULiusj:
**Describe the bug**
I have a model using tf.keras.layers.CudnnGRU op. I try to transform the TF model to onnx, and it reports error:
2019-12-25 17:37:20,725 - ERROR - Tensorflow op [resnet_ada/adapter/general/rnn_adapter1/ReverseV2: ReverseV2] is not supported
2019-12-25 17:37:20,741 - ERROR - Tensorflow op [resnet_ada/adapter/general/rnn_adapter1/CudnnRNN_1: CudnnRNN] is not supported
2019-12-25 17:37:20,741 - ERROR - Tensorflow op [resnet_ada/adapter/general/rnn_adapter1/ReverseV2_1: ReverseV2] is not supported
2019-12-25 17:37:20,746 - ERROR - Tensorflow op [resnet_ada/adapter/general/rnn_adapter1/CudnnRNN: CudnnRNN] is not supported
2019-12-25 17:37:20,774 - ERROR - Tensorflow op [resnet_ada/adapter/general/rnn_adapter2/CudnnRNN: CudnnRNN] is not supported
2019-12-25 17:37:20,774 - ERROR - Tensorflow op [resnet_ada/adapter/general/rnn_adapter2/ReverseV2: ReverseV2] is not supported
2019-12-25 17:37:20,774 - ERROR - Tensorflow op [resnet_ada/adapter/general/rnn_adapter2/CudnnRNN_1: CudnnRNN] is not supported
2019-12-25 17:37:20,775 - ERROR - Tensorflow op [resnet_ada/adapter/general/rnn_adapter2/ReverseV2_1: ReverseV2] is not supported

**Urgency**
I want to publish the model to production but blocked by this issue. I also communicated with Nvidia contact. And they replied that this op should be supported by tf2onnx first.

**System information**
- Tensorflow Version: 1.13+
- Python version: 3.5+

**Expected behavior**
tf2onnx supports tf.keras.layers.CudnnGRU to be transformed to onnx

guschmue:
ReverseV2 is supported if you use ```--opset 11```. 
CudnnRNN we don't support right now. Let me check if we can add this.

THULiusj:
@guschmue Hi, how about the status of CudnnRNN?

guschmue:
have not gotten to it yet.

RandySheriffH:
Adding support now, will share PR with ready.

[...]

Issue #683, opened 2019-09-01
IndexError: list index out of range


stiv-yakovenko:
I am trying to convert frozen .pb graph to onnx format on Ubuntu x64 Linux with nvidia GPU:

python3 -m tf2onnx.convert --input ./lpr_it0.pb --output ./lpr.onnx --outputs=softmax/truediv

and getting this crash
```
/usr/local/lib/python3.6/dist-packages/tensorflow/python/framework/dtypes.py:516: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_qint8 = np.dtype([("qint8", np.int8, 1)])
/usr/local/lib/python3.6/dist-packages/tensorflow/python/framework/dtypes.py:517: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_quint8 = np.dtype([("quint8", np.uint8, 1)])
/usr/local/lib/python3.6/dist-packages/tensorflow/python/framework/dtypes.py:518: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_qint16 = np.dtype([("qint16", np.int16, 1)])
/usr/local/lib/python3.6/dist-packages/tensorflow/python/framework/dtypes.py:519: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_quint16 = np.dtype([("quint16", np.uint16, 1)])
/usr/local/lib/python3.6/dist-packages/tensorflow/python/framework/dtypes.py:520: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  _np_qint32 = np.dtype([("qint32", np.int32, 1)])
/usr/local/lib/python3.6/dist-packages/tensorflow/python/framework/dtypes.py:525: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  np_resource = np.dtype([("resource", np.ubyte, 1)])
[...]

PR #819, opened 2020-03-02
Rashuai/cudnn gru

RandySheriffH:
Address issue:

- https://github.com/onnx/tensorflow-onnx/issues/787
- https://github.com/onnx/tensorflow-onnx/issues/683

Both issues require fix for ReverseV2 and CudnnGRU.


tests/test_backend.py
```diff
@@ -2932,6 +2933,27 @@ def func(input_holder):
         for input_val in input_vals:
             self._run_test_case(func, [_OUTPUT], {_INPUT: input_val})
 
+    @skip_tf_cpu("only tf_gpu can run CudnnGPU")
```
guschmue: Those tests are significant different compared to what is in test_backend.py that they should be in their own file, ie. test_cudnn.py ?

tests/test_backend.py
```diff
@@ -2932,6 +2933,27 @@ def func(input_holder):
...
+    @check_opset_min_version(11, "GRU")
```
guschmue: why opset 11?

tests/test_backend.py
```diff
@@ -2932,6 +2933,27 @@ def func(input_holder):
...
+    def test_cudnngru(self):
+        def func():
+            seq_length = 3
+            batch_size = 5
+            input_size = 2
+            num_layers = 2
+            num_units = 2
+            num_dirs = 2
+            initializer = init_ops.constant_initializer(0.5)
+            x = np.random.randint(0, 100, [seq_length, batch_size, input_size]).astype(np.float32)
```
guschmue: You want random badly to be outside of func() because if you run this against tf and than onnxruntime they will be different. Better to pass them in arguments to func()

tests/test_backend.py
```diff
@@ -2932,6 +2933,27 @@ def func(input_holder):
         for input_val in input_vals:
             self._run_test_case(func, [_OUTPUT], {_INPUT: input_val})
 
+    @skip_tf_cpu("only tf_gpu can run CudnnGPU")
+    @check_opset_min_version(11, "GRU")
+    def test_cudnngru(self):
+        def func():
+            seq_length = 3
```
guschmue: I'd move all of those assignments outside of func and just keep the tf code in func()

tests/test_backend.py
```diff
@@ -2932,6 +2933,27 @@ def func(input_holder):
...
+            batch_size = 5
+            input_size = 2
+            num_layers = 2
+            num_units = 2
+            num_dirs = 2
+            initializer = init_ops.constant_initializer(0.5)
+            x = np.random.randint(0, 100, [seq_length, batch_size, input_size]).astype(np.float32)
+            h = np.random.randint(0, 100, [num_layers * num_dirs, batch_size, num_units]).astype(np.float32).reshape(
+                [num_layers * num_dirs, batch_size, num_units])
+            cudnngru = tf.contrib.cudnn_rnn.CudnnGRU(num_layers, num_units, 'linear_input', 'bidirectional',
```
guschmue: there is no such thing as contrib under tf-2.x

tf2onnx/onnx_opset/rnn.py
```diff
@@ -10,15 +10,14 @@
...
-
 import numpy as np
 from tf2onnx import utils
 from tf2onnx.handler import tf_op
 
 logger = logging.getLogger(__name__)
 
 
-# pylint: disable=unused-argument,missing-docstring
+# pylint: disable=unused-argument,missing-docstring,invalid-name
```
guschmue: can you move this behind variable that triggers this, like 'var  # pylint: disable=...' so it does not apply to the full file.

tf2onnx/onnx_opset/rnn.py
```diff
@@ -169,3 +168,84 @@ def replace_output(old_output, new_output):
     @classmethod
     def version_7(cls, ctx, node, **kwargs):
         cls.version_1(ctx, node, **kwargs)
+
+
+@tf_op("CudnnRNN")
+class CudnnRNN:
+    @classmethod
+    def version_11(cls, ctx, node, **kwargs):
```
guschmue: why opset 11?

RandySheriffH: Need to rewrite function body to cater opset 10 ops - slice for example is the main one that has differed specification between 10 and 11. Shall we?

tests/test_backend.py
RandySheriffH: Tested - we could support from opset 10 with what we have already.

tf2onnx/onnx_opset/rnn.py
RandySheriffH: Lowered to opset 10, current code works fine.

PR #828, opened 2020-03-05
Fix strided slice, reverse stride with unbounded begin and end

### LINKED ISSUE #810: Possible conversion bug for StridedSlice with negative stride and implicit bounds
**Describe the bug**
When a tensorflow model contains a slicing operation with a negative stride and an implicit bound, `onnxruntime` fails on the onnx model produced by `tf2onnx`.

**Urgency**
None.

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Windows 10 1903
- Tensorflow Version: 1.14.0
- Python version: 3.7.6

**To Reproduce**
1. Create a TF model
```python
import tensorflow as tf
ph = tf.placeholder(tf.float32, (1, 16, 32, 1), name='ph')
x = tf.concat([ph[:, :, :10], ph[:, :, 9::-1]], axis=0)
y = tf.add(x, 1, name='y')
g = tf.get_default_graph().as_graph_def()
tf.train.write_graph(g, '.', 'flip.pb', as_text=False)
```
2. Convert to ONNX
```bash
python -m tf2onnx.convert --input flip.pb --output flip.onnx --inputs ph:0 --outputs y:0 --opset 11
```

3. Run inference
```python
import onnxruntime as rt
sess = rt.InferenceSession('flip.onnx')
# onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : Node (concat) Op (Concat) [ShapeInferenceError] Can't merge shape info. Both source and target dimension have values but they differ. Source=0 Target=10 Dimension=2
```

**Expected behavior**
I would expect no error

**Screenshots**

**Additional context**
The same problem occurs if the begin value is missing:
```python
x = tf.concat([ph[:, :, :10], ph[:, :, :21:-1]], axis=0)
```

However, if both begin *and* end values are provided, the model works fine:
```python
x = tf.concat([ph[:, :, :10], ph[:, :, 10:0:-1]], axis=0)  # works fine
```

It seems that the problem comes from the fact that default value of `0` and `1e9` are set for begin/end implicit bounds in [`StrideSlice.version_10`](https://github.com/onnx/tensorflow-onnx/blob/v1.5.5/tf2onnx/onnx_opset/tensor.py#L738), not taking into account the sign of the stride. A fix could perhaps be to switch those values when the stride is negative, or maybe use the actual shape values of the input, `ctx.get_shape(input_x.output[0])`?

jignparm:
Fixes #810 

The current implementation of strided slice ignored the direction of the slice (i.e. backward or forward) when setting values for unbounded 'begins' or 'ends'. The values need to be reversed if slice direction is backwards.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1788,6 +1788,32 @@ class BackendTests(Tf2OnnxBackendTestBase):
         self._run_test_case(func, [_OUTPUT], {_INPUT: x_val, _INPUT1: y_val})
 
     @check_opset_min_version(10, "Slice")
+    def test_strided_slice_reverse_1(self):
+        tf.reset_default_graph()
+        x_val = np.arange(16 * 32).astype(np.float32).reshape((1, 16, 32, 1))
+        def func(x):
+            return tf.concat([x[:, :, :10], x[:, :, :21:-1]], axis=0, name=_TFOUTPUT)
+        self._run_test_case(func, [_OUTPUT], {_INPUT: x_val})
+
+    @check_opset_min_version(10, "Slice")
+    def test_strided_slice_reverse_2(self):
+        tf.reset_default_graph()
+        x_val = np.arange(16 * 32).astype(np.float32).reshape((1, 16, 32, 1))
+        def func(x):
+            return tf.concat([x[:, :, :10], x[:, :, 9::-1]], axis=0, name=_TFOUTPUT)
+        self._run_test_case(func, [_OUTPUT], {_INPUT: x_val})
+
+    @check_opset_min_version(10, "Slice")
+    def test_strided_slice_reverse_3(self):
+        tf.reset_default_graph()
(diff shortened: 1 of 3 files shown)


tests/test_backend.py
```diff
@@ -1787,6 +1787,32 @@ def func(x, y):
             return tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case(func, [_OUTPUT], {_INPUT: x_val, _INPUT1: y_val})
 
+    @check_opset_min_version(10, "Slice")
+    def test_strided_slice_reverse_1(self):
+        tf.reset_default_graph()
```
guschmue: use tf_reset_default_graph() because of tf-2.x. 
But it is not needed because we do it implicit before calling func.

jignparm: Thanks -- removed.

PR #830, opened 2020-03-09
switch to contrib cropandresize

RandySheriffH:
Use contrib cropresize to replace assembled one.

### DIFF
--- a/tf2onnx/onnx_opset/nn.py
+++ b/tf2onnx/onnx_opset/nn.py
@@ -587,67 +587,6 @@ class DepthToSpace:
         cls.version_1(ctx, node, **kwargs)
 
 
-@tf_op(["CropAndResize"])
-class CropAndResize:
-    @classmethod
-    def version_11(cls, ctx, node, **kwargs):
-        # create loop of resize to cater to tensorflow CropAndResize, one box one iteration
-        mode = "nearest" if node.get_attr("method") is not None and node.get_attr(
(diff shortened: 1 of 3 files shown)


tf2onnx/onnx_opset/nn.py
```diff
@@ -587,67 +587,6 @@ def version_11(cls, ctx, node, **kwargs):
         cls.version_1(ctx, node, **kwargs)
 
 
-@tf_op(["CropAndResize"])
```
guschmue: tf2onnx does not depend on onnxruntime or contrib ops. You need to check if the ms namespace is enabled.

RandySheriffH: Sure, but how to verify if the namespace enabled?

guschmue: So by default we use onnx opset. Than we override with the extra-opsets.
So you'd add it to ms.py but keep the original code in controflow.py for runtimes that don't support the ms namespace.
For example the Range op:
https://github.com/onnx/tensorflow-onnx/blob/master/tf2onnx/onnx_opset/controlflow.py#L256
https://github.com/onnx/tensorflow-onnx/blob/master/tf2onnx/custom_opsets/ms.py#L32

RandySheriffH: @guschmue, restored nn.py to keep assembled one.

PR #840, opened 2020-03-13
Push transposes past mul ops to optimize mobilenet_v3

jignparm:
There are  too many residual transpose operators in mobilenet_v3 model. This fix eliminates vast majority of them.


guschmue:
Not sure why the ci pipleline doesn't trigger. Looking at your branch it says:
```This branch is 1 commit ahead, 22 commits behind onnx:master.```
Don't know where is onnx:master comes from ... can you check on it ?

**Comment by jignparm** [comment]
I believe `onnx:master` is the prefix in the URL of this repository. I merged my changes with the master branch (to pull in any recent changes), but still  the pipelines seem to have a problem  running.

The message from the bot seems to thing a branch/path is excluded. Maybe some include/exclude patterns require to be updated?

```
Azure Pipelines could not run because the pipeline triggers exclude this branch/path.
```

PR #843, opened 2020-03-16
Create a rewriter for fusing Conv2d with BiasAdd

PreethaVeera:
In the current conversion flow, the BiasAdd is fused with the Conv2D / ConvBackPropInput during the transpose_optimisation of NHWC based models. But when we have a NCHW based model or when the optimisation is turned off, this fusion will not happen. It is a much needed optimisation and hence if it is a part of rewriters, the fusion of Conv and BiasAdd is ensured.

This PR consists of the rewriter conv2d_with_add_rewriter that fuses Conv2d with BiasAdd.


guschmue:
Hello, thanks for your contribution!
We has some issue with triggers for the ci pipeline. Could you merge master / push so it triggers the CI pipeline ?

PreethaVeera:
Hi @guschmue . I have re-triggered the checks and all the tests passed. Can you kindly accept this PR? Thanks.

guschmue:
super, thanks!

PR #848, opened 2020-03-17
test

guschmue:


guschmue:
@jignparm for some reason the master was busted without the build-batch indicating it.
Maybe this is the reason why your PR doesn't trigger the ci pipeline.
You could review/merge this PR and than merge master into your PR, hope that will trigger the pipeline.

PR #851, opened 2020-03-18
Add support for Select op with Opset 7

PreethaVeera:
Currently Select op is supported only from onnx opset 8 with the help of Loop ops. This PR makes Select op to be compatible with opset 7  by mapping it to certain simpler ops in ONNX as below. 

**Select**  === >  **Add**(**Multiply**(**Cast**(bool input_0, Float32 ), input_1),  **Multiply**(**Cast**(**Not**(bool input_0), float32), input_2)).


guschmue:
Super, thanks!

Issue #824, opened 2020-03-03
DepthwiseConv2DNative model -- converts successfully, but error at run-time


jignparm:
**Describe the bug**
A TF model using DepthwiseConv2DNative  operator converts successfully, but ORT throws error at runtime.

**Urgency**
Not sure

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Windows
- Tensorflow Version: TF 1.15
- Python version: Any

**To Reproduce**
Convert the model here -> https://drive.google.com/file/d/1SU-pEb93eLQoRMxaK5z80jyjemZ0YuIn/view?usp=sharing.

** Error Message **
```
2020-03-03 18:45:35.1126365 [E:onnxruntime:, sequential_executor.cc:184 onnxruntime::SequentialExecutor::Execute] Non-zero status code returned while running Conv node. Name:'FeatureExtractor/MobilenetV2/expanded_conv/depthwise/depthwise' Status Message: Input channels C is not equal to kernel channels * group. C: 34 kernel channels: 1 group: 32
2020-03-03 18:45:35,113 - ERROR - run_pretrained: Run_ONNX FAIL
Traceback (most recent call last):
  File "D:/jignparmtf2onnx/tests/run_pretrained_models.py", line 302, in run_test
    onnx_results = self.run_onnxruntime(name, model_proto, inputs)
  File "D:/jignparmtf2onnx/tests/run_pretrained_models.py", line 189, in run_onnxruntime
    results = m.run(self.output_names, inputs)
  File "d:\onnxruntime5\build\Windows\Debug\Debug\onnxruntime\capi\session.py", line 142, in run
    return self._sess.run(output_names, input_feed, run_options)
onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : Non-zero status code returned while running Conv node. Name:'FeatureExtractor/MobilenetV2/expanded_conv/depthwise/depthwise' Status Message: Input channels C is not equal to kernel channels * group. C: 34 kernel channels: 1 group: 32

```

**More Information**
https://github.com/microsoft/onnxruntime/issues/3114

guschmue:
What is different with this model ?
This (aka the standard mobilenet-v2)
```
python tests\run_pretrained_models.py --tests mobilenet_v2_1.4_224 --opset 11
```
works for me on ort 1.1.

qraleq:
[...]

PR #854, opened 2020-03-19
prevent multi-transpose on pad

RandySheriffH:
Fixing issue:
https://github.com/onnx/tensorflow-onnx/issues/824
When const perm tensor getting consumed by multiple pad op, it gets transposed over and over again, finally go back to its starting value.

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -556,12 +556,14 @@ class TransposeOptimizer(GraphOptimizerBase):
             new_pads = [pads[0], pads[3], pads[1], pads[2], pads[4], pads[7], pads[5], pads[6]]
             node.set_attr("pads", new_pads)
             return self._switch_transpose_and_node(node, trans)
-        if node.inputs[1].is_const() and self._nodes_has_single_consumer_node([node.inputs[1]]):
-            pads = node.inputs[1].get_tensor_value()
-            # NHWC->NCHW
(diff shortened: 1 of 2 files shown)


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -556,7 +556,7 @@ def _pad_handler(self, trans, node):
             new_pads = [pads[0], pads[3], pads[1], pads[2], pads[4], pads[7], pads[5], pads[6]]
             node.set_attr("pads", new_pads)
             return self._switch_transpose_and_node(node, trans)
-        if node.inputs[1].is_const():
+        if node.inputs[1].is_const() and self._nodes_has_single_consumer_node([node.inputs[1]]):
```
jignparm: Not sure if this is masking the issue, or solving the issue.

Is it possible to add a UT that has  one Transpose feeding into 2  Pads followed by 2 transposes?  The transposes should be  set up such that they cancel out, leaving 0 transposes at the end of the optimization.


###Starting graph
```script
Input 1  --> Trans1 --> Pad 1 --> Trans2 --> Output1
                    --> Pad 2 --> Trans3 --> Output2
```
  

###Ending graph
```script
Input 1  --> Pad 1 --> Output1
         --> Pad 2 --> Output2
```

RandySheriffH: Switching the approach - allowing it to continue no matter how many consumers there but keep the tensor from overly tranposed.

PR #857, opened 2020-03-20
Add support for crop_and_resize for opset 10.

PreethaVeera:
Support for Crop and resize for opset 10 is implemented if its inputs boxes and crop_size are Constant. 
Currently only opset 11 is supported. If this version is also added to master, it will be helpful. 
Thanks.


guschmue:
super, thanks!!

PR #858, opened 2020-03-21
Add missing import in tf<2

andhus:
The module `function_def_to_graph` is not imported with TensorFlow version < 2, it is however used here: https://github.com/onnx/tensorflow-onnx/pull/858/files#diff-0445eb7578a5f75813867ce2784e7941R370 and this function is used when I try to convert a model using TF 1.14.0. Unfortunately, I don't have time to time to add a self-contained failing example right now (will try to open an issue later) but seems like an obvious bug, or?

### DIFF
--- a/tf2onnx/tf_loader.py
+++ b/tf2onnx/tf_loader.py
@@ -41,7 +41,7 @@ try:
     from tensorflow.python.framework.function_def_to_graph import function_def_to_graph
 except ImportError:
     function_def_to_graph = _not_implemented_tf_placeholder('function_def_to_graph')
-    
+
 if is_tf2():
     convert_variables_to_constants = tf.compat.v1.graph_util.convert_variables_to_constants
     from tensorflow.python.framework.convert_to_constants import convert_variables_to_constants_v2
--- a/tf2onnx/tf_loader.py
+++ b/tf2onnx/tf_loader.py
@@ -29,6 +29,7 @@ if is_tf2():
     from tensorflow.python.framework import convert_to_constants, func_graph, function_def_to_graph
 else:
     from tensorflow.python.framework.graph_util import convert_variables_to_constants
+    from tensorflow.python.framework import function_def_to_graph
 
 
 if is_tf2():
--- a/tf2onnx/tf_loader.py
+++ b/tf2onnx/tf_loader.py
@@ -21,15 +21,33 @@ logger = logging.getLogger(__name__)
 # pylint: disable=unused-argument,unused-import,no-value-for-parameter,unexpected-keyword-arg,ungrouped-imports
 # pylint: disable=missing-function-docstring,import-outside-toplevel,useless-import-alias,missing-docstring
 
+
 def is_tf2():
     return tf.__version__.startswith("2.")
 
 
+def _not_implemented_tf_placeholder(name):
+    """Creates a placeholder function for missing Tensorflow imports"""
+
+    def not_implemented_tf_placeholder(*args, **kwargs):
+        raise NotImplementedError(
+            f'Tensorflow verison {tf.__version__} does not implement '
+            f'`{name}`, try converting your model with a different version.'
+        )
+    return not_implemented_tf_placeholder
+
+
+try:
+    from tensorflow.python.framework.function_def_to_graph import function_def_to_graph
+except ImportError:
+    function_def_to_graph = _not_implemented_tf_placeholder('function_def_to_graph')
+    
 if is_tf2():
-    from tensorflow.python.framework import convert_to_constants, func_graph, function_def_to_graph
+    convert_variables_to_constants = tf.compat.v1.graph_util.convert_variables_to_constants
+    from tensorflow.python.framework.convert_to_constants import convert_variables_to_constants_v2
 else:
     from tensorflow.python.framework.graph_util import convert_variables_to_constants
-    from tensorflow.python.framework import function_def_to_graph
+    convert_variables_to_constants_v2 = _not_implemented_tf_placeholder('convert_variables_to_constants_v2')
 
 
 if is_tf2():
@@ -64,7 +82,7 @@ else:
 
 
 def from_function(func, input_names, output_names):
-    frozen_func = convert_to_constants.convert_variables_to_constants_v2(func, lower_control_flow=False)
+    frozen_func = convert_variables_to_constants_v2(func, lower_control_flow=False)
     graph_def = frozen_func.graph.as_graph_def(add_shapes=True)
     # output_tensors = {i.name: i for i in frozen_func.outputs}
     tf_reset_default_graph()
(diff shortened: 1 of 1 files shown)


jignparm:
So far haven't  needed to import `function_def_to_graph`  in either TF 1.14 or  TF 2.1, using the CPU packages.

Maybe it's needed in the GPU versions of TF?

andhus:
No, the issue is really clear just by following the stack of calls: `tf2onnx.convert.main` always calls `tf2onnx.tfonnx.process_tf_graph` which then always calls `tf2onnx.tf_loader.resolve_functions` which uses the unimported module (with tf<2) if `tf_graph._functions.items()` is not empty. So the assumption currently must be that `tf_graph._functions.items()` is always empty with tf<2?? If that is the case it is very implicit in the code...

It seems to be empty mostly (otherwise the conversion wouldn't work at all in tf < 2). But one way to trigger the issue is to add a custom function through tf.keras API:

```
import os
import tempfile
import subprocess

import onnx
import tensorflow as tf  # 1.14.0


def get_onnx_proto(model: tf.keras.Model, signatures=None,):
    with tempfile.TemporaryDirectory() as model_dir:
        tf.keras.experimental.export_saved_model(
            model,
            model_dir,
            input_signature=signatures
        )
        onnx_path = os.path.join(model_dir, 'model.onnx')
        cmd = (
            'python -m tf2onnx.convert '
            f'--saved-model {model_dir} '
            f'--output {onnx_path}'
        )
        return_code = subprocess.call(cmd.split(' '))
        if return_code > 0:
            raise RuntimeError('tf2onnx failed to convert model')

        with open(onnx_path, 'rb') as f:
            model_proto = onnx.load(f)

        return model_proto


def swish(x):
    return tf.nn.swish(x)


tf.keras.utils.get_custom_objects().update({'swish': swish})


if __name__ == '__main__':
     x = tf.keras.Input(shape=(3,), name='input')
     y = tf.keras.layers.Dense(2, activation=swish)(x)
     model = tf.keras.Model(x, y)
     proto = get_onnx_proto(model)
```
Currently (48fdca505baad401a0fc69332978ab9b5beeea3e) this crashes with log output:

```
2020-03-22 14:04:14,083 - INFO - Using tensorflow=1.14.0, onnx=1.6.0, tf2onnx=1.6.0/8b4ca0
2020-03-22 14:04:14,084 - INFO - Using opset <onnx, 8>
  File "/usr/local/Cellar/python/3.6.5_1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python/3.6.5_1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/andershuss/Develop/external/tensorflow-onnx/tf2onnx/convert.py", line 169, in <module>
    main()
  File "/Users/andershuss/Develop/external/tensorflow-onnx/tf2onnx/convert.py", line 153, in main
    inputs_as_nchw=args.inputs_as_nchw)
  File "/Users/andershuss/Develop/external/tensorflow-onnx/tf2onnx/tfonnx.py", line 393, in process_tf_graph
    ordered_func = resolve_functions(tf_graph)
  File "/Users/andershuss/Develop/external/tensorflow-onnx/tf2onnx/tf_loader.py", line 370, in resolve_functions
    func = function_def_to_graph.function_def_to_graph(fdef, input_shapes=input_shapes)
NameError: name 'function_def_to_graph' is not defined
```
With the fix in this PR you get:
```
2020-03-22 14:08:56,832 - INFO - Using tensorflow=1.14.0, onnx=1.6.0, tf2onnx=1.6.0/8b4ca0
2020-03-22 14:08:56,832 - INFO - Using opset <onnx, 8>
2020-03-22 14:08:56,844 - WARNING - Shape of placeholder features is unknown, treated it as a scalar
2020-03-22 14:08:56,845 - WARNING - Shape of placeholder grad is unknown, treated it as a scalar
2020-03-22 14:08:56,853 - WARNING - Shape of placeholder features is unknown, treated it as a scalar
2020-03-22 14:08:56,860 - ERROR - Tensorflow op [dense/swish_f32: swish_f32] is not supported
2020-03-22 14:08:56,860 - ERROR - Unsupported ops: Counter({'swish_f32': 1})
2020-03-22 14:08:56,861 - INFO - Optimizing ONNX model
2020-03-22 14:08:56,870 - INFO - After optimization: Identity -1 (1->0)
2020-03-22 14:08:56,871 - INFO - 
2020-03-22 14:08:56,871 - INFO - Successfully converted TensorFlow model /var/folders/7v/xm_lv4311ljdl1c_gg0xcp8r0000gn/T/tmp4nv3tppn to ONNX
2020-03-22 14:08:56,872 - INFO - ONNX model is saved at /var/folders/7v/xm_lv4311ljdl1c_gg0xcp8r0000gn/T/tmp4nv3tppn/model.onnx
```
I understand that the conversion still fails because of the custom function, but that can be worked around with extra opset and the later log output is what is wanted, not the `NameError: name 'function_def_to_graph' is not defined`.

guschmue:
Yes, makes sense, we found a function on some mode in the graph and try to resolve it.

tf2onnx/tf_loader.py
```diff
@@ -29,6 +29,7 @@ def is_tf2():
     from tensorflow.python.framework import convert_to_constants, func_graph, function_def_to_graph
 else:
     from tensorflow.python.framework.graph_util import convert_variables_to_constants
+    from tensorflow.python.framework import function_def_to_graph
```
guschmue: Seems ok but I'm worried this might not be there in super old tensorflow versions. In theory we don't support them anymore but in practice we see people still using things like tf-1.4.
Could we move this import a few lines lower into the 
elif LooseVersion(tf.__version__) >= "1.13"
block?

andhus: Well sure, but then you get the same very uninformative exception in tf 1.12 (and the next person will have to redo my debugging :/), and support is claimed for "tf-1.12 and up".

I think it is a reasonable principle that references to variables that do not exist (NameError) should never happen - it is a bug in your code. The problem is that the responsibility for handling various tf-version is unstructured and distributed across modules: callers of `tf_loader` will have to know that some functions implemented in that module are only usable depending on tf version etc. I updated the PR with a suggestion that will at least give an informative error when this happens.

andhus:
I think it is a reasonable principle that references to variables that do not exist (NameError) should never happen - it is a bug in your code. The problem is that the responsibility for handling various tf-version is unstructured and distributed across modules: callers of `tf_loader` will have to know that some functions implemented in that module are only usable depending on tf version etc. I updated the PR with a suggestion that will at least give an informative error when this happens.

tf2onnx/tf_loader.py
guschmue: I agree, The support for multiple versions is not optimal. We tried to isolate it in tf_loader, tf_util  and unit tests (later is a real royal pain) but forgot that functions can show up before tf2.
Adding a more formal interface to handle multiple version: there are still some issues with tf2 we are working on and once that is done we'll find some time to restructure tf_loader to make this cleaner.

guschmue:
Thanks for the contribution!

andhus:
Cool, thanks for the quick response!

Issue #859, opened 2020-03-23
Request for NonMaxSuppressionV5 support


jf99:
While trying to convert a tensorflow frozen graph to ONNX, I got an error message that NMS V5 is not supported by tf2onnx:
> 2020-03-23 11:13:41,012 - ERROR - Tensorflow op [Postprocessor/BatchMultiClassNonMaxSuppression/map/while/MultiClassNonMaxSuppression/non_max_suppression_with_scores/NonMaxSuppressionV5: NonMaxSuppressionV5] is not supported
2020-03-23 11:13:41,058 - ERROR - Unsupported ops: Counter({'NonMaxSuppressionV5': 1})

According to [Troubleshooting](https://github.com/onnx/tensorflow-onnx/blob/master/Troubleshooting.md) one should create an issue or PR in such a case. Since I lack of resources and python skill to do a PR, I chose to open this issue. Would be nice if you implemented NMS V5 conversion :-).

guschmue:
Yes, we need to add  V5. Looking briefly at it - looks like soft_nms_sigma soft_nms_sigma have been added. soft_nms_sigma might need an onnx change.
How do you call tf.image.non_max_suppression*() ?

hashishoya:
Hi @guschmue, I'm a colleague of @jf99 and we used Tensorflow Object Detection API for the training of the network. In our case, we just tried first to convert the pretrained [ssd_inception_v2_coco](http://download.tensorflow.org/models/object_detection/ssd_inception_v2_coco_2018_01_28.tar.gz) model to ONNX so we can use it later with TensorRT .. 

So I'm not sure how tf.image.non_max_suppression was used in their pre-trained model ..

guschmue:
no worry, @RandySheriffH has a PR to support V4 and V5 as long soft_nms is not used. Hope with get away with it for now.

RandySheriffH:
Fix merged to master, feel free to come to us for further assistance.

PR #860, opened 2020-03-25
Convert NMS V4&V5

RandySheriffH:
Address issue https://github.com/onnx/tensorflow-onnx/issues/859
Support V4 and V5(partially)

### DIFF
--- a/tf2onnx/onnx_opset/tensor.py
+++ b/tf2onnx/onnx_opset/tensor.py
@@ -1478,7 +1478,7 @@ class NonMaxSuppression:
         slice_op = GraphBuilder(ctx).make_slice({"data": new_nonmaxsurppress,
                                                  "axes": [1], "ends": [3], "starts": [2]})
         squeeze_op = ctx.make_node("Squeeze", [slice_op], attr={"axes": [1]})
-        if len(node.input) > 5:  # V5
+        if len(node.input) > 5:  # v5, called by ..._with_scores(), pad_to_max_output_size always False
             ctx.make_node("Cast", inputs=squeeze_op.output, attr={"to": onnx_pb.TensorProto.INT32},
                           outputs=[node.output[0]], dtypes=dtypes[0], shapes=shapes[0])
             ctx.make_node("Gather", inputs=[input_score.input[0], squeeze_op.output[0]],
--- a/tf2onnx/onnx_opset/tensor.py
+++ b/tf2onnx/onnx_opset/tensor.py
@@ -1471,13 +1471,14 @@ class NonMaxSuppression:
         dtypes = [[ctx.get_dtype(output)] for output in node.output]
         shapes = [[ctx.get_shape(output)] for output in node.output]
         max_output_size = node.input[2]
+        utils.make_sure(len(node.inputs) <= 5 or int(node.inputs[5].get_tensor_value(False)) == 0,
+                        "soft_nms_sigma must be 0")
         ctx.remove_node(node.name)
         new_nonmaxsurppress = ctx.make_node(node.type, node.input[: 5]).output[0]
         slice_op = GraphBuilder(ctx).make_slice({"data": new_nonmaxsurppress,
                                                  "axes": [1], "ends": [3], "starts": [2]})
         squeeze_op = ctx.make_node("Squeeze", [slice_op], attr={"axes": [1]})
         if len(node.input) > 5:  # V5
-            logger.warning("NonMaxSuppressionV5 only parltially supported, soft_nms_sigma must be 0.0")
             ctx.make_node("Cast", inputs=squeeze_op.output, attr={"to": onnx_pb.TensorProto.INT32},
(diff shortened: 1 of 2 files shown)


tf2onnx/onnx_opset/tensor.py
```diff
@@ -1464,18 +1465,40 @@ def version_10(cls, ctx, node, **kwargs):
...
         ctx.remove_node(node.name)
-        new_nonmaxsurppress = ctx.make_node(node.type, node.input).output[0]
+        new_nonmaxsurppress = ctx.make_node(node.type, node.input[: 5]).output[0]
         slice_op = GraphBuilder(ctx).make_slice({"data": new_nonmaxsurppress,
                                                  "axes": [1], "ends": [3], "starts": [2]})
         squeeze_op = ctx.make_node("Squeeze", [slice_op], attr={"axes": [1]})
-        ctx.make_node("Cast", inputs=squeeze_op.output, attr={"to": onnx_pb.TensorProto.INT32},
-                      name=node.name, outputs=node.output, dtypes=dtypes, shapes=shapes)
+        if len(node.input) > 5:  # V5
+            logger.warning("NonMaxSuppressionV5 only parltially supported, soft_nms_sigma must be 0.0")
```
jignparm: Is soft_nms_sigma usually 0.0? 

Is there a way to check the value, and then throw the warning? -- it seems like this will throw a warning at every V5 node at conversion time, and if the user ignores the warning, at runtime they'll just get a wrong result. Maybe I'm reading this wrong however.

RandySheriffH: Of course it could by any designated value.
Better to throw an exception during conversion on non-zero soft_nms_sigma, but it is an input meaning could be a non-const.

jignparm: How does the code snippet below look instead? It should throw exception if it cannot handle the ops, otherwise the generated model should be guaranteed to work.

```python
utils.make_sure(node.inputs[index], "node must be const")
val = node.inputs[index].get_tensor_value(False)
utils.make_sure(val == 0, "err msg...")
```

RandySheriffH: Done, :)

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1464,18 +1465,41 @@ def version_10(cls, ctx, node, **kwargs):
...
+        input_score = ctx.insert_new_node_on_input(node, "Unsqueeze", node.input[1], axes=[0, 1])
         ctx.insert_new_node_on_input(node, "Cast", node.input[2], to=onnx_pb.TensorProto.INT64)
         # replace original node with nonmaxsurppress + slice + squeeze + cast
-        dtypes = [ctx.get_dtype(node.output[0])]
-        shapes = [ctx.get_shape(node.output[0])]
+        dtypes = [[ctx.get_dtype(output)] for output in node.output]
+        shapes = [[ctx.get_shape(output)] for output in node.output]
+        max_output_size = node.input[2]
+        utils.make_sure(len(node.inputs) <= 5 or int(node.inputs[5].get_tensor_value(False)) == 0,
+                        "soft_nms_sigma must be 0")
```
jignparm: is node.inputs[5] guaranteed to be const? If not, should check that it's const before getting tensor value.
If it's not const, should throw exception.

RandySheriffH: get_tensor_value throws exception on non-const case in the very beginning.

tf2onnx/onnx_opset/tensor.py
```diff
@@ -1464,18 +1465,41 @@ def version_10(cls, ctx, node, **kwargs):
...
                                                  "axes": [1], "ends": [3], "starts": [2]})
         squeeze_op = ctx.make_node("Squeeze", [slice_op], attr={"axes": [1]})
-        ctx.make_node("Cast", inputs=squeeze_op.output, attr={"to": onnx_pb.TensorProto.INT32},
-                      name=node.name, outputs=node.output, dtypes=dtypes, shapes=shapes)
+        if len(node.input) > 5:  # V5
+            ctx.make_node("Cast", inputs=squeeze_op.output, attr={"to": onnx_pb.TensorProto.INT32},
+                          outputs=[node.output[0]], dtypes=dtypes[0], shapes=shapes[0])
+            ctx.make_node("Gather", inputs=[input_score.input[0], squeeze_op.output[0]],
+                          outputs=[node.output[1]], dtypes=dtypes[1], shapes=shapes[1])
+        elif "pad_to_max_output_size" in node.attr:  # V4
```
jignparm: What happens if op=NMS_V5 and "pad_to_max_output_size"==True? We will not do this step?

RandySheriffH: v5 is triggered via non_max_suppression_with_scores, within which the pad_to_max_output_size always set to False.

PR #864, opened 2020-03-31
cherry pick fix to set IR based on opset

guschmue:
this prevents that the converter will write onnx files with IR7 when onnx-1.7 comes out. Same change is in master.

### DIFF
--- a/tf2onnx/constants.py
+++ b/tf2onnx/constants.py
@@ -40,5 +40,5 @@ ENV_TF2ONNX_DEBUG_MODE = "TF2ONNX_DEBUG_MODE"
 
 # Mapping opset to IR version.
 OPSET_TO_IR_VERSION = {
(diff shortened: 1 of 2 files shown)


guschmue:
ort is unhappy with PlaceholderWithDefault on IR3 which overrides the initializer with an given input.
Same works on master for some reason.
Debugging ...

guschmue:
Ah, in master with the new func() interface we can't do the PlaceholderWithDefault tests and we had disabled them. I think I'll set opset 7 to IR4.

jignparm:
>Ah, in master with the new func() interface we can't do the PlaceholderWithDefault tests and we had disabled them. I think I'll set opset 7 to IR4.

That's a good finding. Surprisingly, there were little/no complaints earlier  without this mapping....probably the models were not too sensitive to IR version earlier.

guschmue:
we require onnx-1.4 or better so we never did use IR3

PR #869, opened 2020-04-02
Rashuai/dynamic random

RandySheriffH:
Last fix for office intent search model - accept dynamic input shape.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1370,18 +1370,13 @@ class BackendTests(Tf2OnnxBackendTestBase):
             return tf.identity(x_, name=_TFOUTPUT)
         # since results are random, compare the shapes only
         self._run_test_case(func, [_OUTPUT], {}, check_value=False, check_shape=True)
-
+    '''
     @skip_caffe2_backend()
     def test_randomuniform_dyn_shape(self):
         # test for dynamic shape coming from a shape op
-        x_val = np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]], dtype=np.float32)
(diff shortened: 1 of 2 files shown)


tests/test_backend.py
```diff
@@ -1374,14 +1374,9 @@ def func():
     @skip_caffe2_backend()
     def test_randomuniform_dyn_shape(self):
         # test for dynamic shape coming from a shape op
-        x_val = np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]], dtype=np.float32)
+        x_val = np.array([0,1,2,3,5], dtype=np.int64)
         def func(x):
-            x_ = tf.stack([x, x])
-            x_ = tf.identity(x_)
-            x_ = tf.shape(x_, name="shape")
-            x_ = random_uniform(x_, name="rand", dtype=tf.float32)
```
RandySheriffH, line 1382: All these steps count for nothing - input still is constant.

tf2onnx/onnx_opset/generator.py
```diff
@@ -44,6 +44,18 @@ def version_1(cls, ctx, node, **kwargs):
             node.set_attr("shape", shape)
             ctx.set_shape(node.output[0], shape)
 
+    @classmethod
+    def version_9(cls, ctx, node, **kwargs):
+        if node.inputs[0].is_const():
+            cls.version_1(ctx, node, **kwargs)
+        else:
```
jignparm, line 51: why not move "else:" into version 1 above? That way all versions are fixed (not just 9)?

RandySheriffH: Because "ConstantOfShape" comes since version 9. :-)

PR #870, opened 2020-04-03
Fix scatternd - inputs bound to different type

jignparm:
Fix ScatterND -- the input type bounds are different between TF and Onnx, hence require casting.

### DIFF
--- a/tf2onnx/onnx_opset/tensor.py
+++ b/tf2onnx/onnx_opset/tensor.py
@@ -13,7 +13,6 @@ import logging
 import sys
 
 import numpy as np
-from onnx import numpy_helper
 from onnx import onnx_pb
 from onnx.onnx_pb import TensorProto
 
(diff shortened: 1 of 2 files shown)


tf2onnx/onnx_opset/tensor.py
```diff
@@ -517,22 +516,12 @@ def version_11(cls, ctx, node, **kwargs):
...
-        if ctx.get_dtype(input0) != TensorProto.INT64:
-            ctx.insert_new_node_on_input(const_of_shape, "Cast", input0, to=TensorProto.INT64)
-
-        # cast edge to INT64 if not already
-        input0 = node.input[0]
-        if ctx.get_dtype(input0) != TensorProto.INT64:
-            ctx.insert_new_node_on_input(node, "Cast", input0, to=TensorProto.INT64)
-
+        onnxdtype = ctx.get_dtype(node.input[1])
+        dtype = utils.map_onnx_to_numpy_type(onnxdtype)
```
RandySheriffH, line 520: So "dtype" not referred any more?

tf2onnx/onnx_opset/tensor.py
```diff
@@ -517,22 +516,12 @@ def version_11(cls, ctx, node, **kwargs):
...
+        const_of_shape = ctx.insert_new_node_on_input(node, "ConstantOfShape", node.input[2])
+        ctx.insert_new_node_on_input(const_of_shape, "Cast", const_of_shape.input[0], to=TensorProto.INT64)
+        ctx.insert_new_node_on_input(node, "Cast", node.input[0], to=TensorProto.INT64)
```
RandySheriffH, line 523: Are we binding input 0 and 2 to INT64 statically?
Suppose in the begging we have input 0 as float while 2 as INT64, and downstream of ScatterND is expecting float, what shall we do?

tf2onnx/onnx_opset/tensor.py
jignparm: sorry -- removed dtype

tf2onnx/onnx_opset/tensor.py
jignparm: The the end of the last line of this functions, the input types into the scatternd op should be as follows:

input_0 = onnxdtype
input_1 = int64
input_2 = onnxdtype

RandySheriffH: So we are setting input 0 to be of type of input 2, right? Why not reserve?

jignparm: In TF-ScatterND, input_2 is the 'data', but in ONNX-ScatterND, input_0 is the 'data'. 

Therefore, we have to get the type of the 'data' (T) from original input_2, and make sure that final input_0 and input_2 are equal to T.

The Onnx op doesn't exactly follow the TF spec -- so there's a bit of a criss-cross of type constraints. If you look slightly deeply it looks strange -- but if you look very deeply, it starts to look OK.

PR #877, opened 2020-04-08
Add getting started section to README

natke:
And cleaned up some markdown formatting issues.

- [x] Can we remove the install from source section (as that is covered by the install from GitHub)?
        _Clarified purpose with heading text instead_
- [x] Can we remove references to "trying to freeze the graph", except for in the how does it work? section?
- [x] The default opset is 8 in master. When do we expect to update the released pypi package?
        _Updated to 8_

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,10 @@ To create a distribution:
 
 ## Getting started
 
-To get started with `tensorflow-onnx`, run the `t2onnx.convert` command, providing the name of your TensorFlow model directory (where the model is in `saved model` format), and a name for the ONNX output file:
+To get started with `tensorflow-onnx`, run the `t2onnx.convert` command, providing:
+
+* the name of your TensorFlow model directory (where the model is in `saved model` format)
+* a name for the ONNX output file:
 
 ```python -m tf2onnx.convert --saved-model tensorflow-model-directory --output model.onnx```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,27 +143,27 @@ python -m tf2onnx.convert
...
 ```
 
-## Parameters
-### --input or --graphdef
+### Parameters
+#### --saved-model
+TensorFlow model as saved_model. We expect the path to the saved_model directory. tf2onnx will try to freeze the graph.
+#### --input or --graphdef
 TensorFlow model as graphdef file. If not already frozen we'll try to freeze the model.
 More information about freezing can be found here: [freeze graph tool](#freeze_graph).
```
jignparm, line 151: Thanks for looking into this -- re-reading it, it seems like some clarifications are needed!

Is it possible to convey to the user that they should only use only 1 of the following 4 parameters. Each one specifies the location of the input model file, but also tells tf2onnx what format to use to parse the model file  -- so if they specify multiple of these, it's just weird (@guschmue , should we change code to thrown an exception? Not sure if there's some history behind this however).

 Also, the ordering below is from "most common" to "least common", so it'd be good to have the ordering preserved.

   * --input 
   * --graphdef  (same description as --input, but used for forward compatibility)
   * --checkpoint
   * --saved-model

guschmue: We have a check to make sure there is at least 1 format but maybe we should check for exactly 1
https://github.com/onnx/tensorflow-onnx/blob/master/tf2onnx/convert.py#L79

README.md
```diff
@@ -109,27 +143,27 @@ python -m tf2onnx.convert
...
+#### --inputs-as-nchw
 By default we preserve the image format of inputs (nchw or nhwc) as given in the TensorFlow model. If your hosts (for example windows) native format nchw and the model is written for nhwc, ```--inputs-as-nchw``` tensorflow-onnx will transpose the input. Doing so is convinient for the application and the converter in many cases can optimize the transpose away. For example ```--inputs input0:0,input1:0 --inputs-as-nchw input0:0``` assumes that images are passed into ```input0:0``` as nchw while the TensorFlow model given uses nhwc.
-### --opset
+#### --opset
 By default we use the opset 7 to generate the graph. By specifying ```--opset``` the user can override the default to generate a graph with the desired opset. For example ```--opset 5``` would create a onnx graph that uses only ops available in opset 5. Because older opsets have in most cases fewer ops, some models might not convert on a older opset.
```
guschmue, line 161: can we make this opset 8 since that is the default in master?

README.md
```diff
@@ -109,27 +143,27 @@ python -m tf2onnx.convert
...
-### --target 
+#### --target 
 Some models require special handling to run on some runtimes. In particular, the model may use unsupported data types. Workarounds are activated with ```--target TARGET```. Currently supported values are listed on this [wiki](https://github.com/onnx/tensorflow-onnx/wiki/target). If your model will be run on Windows ML, you should specify the appropriate target value.
-### --custom-ops
+#### --custom-ops
 the runtime may support custom ops that are not defined in onnx. A user can asked the converter to map to custom ops by listing them with the --custom-ops option. Tensorflow ops listed here will be mapped to a custom op with the same name as the tensorflow op but in the onnx domain ai.onnx.converters.tensorflow. For example: ```--custom-ops Print``` will insert a op ```Print``` in the onnx domain ```ai.onnx.converters.tensorflow``` into the graph. We also support a python api for custom ops documented later in this readme.
```
guschmue, line 165: s/the/The/

jignparm:
>Can we remove the install from source section (as that is covered by the install from GitHub)?

Another option is to rename the sections "Install from source (for users)" and "Install from source (for devs)?"  The second way gives users a way make code changes and re-install (more granular)

>Can we remove references to "trying to freeze the graph", except for in the how does it work? section?

Sounds Ok to me, unless @guschmue has an alternate view :)

guschmue:
yes, sure.

README.md
```diff
@@ -62,34 +69,61 @@ For pytorch/caffe2, follow the instructions here:
...
-# Usage
+```python setup.py bdist_wheel```
+
+## Getting started
+
+To get started with `tensorflow-onnx`, run the `t2onnx.convert` command, providing the name of your TensorFlow model directory (where the model is in `saved model` format), and a name for the ONNX output file:
+
+```python -m tf2onnx.convert --saved-model tensorflow-model-directory --output model.onnx```
+
+The above command uses a default of `7` for the ONNX opset. If you need a newer opset, or want to limit your model to use an older opset then you can provide the `--opset` argument to the command. If you are unsure about which opset to use, refer to the [ONNX operator documentation](https://github.com/onnx/onnx/releases).
```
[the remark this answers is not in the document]
natke: And 7 in the deployed package? (How often do we deploy?)

README.md
```diff
@@ -62,34 +69,61 @@ For pytorch/caffe2, follow the instructions here:
...
 
-# Usage
+```python setup.py bdist_wheel```
+
+## Getting started
+
+To get started with `tensorflow-onnx`, run the `t2onnx.convert` command, providing the name of your TensorFlow model directory (where the model is in `saved model` format), and a name for the ONNX output file:
+
+```python -m tf2onnx.convert --saved-model tensorflow-model-directory --output model.onnx```
+
```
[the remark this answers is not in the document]
natke: Sure. (For markdown files, you also wrap the text in vscode)

natke:
> Can we remove the install from source section (as that is covered by the install from GitHub)?

> Another option is to rename the sections "Install from source (for users)" and "Install from source (for devs)?" The second way gives users a way make code changes and re-install (more granular)

That makes sense. We could also move it into CONTRIBUTING.md if we want to reduce clutter in this README.

README.md
natke: Fixed

README.md
```diff
@@ -62,34 +69,61 @@ For pytorch/caffe2, follow the instructions here:
...
-names with ```--inputs INPUTS``` and ```--outputs OUTPUTS```.
+For `graphdef` format:
+
+```python -m tf2onnx.convert --graphdef  tensorflow-model-graphdef-file --output model.onnx --inputs input0:0,input1:0 --outputs output0:0```
+
+For `checkpoint` format:
+
+```python -m tf2onnx.convert --checkpoint  tensorflow-model-meta-file --output model.onnx --inputs input0:0,input1:0 --outputs output0:0```
+
+If your model is not in `saved model` format and you do not know the input and output nodes of the model, you can use the [summarize_graph](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/graph_transforms) TensorFlow utility. The `summarize_graph` tool does need to be downloaded and built from source. If you have the option of going to your model provider and obtaining the model in `saved model` format, then we recommend doing so.
```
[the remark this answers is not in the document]
natke: I was going for something slightly different here. The sentence above was to make sure users know they don't have to provide the inputs and outputs if they have saved model format. This is for the scenario where they don't have it but we recommend they get it.

README.md
```diff
@@ -109,48 +146,55 @@ python -m tf2onnx.convert
...
+TensorFlow model as saved_model. We expect the path to the saved_model directory.
+
+#### --input or --graphdef
+
+TensorFlow model as graphdef file.
+
+#### --checkpoint
+
+TensorFlow model as checkpoint. We expect the path to the .meta file.
+
```
jignparm, line 162: Can we keep the ordering of the arguments the same as the command line above?  Makes it easier to connect the dots visually.

natke: Yes, sure, that makes sense. I'll also try to clarify that only one of the model format options should be provided

natke: Fixed

Issue #883, opened 2020-04-14
Your ONNX model has been generated with INT64 weights, while TensorRT does not natively support INT64. Attempting to cast down to INT32


carlosgalvezp:
Hi,

I'm trying to use the TensorFlow-ONNX-TensorRT workflow described here:

https://devblogs.nvidia.com/speeding-up-deep-learning-inference-using-tensorflow-onnx-and-tensorrt/

Versions:
Tensorflow: 1.14.0 (GPU)
ONNX: 1.6.0
Tensorflow-onnx: master at the time of writing
TensorRT: 7.0

I could convert my TensorFlow model to ONNX successfully with opset 11. However when passing it through "trtexec" I get the following errors:

```
[V] [TRT] ModelImporter.cpp:90: Importing initializer: const_fold_opt__961
[W] [TRT] onnx2trt_utils.cpp:198: Your ONNX model has been generated with INT64 weights, while TensorRT does not natively support INT64. Attempting to cast down to INT32.
```

Why are the weights not FP32? In some cases it even fails to convert to INT32:

```
BatchToSpaceND_int_max__752__753
[W] [TRT] onnx2trt_utils.cpp:198: Your ONNX model has been generated with INT64 weights, while TensorRT does not natively support INT64. Attempting to cast down to INT32.
[V] [TRT] onnx2trt_utils.cpp:212: Weight at index 0: 9223372036854775807 is out of range. Clamping to: 2147483647
```

It seems the errors come from layers inserted by the ONNX converter. 

Thanks!

jignparm:
>Your ONNX model has been generated with INT64 weights

In Onnx, some operators require `int_max` or `int_min` as special values to denote 'infinity'(e.g. Slice operator), which is probably where the large integer values are coming from.

TRT seems to be doing the right thing above by clamping and/or casting to INT32 however. Are these treated as critical errors, or just warnings by TRT?

carlosgalvezp:
Hi!

It's indeed "just" a warning, but it's not a good idea to keep warnings, they produce noise and distract the user from real warnings and errors that should be taken care off. Why does ONNX need 64-bit integers for these parameters? Infinity can very well be represented with 32-bit integers? I think fixing this warning would greatly improve the user experience :)

Anyway the actual crash I got was here:

```
[V] [TRT] ModelImporter.cpp:107: Parsing node: Main_conv_dilated/SpaceToBatchND_pad__833 [Pad]
[...]

PR #888, opened 2020-04-20
move some ops to generators.py, new version of supported ops doc

guschmue:


guschmue:
push a new version of support_status.md mentioned in https://github.com/onnx/tensorflow-onnx/issues/883.

Moved IteratorV2, IteratorGetNextops from controlflow.py to generators.py.
Also the registration for the ops above was not correct, fixed that as well.

PR #890, opened 2020-04-21
Support MatrixDiagPart v2 and v3

RandySheriffH:
Support MatrixDiagPart v2 and v3 by extracting diagonals with proper pads.


tests/test_backend.py
```diff
@@ -3108,6 +3108,63 @@ def func(x, y):
...
+
+        def func_v2(X, K):
+            return tf.raw_ops.MatrixDiagPartV2(input=X, k=K, padding_value=0.123, name=_TFOUTPUT)
+
+        def func_v3(X, K):
+            return tf.raw_ops.MatrixDiagPartV3(input=X, k=K, padding_value=0.123, align='RIGHT_LEFT', name=_TFOUTPUT)
+
+        k_val = np.array([0]).astype(np.int32)
+        self._run_test_case(func_v2, [_OUTPUT], {_INPUT: x_val, _INPUT1: k_val})
+        self._run_test_case(func_v3, [_OUTPUT], {_INPUT: x_val, _INPUT1: k_val})
```
jignparm, line 3124: Maybe  a function like run_test_v2_v3(...) can reduce the number of lines here?

RandySheriffH: Good point.

RandySheriffH: lines trimmed by latest commit.

tests/test_backend.py
```diff
@@ -3108,6 +3108,51 @@ def func(x, y):
...
+    @check_tf_min_version("2.2")
+    def test_matrix_diag_part_v3(self):
+
+        def func(X, K):
+            v2 = tf.raw_ops.MatrixDiagPartV2(input=X, k=K, padding_value=0.123, name=_TFOUTPUT)
+            v3 = tf.raw_ops.MatrixDiagPartV3(input=X, k=K, padding_value=0.123, align='RIGHT_LEFT', name=_TFOUTPUT1)
+            return v2, v3
+
+        x_val = np.random.random([4, 5]).astype(np.float32)
+        k_val = np.array([0]).astype(np.int32)
```
jignparm, line 3121: It looks like all the k_vals and x_vals can be put a loop as well -- would reduce test quite a bit.

```python
for x in [[4,5], [2, 3, 4, 5]]:
    x_val = np(x, ...)
    for k in [0, 1, -1,...]:
        k_val = np([k],...)
        self._run_test(...)
```

RandySheriffH: Sure why not.

PR #897, opened 2020-04-24
opset 12 support

RandySheriffH:
Adding support for opset 12 ops.

### DIFF
--- a/tf2onnx/custom_opsets/ms.py
+++ b/tf2onnx/custom_opsets/ms.py
@@ -103,3 +103,11 @@ class CropAndResize:
         ctx.insert_new_node_on_input(node, "Transpose", node.input[0], perm=constants.NHWC_TO_NCHW)
         ctx.insert_new_node_on_output("Transpose", node.output[0], node.name + '_transposed',
                                       None, perm=constants.NCHW_TO_NHWC)
+
+@tf_op("MatrixInverse", onnx_op="Inverse")
+class Inverse:
+    @classmethod
+    def version_12(cls, ctx, node, **kwargs):
+        utils.make_sure(node.get_attr('adjoint').i == 0, "adjoint must be false")
+        del node.attr["adjoint"]
+        node.domain = constants.MICROSOFT_DOMAIN
--- a/tf2onnx/custom_opsets/ms.py
+++ b/tf2onnx/custom_opsets/ms.py
@@ -104,7 +104,7 @@ class CropAndResize:
         ctx.insert_new_node_on_output("Transpose", node.output[0], node.name + '_transposed',
                                       None, perm=constants.NCHW_TO_NHWC)
 
-@tf_op("MatrixInverse", onnx_op="Inverse")
+@tf_op("MatrixInverse", domain=constants.MICROSOFT_DOMAIN, onnx_op="Inverse")
 class Inverse:
     @classmethod
     def version_12(cls, ctx, node, **kwargs):
--- a/tf2onnx/onnx_opset/logical.py
+++ b/tf2onnx/onnx_opset/logical.py
@@ -110,28 +110,17 @@ class GreaterLess:
         target_dtype = TensorProto.FLOAT
         _add_cast_to_inputs(ctx, node, supported_dtypes, target_dtype)
 
-
-@tf_op("GreaterEqual", onnx_op="Less")
-@tf_op("LessEqual", onnx_op="Greater")
+@tf_op(["GreaterEqual", "LessEqual"])
 class GreaterLessEqual:
     @classmethod
     def version_7(cls, ctx, node, **kwargs):
         GreaterLess.version_7(ctx, node, **kwargs)
         output_name = node.output[0]
+        node.op.op_type = "Less" if node.op.op_type == "GreaterEqual" else "Greater"
         new_node = ctx.insert_new_node_on_output("Not", output_name, name=utils.make_name(node.name))
         ctx.copy_shape(output_name, new_node.output[0])
         ctx.set_dtype(new_node.output[0], ctx.get_dtype(output_name))
 
-
-@tf_op("GreaterEqual", onnx_op="GreaterOrEqual")
-class GreaterEqual:
     @classmethod
     def version_12(cls, ctx, node, **kwargs):
-        pass
-
-
-@tf_op("LessEqual", onnx_op="LessOrEqual")
-class LessEqual:
-    @classmethod
-    def version_12(cls, ctx, node, **kwargs):
-        pass
+        node.op.op_type = "GreaterOrEqual" if node.op.op_type == "GreaterEqual" else "LessOrEqual"
--- a/tf2onnx/onnx_opset/logical.py
+++ b/tf2onnx/onnx_opset/logical.py
@@ -121,3 +121,17 @@ class GreaterLessEqual:
         new_node = ctx.insert_new_node_on_output("Not", output_name, name=utils.make_name(node.name))
         ctx.copy_shape(output_name, new_node.output[0])
         ctx.set_dtype(new_node.output[0], ctx.get_dtype(output_name))
+
+
+@tf_op("GreaterEqual", onnx_op="GreaterOrEqual")
+class GreaterEqual:
+    @classmethod
+    def version_12(cls, ctx, node, **kwargs):
+        pass
+
+
+@tf_op("LessEqual", onnx_op="LessOrEqual")
+class LessEqual:
+    @classmethod
+    def version_12(cls, ctx, node, **kwargs):')
+        pass
--- a/tf2onnx/onnx_opset/logical.py
+++ b/tf2onnx/onnx_opset/logical.py
@@ -133,5 +133,5 @@ class GreaterEqual:
 @tf_op("LessEqual", onnx_op="LessOrEqual")
 class LessEqual:
     @classmethod
-    def version_12(cls, ctx, node, **kwargs):')
+    def version_12(cls, ctx, node, **kwargs):
         pass
--- a/tf2onnx/onnx_opset/math.py
+++ b/tf2onnx/onnx_opset/math.py
@@ -567,8 +567,3 @@ class Einsum:
     @classmethod
     def version_12(cls, ctx, node, **kwargs):
         del node.attr["N"]
-
-
-
-
-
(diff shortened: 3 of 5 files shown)


tf2onnx/onnx_opset/math.py
```diff
@@ -544,3 +544,26 @@ def version_11(cls, ctx, node, **kwargs):
             cast_back_node.set_attr("to", dtypes[0])
             ctx.set_dtype(cast_back_node.output[0], dtypes[0])
             ctx.copy_shape(node.name, cast_back_node.output[0])
+
+@tf_op("MatrixInverse")
+class Inverse:
+
+    @classmethod
```
guschmue, line 551: this should be in tensor.py ?

tf2onnx/onnx_opset/math.py
```diff
@@ -22,7 +22,7 @@
 
 # pylint: disable=unused-argument,missing-docstring
 
-@tf_op(["Add", "AddV2", "Div", "Mul", "Sub"])
+@tf_op(["Add", "AddV2", "Div", "Mul", "Sub", "LessOrEqual"])
```
guschmue, line 25: LessOrEqual belongs into logical.py. Also - we do support his already so you need to be careful how you do this.

tf2onnx/onnx_opset/math.py
```diff
@@ -544,3 +544,26 @@ def version_11(cls, ctx, node, **kwargs):
...
+                      shapes=shapes, dtypes=dtypes)
+
+@tf_op("SquaredDistance")
+class SquaredDistance:
+
+    @classmethod
+    def version_12(cls, ctx, node, **kwargs):
+        shapes = node.output_shapes
+        dtypes = node.output_dtypes
+        ctx.remove_node(node.name)
```
guschmue, line 567: why not just set node,type = "MeanSquaredDistance"

tf2onnx/onnx_opset/math.py
```diff
@@ -544,3 +544,26 @@ def version_11(cls, ctx, node, **kwargs):
...
+@tf_op("MatrixInverse")
+class Inverse:
+
+    @classmethod
+    def version_12(cls, ctx, node, **kwargs):
+        utils.make_sure(node.get_attr('adjoint').i == 0, "adjoint must be false")
+        shapes = node.output_shapes
+        dtypes = node.output_dtypes
+        ctx.remove_node(node.name)
+        ctx.make_node("Inverse", inputs=node.input, outputs=node.output, name=node.name,
```
guschmue, line 557: why not just set node,type = "Inverse"?

tf2onnx/onnx_opset/math.py
```diff
@@ -544,3 +544,26 @@ def version_11(cls, ctx, node, **kwargs):
...
+                      shapes=shapes, dtypes=dtypes)
```
jignparm, line 558: A good trick I found is to use the onnx_op= parameter -- I think the snippet below should work the same way, and no need to remove + add nodes.

```python
@tf_op("MatrixInverse", onnx_op="Inverse")
class Inverse:
    @classmethod
    def version_12(cls, ctx, node, **kwargs):
        utils.make_sure(node.get_attr('adjoint').i == 0, "adjoint must be false")
```

tf2onnx/onnx_opset/math.py
```diff
@@ -544,3 +544,26 @@ def version_11(cls, ctx, node, **kwargs):
...
+@tf_op("SquaredDistance")
+class SquaredDistance:
+
+    @classmethod
+    def version_12(cls, ctx, node, **kwargs):
+        shapes = node.output_shapes
+        dtypes = node.output_dtypes
+        ctx.remove_node(node.name)
+        ctx.make_node("MeanSquaredDistance", inputs=node.input, outputs=node.output, name=node.name,
+                      shapes=shapes, dtypes=dtypes, attr={"reduction": "none"})
```
jignparm, line 569: Same thing here I believe -- onnx_op="MeanSquaredDistance"

tf2onnx/onnx_opset/math.py
guschmue: onnx has it in math and while I think this is wrong we should do the same

tf2onnx/onnx_opset/math.py
```diff
@@ -555,7 +555,7 @@ def version_12(cls, ctx, node, **kwargs):
         dtypes = node.output_dtypes
         ctx.remove_node(node.name)
         ctx.make_node("Inverse", inputs=node.input, outputs=node.output, name=node.name,
-                      shapes=shapes, dtypes=dtypes)
+                      domain=constants.MICROSOFT_DOMAIN, shapes=shapes, dtypes=dtypes)
```
guschmue: Inverse is in the onnx domain, no?
https://github.com/onnx/onnx/blob/master/docs/Operators.md#Inverse

tf2onnx/onnx_opset/math.py
RandySheriffH: Yes, apologies.
But for the one in logical.py, shall we map it to the latest LessOrEqual or just leave it as it is.

tf2onnx/onnx_opset/math.py
RandySheriffH: Done.

tf2onnx/onnx_opset/math.py
RandySheriffH: Done.

tf2onnx/onnx_opset/math.py
RandySheriffH: Good point, thanks!

tf2onnx/onnx_opset/math.py
RandySheriffH: Done.

tf2onnx/onnx_opset/math.py
RandySheriffH: No, it is a contrib op.
https://github.com/microsoft/onnxruntime/pull/3485

tf2onnx/onnx_opset/math.py
RandySheriffH: Added opset 12 GreaterOrEqual and LessOrEqual to simplify the logic.

tf2onnx/onnx_opset/math.py
guschmue: the onnx spec says Inverse is in onnx namespace - we follow that spec. It is correct to put it here regardless if onnxruntime supports it or not since they spec says so.

If onnxruntime implements this in ms domain and you want to support it there, you need to add a  handler here: tensorflow-onnx/tf2onnx/custom_opsets/ms.py.

RandySheriffH: moved to ms.py.

tf2onnx/custom_opsets/ms.py
```diff
@@ -103,3 +103,11 @@ def version_11(cls, ctx, node, **kwargs):
         ctx.insert_new_node_on_input(node, "Transpose", node.input[0], perm=constants.NHWC_TO_NCHW)
         ctx.insert_new_node_on_output("Transpose", node.output[0], node.name + '_transposed',
                                       None, perm=constants.NCHW_TO_NHWC)
+
+@tf_op("MatrixInverse", onnx_op="Inverse")
```
guschmue, line 107: domain= is missing.

tf2onnx/onnx_opset/logical.py
```diff
@@ -121,3 +121,17 @@ def version_7(cls, ctx, node, **kwargs):
         new_node = ctx.insert_new_node_on_output("Not", output_name, name=utils.make_name(node.name))
         ctx.copy_shape(output_name, new_node.output[0])
         ctx.set_dtype(new_node.output[0], ctx.get_dtype(output_name))
+
+
+@tf_op("GreaterEqual", onnx_op="GreaterOrEqual")
+class GreaterEqual:
```
guschmue, line 127: so how this this work for opset < 12 ?
If the registration at the old place works it is accidental ... you should move the pre opset 11 code to here was well.

tf2onnx/onnx_opset/logical.py
```diff
@@ -121,3 +121,17 @@ def version_7(cls, ctx, node, **kwargs):
...
+    @classmethod
+    def version_12(cls, ctx, node, **kwargs):
+        pass
+
+
+@tf_op("LessEqual", onnx_op="LessOrEqual")
+class LessEqual:
```
guschmue, line 134: same as above.

tf2onnx/custom_opsets/ms.py
guschmue: And since you just moved it, it doesn't satisfy the onnx spec so it does not work for trt as example. To implement the onnx spec you need to keep it in math.py as well.

RandySheriffH: Done.

tf2onnx/onnx_opset/logical.py
RandySheriffH: Added case covering older version.

tf2onnx/onnx_opset/logical.py
RandySheriffH: Added case covering older version.

tf2onnx/onnx_opset/logical.py
RandySheriffH: dup removed.

tf2onnx/onnx_opset/math.py
```diff
@@ -544,3 +544,26 @@ def version_11(cls, ctx, node, **kwargs):
...
+    @classmethod
+    def version_12(cls, ctx, node, **kwargs):
+        del node.attr["N"]
+
+
+@tf_op("MatrixInverse", onnx_op="Inverse")
+class Inverse:
+    @classmethod
+    def version_12(cls, ctx, node, **kwargs):
+        utils.make_sure(node.get_attr('adjoint').i == 0, "adjoint must be false")
```
jignparm, line 567: Why "adjoint must be false"? Does this parameter simply mean we should compute the inverse of the adjoint of this matrix if set to true? It seems we should be able to handle "true" as well.

RandySheriffH: Tested with inv(A) = adj(A)/det(A) approach, witnessed big diff. Will address adjoint case later.

Issue #861, opened 2020-03-26
Flatten equations for rnn patterns


buddhapuneeth:
In rnn_utils.py file, we can see all the regex of the LSTM and GRU nodes. It will be helpful if you add the actual mathematical equation also there as everyone follow different notations, so it will be more readable.

buddhapuneeth:
Also we have operator type 'Enter' in the pattern, explanation of that operator is not available anywhere.

jignparm:
These op patterns are what are observed in TF graphs, with  'Enter' and 'Exit' denoting flow-control operators. Most of the patterns in rnn_utils.py should read like a textbook (e.g. see below for a sequence of ConcatV2, MatMul, BiasAdd etc.). As for mathematical equations, that could be a nice touch, although that could apply to almost every other (regular) operator as well.

Is there some model that's not working/converting? 
```python
# TensorFlow LSTMCell/BasicLSTMCell computation graph matching
xc_pattern = OpTypePattern('Split', inputs=[
    OpTypePattern("Const"), # axis for split
    OpTypePattern("BiasAdd", name="bias_add", inputs=[
        OpTypePattern("MatMul", inputs=[
            OpTypePattern("ConcatV2|Concat", name="xh"),
            OpTypePattern("Enter", inputs=[
                OpTypePattern("*", name="cell_kernel"),
            ]),
        ]),
        OpTypePattern("Enter", inputs=[
            OpTypePattern("*", name="cell_bias"),
        ]),
    ]),
])

```

buddhapuneeth:
Thanks for explaining on Enter and Exit operators.
Yes, I am facing issue while converting LSTM models. The sub-graph pattern matching is failing. 
It was working after making small change as below (by skipping Enter):
  ```python
xc_pattern = OpTypePattern('Split', inputs=[
    OpTypePattern("Const"), # axis for split
    OpTypePattern("BiasAdd", name="bias_add", inputs=[
        OpTypePattern("MatMul", inputs=[
            OpTypePattern("ConcatV2|Concat", name="xh"),
            OpTypePattern("*", name="cell_kernel"),
        ]),
        OpTypePattern("*", name="cell_bias"),
    ]),
])
```

This is working but need to understand why Enter operator is not coming.

guschmue:
[...]

PR #913, opened 2020-05-07
Fix LSTM pattern matching for version between 1.15.0 and 2.x.

RandySheriffH:
A tentative fix with trivial change to match LSTM for version between 1.15.0 and 2.x.
Addressing issue https://github.com/onnx/tensorflow-onnx/issues/861

### DIFF
--- a/tf2onnx/rewriter/rnn_utils.py
+++ b/tf2onnx/rewriter/rnn_utils.py
@@ -51,10 +51,10 @@ lstmcell_pattern = \
     OpTypePattern('Mul', name='ht', inputs=[
         OpTypePattern("Sigmoid", name="ot", inputs=[xc_pattern]),
         OpTypePattern('Tanh', inputs=[
-            OpTypePattern("Add", name="ct", inputs=[
+            OpTypePattern("Add|AddV2", name="ct", inputs=[
                 OpTypePattern("Mul", name="ct_identity_consumer", inputs=[
                     OpTypePattern("Sigmoid", name="ft", inputs=[
-                        OpTypePattern("Add", inputs=[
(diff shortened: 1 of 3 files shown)


tf2onnx/rewriter/rnn_utils.py
```diff
@@ -68,6 +69,39 @@ class REWRITER_RESULT(Enum):
...
+                    OpTypePattern("*"),
+                ]),
+                OpTypePattern("Mul", inputs=[
+                    OpTypePattern("Sigmoid", name="it", inputs=[xc_pattern_optimized]),
+                    OpTypePattern("Tanh", name="gt", inputs=[xc_pattern_optimized]),
+                ]),
+            ]),
+        ]),
+    ])
+
```
jignparm, line 104: Any UT that can be re/activated to check if this new pattern is working as expected? -- or are the current tests already sufficient?

RandySheriffH: Current cases in test_lstm.py already covering it.
I have enabled check_lstm_count(...) in common.py to make sure all cases are passing.

tf2onnx/rewriter/rnn_utils.py
```diff
@@ -68,6 +69,39 @@ class REWRITER_RESULT(Enum):
         ]),
     ])
 
+xc_pattern_optimized = \
```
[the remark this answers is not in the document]
RandySheriffH: @guschmue :
My understanding is for tf >= 1.15.0 when bias is all zero, "BiasAdd" will be optimized to "Identity" by tf optimizer, hence the name.

PR #919, opened 2020-05-12
Support for QuantizeAndDequantize operation

peri044:
This PR adds support for tf.quantization.quantize_and_dequantize operations. TF op is split into QuantizeLinear and DequantizeLinear ops which perform the same operation. Added test cases for both unsigned and signed inputs.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1917,7 +1917,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
 
         self._run_test_case(func_fusedbn, [_OUTPUT], {_INPUT: x_val}, rtol=1e-05, graph_validator=graph_validator)
 
-    @check_tf_min_version("1.15", "not supported in tf-2.0")
+    @check_tf_min_version("1.15", "skip_tf2")
     @check_opset_min_version(10, "quantize_and_dequantize")
     def test_qdq_unsigned_input(self):
         x_shape = [3, 3, 2]
@@ -1927,7 +1927,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
             return tf.identity(x_, name=_TFOUTPUT)
         _ = self._run_test_case(func, [_OUTPUT], {_INPUT: x_val})
 
-    @check_tf_min_version("1.15", "not supported in tf-2.0")
+    @check_tf_min_version("1.15", "skip_tf2")
     @check_opset_min_version(10, "quantize_and_dequantize")
     def test_qdq_signed_input(self):
         x_shape = [3, 3, 2]
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1917,7 +1917,8 @@ class BackendTests(Tf2OnnxBackendTestBase):
 
         self._run_test_case(func_fusedbn, [_OUTPUT], {_INPUT: x_val}, rtol=1e-05, graph_validator=graph_validator)
 
-    @check_tf_min_version("1.15", "skip_tf2")
+    @check_tf_min_version("1.15")
+    @skip_tf2()
     @check_opset_min_version(10, "quantize_and_dequantize")
     def test_qdq_unsigned_input(self):
         x_shape = [3, 3, 2]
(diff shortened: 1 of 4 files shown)


guschmue:
can you sign the cla ?

peri044:
@guschmue I did try it multiple times. It always shows the status as pending. Is there anything that I need to do additionally? Thanks

guschmue:
The individual commits are from user dheerajperi - maybe that is  local user id ?

peri044:
@guschmue  Yeah that's the issue. Had two user accounts. Signed it now. Thanks !!

tf2onnx/rewriter/quantization_ops_rewriter.py
```diff
@@ -0,0 +1,106 @@
...
+        # Calculate scales from the min and max values
+        scale_from_min_side = min_quantized/min_value if min_quantized*min_value > 0 else max_quantized
+        scale_from_max_side = max_quantized/max_value if max_quantized*max_value > 0 else max_quantized
+
+        if scale_from_min_side < scale_from_max_side:
+            scale = scale_from_min_side
+        else:
+            scale = scale_from_max_side
+
+        assert scale > 0
```
guschmue, line 48: can you change the assert to utils.make_sure ?

tests/test_backend.py
```diff
@@ -77,6 +77,7 @@
     random_uniform = tf.compat.v1.random_uniform
     fused_batch_norm = tf.compat.v1.nn.fused_batch_norm
     dropout = tf.compat.v1.nn.dropout
+    quantize_and_dequantize = tf.compat.v1.quantization.quantize_and_dequantize
```
guschmue, line 80: just wondering - why only tf-1.15? We should be able to make this work for other versions as well, right?

tf2onnx/rewriter/quantization_ops_rewriter.py
peri044: Done

tests/test_backend.py
peri044: Added support for TF2 as well now. The reason this doesn't work TF <=1.14 does not have narrow range argument to restrict the quantization range to [-127, 127]. TF <=1.14 has a default quantization range of [128, 128]. ONNX Quantize seems to be using [-127, 127] by default. Due to this slight change in the range and no argument in ONNX to control the range, the test case will fail for the extreme elements. For the testcase in the code, there are 3 mismatched elements for TF <=1.14. This is due to the range mismatch. 

>  Mismatched elements: 3 / 18 (16.7%)
Max absolute difference: 0.04724407
Max relative difference: 0.00787401
 x: array([[[-6.047244, -6.047244],
        [-6.047244, -6.      ],
        [-5.007874, -4.015748]],...
 y: array([[[-6.      , -6.      ],
        [-6.      , -6.      ],
        [-5.007874, -4.015748]],...

peri044: Added support for TF2 as well now. The reason this doesn't work TF <=1.14 does not have `narrow_range` argument to restrict the quantization range to [-127, 127]. TF <=1.14 has a default quantization range of [-128, 127]. ONNX Quantize seems to be using [-127, 127] by default. Due to this slight change in the range and no argument in ONNX to control the range, the test case will fail for the extreme elements. For the testcase in the code, there are 3 mismatched elements for TF <=1.14. This is due to the range mismatch. 

>  Mismatched elements: 3 / 18 (16.7%)
Max absolute difference: 0.04724407
Max relative difference: 0.00787401
 x: array([[[-6.047244, -6.047244],
        [-6.047244, -6.      ],
        [-5.007874, -4.015748]],...
 y: array([[[-6.      , -6.      ],
        [-6.      , -6.      ],
        [-5.007874, -4.015748]],...

guschmue: got it. Maybe we can come up with some solution for this later.

guschmue:
Awesome, thanks for the contribution!!

PR #923, opened 2020-05-13
Activate opset12 tests

jignparm:
Activate Opset12 testing


guschmue:
your runtime is too old to run opset 12 :(

guschmue:
@peri044  fyi

peri044:
@guschmue Thanks for letting me know. I did find it weird that the doc mentioned [-128, 127] but used [-127, 127] instead. Glad it is fixed now.

PR #925, opened 2020-05-16
Add stacked LSTM support

buddhapuneeth:
- added new base class exclusive for LSTM which supports stacked LSTMs
- this base class can be combined with unit_rnn_rewriter_base once we add support for stacked layers in other RNN types like GRU
- modified LSTM rewriter such that it allows multiple cell matches and for all the variable finder and weight-bias methods we pass the LSTM index
- added new test for the same

### DIFF
--- a/tf2onnx/rewriter/lstm_rewriter.py
+++ b/tf2onnx/rewriter/lstm_rewriter.py
@@ -16,7 +16,7 @@ from tf2onnx.graph_builder import GraphBuilder
 from tf2onnx.rewriter.rnn_utils import RNNUnitType, get_weights_from_const_node
 from tf2onnx.utils import is_tf_concat_op, is_tf_slice_op
 
-from tf2onnx.rewriter.unit_rnn_rewriter_base import UnitRnnRewriterBase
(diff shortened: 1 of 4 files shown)


guschmue:
looking

tf2onnx/rewriter/lstm_rewriter.py
```diff
@@ -16,36 +16,40 @@
...
     def run(self):
         logger.debug("enter lstm rewriter")
-        return super(LSTMUnitRewriter, self).run()
+        return super(LSTMRewriter, self).run()
 
     def find_cell(self, context):
         lstm_cell_types = [RNNUnitType.LSTMCell, RNNUnitType.LSTMBlockCell]
         for cell_type in lstm_cell_types:
             cell_match = self._match_cell(context, cell_type)
+            if cell_match and len(cell_match) >= 1:
```
guschmue, line 41: nit: no need for 'len(cell_match) >= 1'

guschmue:
@RandySheriffH - want to take a look at this PR?

guschmue:
@buddhapuneeth thanks for the contribution!!

PR #929, opened 2020-05-20
Multiple fixes for Bert Model (fine-tuned) 

jignparm:
Allow for dynamic output shapes in Onnx model [ i.e. rank=1, shape =(?,)] 
Fix 'Squeeze' op to handle shape of (?,)
Refactor 'Constant_folding'  rewriter to pass unit tests

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -65,14 +65,10 @@ def rewrite_constant_fold(g, ops):
             if set(op.output) & set(g.outputs): continue
             try:
                 inputs = []
-                skip = False
                 for node in op.inputs:
                     if not node.is_const():
-                        skip = True
                         break
                     inputs.append(node.get_tensor_value(as_list=False))
-
-                if skip: continue
 
                 logger.debug("op name %s, %s, %s", op.name, len(op.input), len(inputs))
                 if inputs and len(op.input) == len(inputs):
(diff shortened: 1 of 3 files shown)


tf2onnx/tfonnx.py
```diff
@@ -54,28 +54,26 @@ def rewrite_constant_fold(g, ops):
...
         keep_looking = False
         for idx, op in enumerate(ops):
             func = func_map.get(op.type)
-            if func is None:
-                continue
+            if func is None: continue
+            if set(op.output) & set(g.outputs): continue
             try:
                 inputs = []
+                skip = False
```
guschmue, line 68: need the skip? we already check that all inputs are const below

tf2onnx/tfonnx.py
```diff
@@ -466,7 +460,7 @@ def compat_handler(ctx, node, **kwargs):
                  rewrite_single_direction_lstm, rewrite_bi_direction_lstm,
                  rewrite_single_direction_gru, rewrite_bi_direction_gru,
                  rewrite_custom_rnn_cell, rewrite_generic_loop, rewrite_cond,
-                 rewrite_biasadd_with_conv2d,
+                 rewrite_biasadd_with_conv2d, rewrite_constant_fold
```
guschmue, line 463: might help to run const_cold first in the longer term but maybe more risk right now

tf2onnx/tfonnx.py
jignparm: Thanks for the catch!  `inputs and len(op.input) == len(inputs)` will only run if all are `const`. Removed the `skip`

tf2onnx/tfonnx.py
jignparm: Moved it to first rewriter.

Issue #914, opened 2020-05-07
ResizeBilinear coordinate_transformation_mode half_pixel


jeffpyke:
**Describe the bug**
ResizeBilinear coordinate transformation mode is not read from the TensorFlow graph. Therefore the ONNX model does not match TensorFlow when v2 ResizeBilinear op is used (half_pixel_centers=True).

https://github.com/onnx/tensorflow-onnx/blob/17e5c398f599d34b01896fad82032773874176b0/tf2onnx/onnx_opset/nn.py#L741

https://github.com/onnx/onnx/blob/master/docs/Operators.md#attributes-66

**Urgency**
medium

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04):
- Tensorflow Version: 1.15+
- Python version: 3

**To Reproduce**
Convert a graph with compat.v1 and compat.v2. Only one works (compat.v1).

**Expected behavior**
compat.v1.resize and compat.v2.resize are both supported.

**Screenshots**

**Additional context**
Add any other context about the problem here. If the issue is about a particular model, please share the model details as well to facilitate debugging.

jignparm:
Fixed via https://github.com/onnx/tensorflow-onnx/pull/932.

PR #932, opened 2020-05-24
Add half pixel transformation to resize bilinear op

peri044:
resize bilinear defaults to half_pixel_centers=True in TF 2.X. This PR sets the transformation mode of resize accordingly.

### DIFF
--- a/tf2onnx/onnx_opset/nn.py
+++ b/tf2onnx/onnx_opset/nn.py
@@ -736,9 +736,12 @@ class Resize:
             const_empty_float.output[0],
             concat_shape.output[0]
         ]
+        transformation_mode = "asymmetric"
+        if "half_pixel_centers" in node.attr and node.attr["half_pixel_centers"].i:
+            transformation_mode = "half_pixel"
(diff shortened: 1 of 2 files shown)


tf2onnx/onnx_opset/nn.py
```diff
@@ -736,9 +736,12 @@ def version_11(cls, ctx, node, **kwargs):
             const_empty_float.output[0],
             concat_shape.output[0]
         ]
+        transformation_mode = "asymmetric"
+        if "half_pixel_centers" in node.attr and node.attr["half_pixel_centers"].i:
+            transformation_mode = "half_pixel"
```
jignparm, line 741: Thanks for the contribution. Seems like this should address issue: https://github.com/onnx/tensorflow-onnx/issues/914 as well.

Is that assumption correct?

peri044: @jignparm That's right. This should fix the issue.

jignparm: Thanks for the contribution and the unit test! LGTM.

PR #936, opened 2020-05-26
add support for tf.math.is_finite

guschmue:
- add support for tf.math.is_finite
- expanddims() does not need to depend on shape in opset-11
- remove Inverse() from onnx_opset/math.py since this is in ms-domain

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -3191,7 +3191,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
                 self._run_test_case(func, [_OUTPUT, _OUTPUT1], {_INPUT: x_val, _INPUT1: k_val})
 
     @test_ms_domain()
-    def test_inverse(self):
+    def test_inverse(self, extra_opset):
         # this depends on onnx Inverse which was removed from opset-12 but does exists in the ms-domain
         x_val = np.random.random([5, 5]).astype(np.float32)
         def func(x):
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -3224,6 +3224,14 @@ class BackendTests(Tf2OnnxBackendTestBase):
                    tf.math.greater_equal(x, y, name=_TFOUTPUT1)
         self._run_test_case(func, [_OUTPUT, _OUTPUT1], {_INPUT: x_val, _INPUT1: y_val})
 
+    @check_opset_min_version(10)
+    def test_is_finite(self):
+        x_val = np.array([5.0, 4.8, 6.8, np.inf, np.nan], dtype=np.float32)
+        def func(x):
+            y = tf.math.is_finite(x)
+            return tf.identity(y, name=_TFOUTPUT)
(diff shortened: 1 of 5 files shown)


tests/test_backend.py
```diff
@@ -3224,6 +3224,14 @@ def func(x, y):
                    tf.math.greater_equal(x, y, name=_TFOUTPUT1)
         self._run_test_case(func, [_OUTPUT, _OUTPUT1], {_INPUT: x_val, _INPUT1: y_val})
 
+    @check_opset_min_version(10)
+    def test_is_finite(self):
+        x_val = np.array([5.0, 4.8, 6.8, np.inf, np.nan], dtype=np.float32)
```
jignparm, line 3229: Sometimes data has negative-infinity, so might be something to check.

tf2onnx/onnx_opset/math.py
```diff
@@ -566,10 +566,23 @@ def version_12(cls, ctx, node, **kwargs):
         del node.attr["N"]
 
 
-@tf_op("MatrixInverse", onnx_op="Inverse")
-class Inverse:
```
jignparm, line 570: BTW, are we dropping MatrixInverse operator? Seems to be useful (or was this a duplicate implementation?)

tf2onnx/onnx_opset/tensor.py
```diff
@@ -625,6 +625,16 @@ def version_7(cls, ctx, node, **kwargs):
...
+    @classmethod
+    def version_11(cls, ctx, node, **kwargs):
+        dim_node = node.inputs[1]
+        if dim_node.is_const():
+            node.type = "Unsqueeze"
+            dim = dim_node.get_tensor_value()
+            node.set_attr("axes", [dim])
+            ctx.remove_input(node, node.input[1])
+            return
+        raise ValueError("non-const dim is not supported")
```
jignparm, line 638: Is ver_11 same as ver_7, but without the negative axis check? ... if **yes**, the should we just call ver_7 (less code)? If **no**, then ignore this comment :)

tf2onnx/onnx_opset/math.py
guschmue: the problem with MatrixInverse is that it depends on the Inverse operator that was taken out of opset-12 and if it is not in onnx it can't be used in onnx_opset/*, aka doing node.domain is MS_DOMAIN can't be in onnx_opset/*.
But that same code is also here:
https://github.com/onnx/tensorflow-onnx/blob/master/tf2onnx/custom_opsets/ms.py#L107
so this works as long you enable ms-domain.

jignparm: Ok, that makes sense. It looks like Inverse was dropped after 2 other ops Onnx ops were dropped 

Yup, agreed that MS_DOMAIN cannot be used in any general ops.

tf2onnx/onnx_opset/tensor.py
guschmue: opset-11 added support for the negative axes.

PR #937, opened 2020-05-28
tutorial how to convert efficientdet to onnx

guschmue:


jignparm:
Couldn't find a way to comment at the code location, so pasting the snippet below.
Should this be `signatures` instead?  Seems to be running fine, so perhaps this variable is ignored.

```python
       self.signitures = {
            'image_arrays': raw_images,
            'prediction': detections,
        }
        return self.signitures
```

guschmue:
signitures - good catch. But we need to spell it like this because the baseclass defines it like that and driver.export()  is using it.

PR #942, opened 2020-05-29
Jignparm/matrixdiagpartv3

jignparm:
Add non-loop version of MatrixDiagPartV3, opset=12


guschmue:
lgtm. But I think Randy is much more qualified to review.

PR #948, opened 2020-06-02
MatrixDiagPartV3: Change consts to dynamic ops 

jignparm:
The input data shape may contain -1 in one of the dimensions, so the shape is not always known before hand.

This PR handles the scenario of a -1 in the shape, which is required by MatrixDiagSetV3

### DIFF
--- a/tf2onnx/onnx_opset/tensor.py
+++ b/tf2onnx/onnx_opset/tensor.py
@@ -1859,6 +1859,7 @@ class MatrixDiagPartV2V3:
         const_zero_float, const_neg_one_float = mkconsts([[0], [-1]], np.float32)
         const_zero, const_one, const_neg_one, const_neg_two, const_pad_vals, const_t = \
             mkconsts([[0], [1], [-1], [-2], pads, [-1, 1]])
+        const_zero_scalar, const_one_scalar, const_neg_one_scalar = mkconsts([0, 1, -1])
 
         m_shape = ctx.make_node('Shape', [node.input[0]]).output[0]
         xlen = ctx.make_node('Gather', [m_shape, const_neg_one]).output[0]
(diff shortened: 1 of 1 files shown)


tf2onnx/onnx_opset/tensor.py
```diff
@@ -1835,94 +1835,97 @@ class MatrixDiagPartV2V3:
...
         m_padded = ctx.make_node('Pad', [m, const_pad_vals, node.input[2]])
 
         # starting indexes for super diagonals
-        xstart_0 = ctx.make_node('Cast', [k0.output[0]], attr={'to': TensorProto.FLOAT})
+        xstart_0 = ctx.make_node('Cast', [k0], attr={'to': TensorProto.FLOAT})
         xstart_1 = ctx.make_node('Max', [const_zero_float, xstart_0.output[0]])
         xstart_2 = ctx.make_node('Cast', [xstart_1.output[0]], attr={'to': TensorProto.INT64})
         xstart_3 = ctx.make_node('Add', [xstart_2.output[0], const_neg_one])
-        xstart_4 = ctx.make_node('Range', [k1_scalar.output[0], xstart_3.output[0], const_neg_one])
+        xstart_4 = ctx.make_node('Range', [k1_scalar, xstart_3.output[0], const_neg_one])
```
RandySheriffH, line 1893: Is const_neg_one a scalar?

jignparm: It's not a scalar, but it seems to work successfully with Range() function in opset 11 (i.e. unit tests pass).

RandySheriffH: What if I am with opset 12 then?

jignparm: Both opset 11 and 12 are passing unit tests (see checks).

Issue #944, opened 2020-05-30
RandomStandardNormal causes "not supported" error on master


chriselion:
**Describe the bug**
RandomStandardNormal was supported in the latest release, but converting a graphdef with it using the latest master produces an error:

```
2020-05-29 17:17:12,706 - ERROR - Tensorflow op [policy_1/random_normal/RandomStandardNormal: RandomStandardNormal] is not supported
2020-05-29 17:17:12,707 - ERROR - Unsupported ops: Counter({'RandomStandardNormal': 1})
```

**Urgency**
Not causing any immediate problems (we still haven't switched fully to onnx), but potentially a regression.

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): MacOS 10.14.6
- Tensorflow Version: tested on 1.15.2 and 2.2.0
- Python version: Python 3.7.6

**To Reproduce**
Download and uncompress attached [frozen_graph_def.pb.zip](https://github.com/onnx/tensorflow-onnx/files/4704834/frozen_graph_def.pb.zip)

Logs:
```
$ python -m tf2onnx.convert --graphdef frozen_graph_def.pb --inputs vector_observation:0 --outputs action_probs:0,action:0
...
2020-05-29 17:17:12,654 - INFO - Using tensorflow=1.15.2, onnx=1.6.0, tf2onnx=1.6.0/9f5d63
2020-05-29 17:17:12,654 - INFO - Using opset <onnx, 8>
2020-05-29 17:17:12,706 - ERROR - Tensorflow op [policy_1/random_normal/RandomStandardNormal: RandomStandardNormal] is not supported
2020-05-29 17:17:12,707 - ERROR - Unsupported ops: Counter({'RandomStandardNormal': 1})
2020-05-29 17:17:12,709 - INFO - Optimizing ONNX model
2020-05-29 17:17:12,752 - INFO - After optimization: Identity -16 (16->0)
2020-05-29 17:17:12,754 - INFO - 
2020-05-29 17:17:12,754 - INFO - Successfully converted TensorFlow model frozen_graph_def.pb to ONNX
2020-05-29 17:17:12,754 - INFO - To export ONNX model to file, please run with `--output` option
```

**Expected behavior**
tf2onnx.convert should report no errors. Here are the logs for the same file and commandline using 1.5.6 installed from pypi:
```
[...]

PR #949, opened 2020-06-02
Add new pattern for RandomStandardNormal op in TF2

jignparm:
When mean=0.0 and scale=1.0 for RandomStandardNormal, the TF2 optimizer Grappler) will replace the downstream Mul and Add operators with Identity ops instead. So we need to match this pattern to translate the RandomStandardNormal accurately.


jignparm:
Fixes #944

PR #955, opened 2020-06-04
Enable tf22 ci

RandySheriffH:
Enable UT with tensorflow 2.2.0.

### DIFF
--- a/ci_build/azure_pipelines/unit_test.yml
+++ b/ci_build/azure_pipelines/unit_test.yml
@@ -6,7 +6,7 @@ stages:
     - template: 'templates/job_generator.yml'
       parameters:
         python_versions: ['3.7']
-        tf_versions: ['1.14.0','1.15.2','2.1.0']
+        tf_versions: ['1.14.0','1.15.2','2.1.0', '2.2.0']
         onnx_opsets: ['']
         job:
           steps:
--- a/ci_build/azure_pipelines/unit_test.yml
+++ b/ci_build/azure_pipelines/unit_test.yml
@@ -6,7 +6,7 @@ stages:
     - template: 'templates/job_generator.yml'
       parameters:
         python_versions: ['3.7']
-        tf_versions: ['1.14.0','1.15.2','2.1.0', '2.2.0']
+        tf_versions: ['1.14.0','1.15.2','2.1.0','2.2.0']
         onnx_opsets: ['']
         job:
           steps:
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -202,6 +202,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
             self._test_expand_dims_more_unknown_rank(i)
 
     @check_opset_min_version(9, "ConstantOfShape")
+    @check_opset_after_tf_version("2.2", 12, "ConstantOfShape")
     def test_eye_non_const1(self):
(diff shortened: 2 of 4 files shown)


tests/test_backend.py
```diff
@@ -202,6 +202,7 @@ def test_expand_dims_more_unknown_rank(self):
             self._test_expand_dims_more_unknown_rank(i)
 
     @check_opset_min_version(9, "ConstantOfShape")
+    @check_opset_after_tf_version("2.2", 12, "ConstantOfShape")
```
guschmue, line 205: uh, this is a problem. Why do we need this?

tests/test_backend.py
```diff
@@ -223,6 +224,7 @@ def func(x):
 
     @check_tf_min_version("1.11", "eye has bug when version is below 1.11")
     @check_opset_min_version(9, "ConstantOfShape")
+    @check_opset_after_tf_version("2.2", 12, "ConstantOfShape")
```
guschmue, line 227: Why do we need this?

tests/test_backend.py
guschmue: basically, onnx has ConstantOfShape that has not changed since opset-9. So why do we depend on opset-12?

RandySheriffH: Because the implementation of MatrixDiag...V3 depends a lot on opset 12 Max/Min which accept integer as input.

tests/test_backend.py
RandySheriffH: Same as above.

ci_build/azure_pipelines/unit_test.yml
```diff
@@ -6,7 +6,7 @@ stages:
     - template: 'templates/job_generator.yml'
       parameters:
         python_versions: ['3.7']
-        tf_versions: ['1.14.0','1.15.2','2.1.0']
+        tf_versions: ['1.14.0','1.15.2','2.1.0', '2.2.0']
```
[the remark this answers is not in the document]
RandySheriffH: Done.

tests/test_backend.py
guschmue: check_opset_after_tf_version("2.2", 12, "ConstantOfShape") ... the last parameter is the reason. If I read your comment correctly ConstantOfShape is wrong.
But I think this is a something else wrong here:

We handle tf.eye() here:
https://github.com/onnx/tensorflow-onnx/blob/master/tf2onnx/rewriter/eye_rewriter.py#L26

see the where the fix would be ?

RandySheriffH: Have changed last parameter to "MatrixDiag", does it make sense now?

guschmue: no no.
Look at the re-writer - it is suppose MatrixDiag out and replace it with the onnx EyeLike op.
But it doesn't know about the newer MatrixDiag* versions.

guschmue: I mean why would we go via matrix diag when we have a eye operator.

guschmue: we have fix this in the rewriter because most don't have opset-12 (ie. windows, tensorrt) and for things we supported since ages we should avoid requiring such a new opset.

RandySheriffH: Done, fixed rewriter.

PR #963, opened 2020-06-08
update the out-dated shape in _handle_node_having_branches()

daquexian:
the shape of node in `_handle_node_having_branches()` should be updated as what in `_switch_transpose_and_node`


guschmue:
super, thanks!

PR #969, opened 2020-06-11
UnifyConst

RandySheriffH:
move mkconst into graph to avoid duplicated nodes.

### DIFF
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -405,7 +405,6 @@ class Graph(object):
         if target is None:
             target = []
         self._nodes = []
-        self._consts = {}
         self._nodes_by_name = {}
         self._output_to_node_name = {}
         self.shapes = {}
@@ -490,13 +489,7 @@ class Graph(object):
         consts = []
         for value in values:
             np_val = np.array(value).astype(np_type)
-            key = str(np_val) + "_" + str(np_val.dtype)
-            if key in self._consts:
-                consts.append(self._consts[key])
-            else:
-                const_node = self.make_const(utils.make_name("const"), np_val, skip_conversion, raw)
-                self._consts[key] = const_node.output[0]
-                consts.append(const_node.output[0])
+            consts.append(self.make_const(utils.make_name("const"), np_val, skip_conversion, raw))
         return consts
 
     def make_const(self, name, np_val, skip_conversion=False, raw=True):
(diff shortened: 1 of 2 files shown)


tf2onnx/graph.py
```diff
@@ -492,6 +501,11 @@ def make_const(self, name, np_val, skip_conversion=False, raw=True):
             skip_conversion: bool, indicate whether this created node would be mapped during conversion.
             raw: whether to store data at field of raw_data or the specific field according to its dtype
         """
+
+        key = str(np_val) + "_" + str(np_val.dtype)
+        if key in self._consts:
+            return self._consts[key]
+
```
jignparm, line 508: It looks like 'name' will be ignored of subsequent const tensors, if their value is reserved already. So, we cannot create  a const with a specific name that we want (if it's value is already in the _consts dict). 

Don't know if that's OK or not -- could be a perfectly valid thing to do, but just not fully sure if it'll break initialiizers.

RandySheriffH: Reverting back this part, good point!

tf2onnx/graph.py
```diff
@@ -484,6 +485,20 @@ def inputs(self):
...
+        for value in values:
+            np_val = np.array(value).astype(np_type)
+            key = str(np_val) + "_" + str(np_val.dtype)
+            if key in self._consts:
+                consts.append(self._consts[key])
+            else:
+                const_node = self.make_const(utils.make_name("const"), np_val, skip_conversion, raw)
+                self._consts[key] = const_node.output[0]
+                consts.append(const_node.output[0])
+        return consts
```
jignparm, line 500: The logic looks ok, only thing is that the API seems a bit inconsistent. See explanation below:

*   The return type of **make_consts()**  is a list of 'Node.output[0]', which is a list of strings. 
*   Whereas the return type of the function below, **make_const()**, is the Node object itself.

It feels a bit odd that the two similar functions are  returning  very different things (i.e. Node objects versus string lists)

Maybe this function can be refactored to return a list of Nodes, and then the caller can just use a comprehension to extract out the output[0] values.  Thoughts?

RandySheriffH: Then how about we rename the new function like make_const_names or something else, since we are creating it to facilitate the consts creation as much as we could?

PR #970, opened 2020-06-12
Fix issue with shape of LSTM node

buddhapuneeth:
With the recent support of stacked LSTM #925 this issue is introduced.
Variable context.hidden_size is now a list. So for the ith LSTM, we need to assign the ith value instead of entire list.


guschmue:
thanks @buddhapuneeth !

Issue #967, opened 2020-06-09
TF2.2 PB file conversion to ONNX “<function rewrite_single_direction_lstm” failure


Yaduvir-Singh-Bose:
**Describe the bug**
I have trained a TF2.2 model, which i am trying to convert to onnx. it gives me “<function rewrite_single_direction_lstm” failure. please find saved_model and TF2.2.0 PB file.

**Urgency**
This is blocking issue for our project, until we run the model on nvidia xavier tensorrt we can not make any progress.This issue is very urgent for us.

**System information**
TensorRT Version : 7.0.0.11
GPU Type : Tesla K80
Nvidia Driver Version : 450.36.06
CUDA Version : CUDA Version: 11.0
CUDNN Version : 8.0
Operating System + Version : Ubuntu x86_64
Python Version (if applicable) : 3.7.7
TensorFlow Version (if applicable) : 2.2.0
PyTorch Version (if applicable) :
Baremetal or Container (if container which image + tag) : barematel

**To Reproduce**
(tensor2) ubuntu@usva-asd-ml-ear1:~/notebook/cnn-facial-landmark$ python -m tf2onnx.convert --graphdef frozen_graph_t2.pb --output frozen_graph.onnx --inputs image_tensor:0 --outputs layer6/final_dense:0 --fold_const --opset 11

2020-06-09 15:13:01,934 - INFO - Froze 0 variables.
INFO:tensorflow:Converted 0 variables to const ops.
2020-06-09 15:13:01,948 - INFO - Converted 0 variables to const ops.
2020-06-09 15:13:02,322 - INFO - Using tensorflow=2.2.0, onnx=1.7.0, tf2onnx=1.6.1/aafc83
2020-06-09 15:13:02,322 - INFO - Using opset <onnx, 11>
2020-06-09 15:13:03,327 - ERROR - rewriter <function rewrite_single_direction_lstm at 0x7f75c1a27cb0>: exception (‘switch_true map/while/map/while/loop_counter_switch/_15 has unexpected count of consumers:’, [‘map/while/loop_body_control/_21’, ‘Func/map/while/body/_1/input/_44’])

**Expected behavior**
tf2onnx.convert should convert PB file to ONNX file.

**Screenshots**
Error Logs in Step's to reproduce.Here are the links for saved_model and PB file.
frozen_graph.pb:
https://drive.google.com/file/d/1DPyry67p6VGOTR6qTfCDT5Kr_QsuLRa_/view?usp=sharing
[...]

PR #975, opened 2020-06-16
handle 2 switch output consumers

guschmue:


guschmue:
https://github.com/onnx/tensorflow-onnx/issues/967

PR #978, opened 2020-06-19
Enhance Transpose optimizer "Mul" handler

### LINKED ISSUE #977: Inputs_as_NCHW adds Transposes
Hello,
I am converting a model from TF with the --inputs-as-nchw option
And the resulting model still contains a good amount of Transposes.

Here is the model
https://drive.google.com/file/d/17qLexicRukvDqrw3SjFh-0BSjj_ABxjR/view?usp=sharing

For examples
image
becomes
image
But there are similar issues downstream in the model.

I am calling as follows:
```
python -m tf2onnx.convert --graphdef .\UnityProject\Assets\Resources\yolov2-tiny-arcityv3.pb --output output.onnx --inputs input:0 --outputs output:0 --inputs-as-nchw input:0
```

and I am on 1.5.6

Thanks!

jignparm:
Fixes #977 

Previously Tranpose_Mul handler could only handle constant multipliers of shape (1). Added support fo shape(N).

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -482,8 +482,21 @@ class TransposeOptimizer(GraphOptimizerBase):
                 self._g.remove_node(node.name)
                 return True
 
-        # if the shape is () or (1), we just move transpose after the mul
-        if not multiplier.shape or (len(multiplier.shape) == 1 and multiplier.shape[0] == 1):
+        # if the shape is (), we just move transpose after the mul
+        if not multiplier.shape:
+            return self._switch_transpose_and_node(node, trans)
+
+        # if multiplier is 1-D
+        if len(multiplier.shape) == 1:
+            if multiplier.shape[0] == 1:
+                # shape is (1)
+                return self._switch_transpose_and_node(node, trans)
+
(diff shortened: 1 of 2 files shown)


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -482,8 +482,21 @@ def _mul_handler(self, trans, node):
...
 
-        # if the shape is () or (1), we just move transpose after the mul
-        if not multiplier.shape or (len(multiplier.shape) == 1 and multiplier.shape[0] == 1):
+        # if the shape is (), we just move transpose after the mul
+        if not multiplier.shape:
+            return self._switch_transpose_and_node(node, trans)
+
+        # if multiplier is 1-D
+        if len(multiplier.shape) == 1:
+            if multiplier.shape[0] == 1:
```
RandySheriffH, line 491: Is it possible for "if..." at line 491 be merged to 486 like:
if not multiplier.shape or multipier.shape == [1]:
?

jignparm: Updated unit test to see if 1D const tensors can be jumped over.

For the 'if' merging, it seems cleaner to separate handling of scalars and 1D tensors into 2 blocks (mostly for readability)

PR #984, opened 2020-06-24
Add support for TF2.x saved_models from TFHub, as well as --tag & -concrete_function  cmd line parameters

jignparm:
Added support for TF2.x saved_models from TFHub

* Added --tag parameter.  Lets user specify tag set, in case multiple are available. Also handles the case of an empty tag value (rare, but seen in some models)

* Added --concrete_function parameter. Some models do not contain signature_def values. Instead, they rely on concrete function definitions under the __call__ method, which require fetching using a TensorSpec().


jignparm:
>We should add those to the README.

Good idea! Wanted to try out a few more models before putting it in the official documentation -- in case any changes or fine-tuning is needed.

PR #995, opened 2020-07-03
Fixed dropout rewriter to properly read ratio

TomWildenhain-Microsoft:


guschmue:
test_nn_dropout() is failing, looks like it input1:0 is not in the graph:
onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Invalid Feed Input Name:input1:0

tf2onnx/rewriter/dropout_rewriter.py
```diff
@@ -60,17 +71,20 @@ def rewrite_dropout(g, ops):
...
+            nodes_to_remove = []
+            for node in match.get_nodes():
+                if node.name != input3.name:
+                    nodes_to_remove.append(node)
+            if g.safe_to_remove_nodes(nodes_to_remove):
+                for n in nodes_to_remove:
+                    g.remove_node(n.name)
+            else:
+                logger.error("Nodes replaced by dropout node cannot be removed because intermediate results are "
+                             "referenced elsewhere in graph")
```
jignparm, line 88: in the 'else' case, is there another option to move forward and continue with model conversion? Or it's a true halting condition?

TomWildenhain-Microsoft: It logs an error but does not halt conversion.  The end result is that the intermediate nodes are left in the graph, but conversion should complete successfully.

tf2onnx/rewriter/dropout_rewriter.py
```diff
@@ -60,17 +71,20 @@ def rewrite_dropout(g, ops):
...
-    # remove dropout if its ratio is 1.0
-    for node in g.get_nodes():
-        if node.type == "Dropout" and node.get_attr("ratio").f == 1.0:
-            g.replace_all_inputs(g.get_nodes(), node.output[0], node.input[0])
-            g.remove_node(node.name)
+            nodes_to_remove = []
+            for node in match.get_nodes():
+                if node.name != input3.name:
+                    nodes_to_remove.append(node)
+            if g.safe_to_remove_nodes(nodes_to_remove):
```
guschmue, line 83: we should maybe use make_sure()  and raise if this happens. Don't think tensorflow would ever give us such a graph.

tf2onnx/rewriter/dropout_rewriter.py
jignparm: Ok, great. Seems like it might be better to use logger.warning(), or else to just ignore any messaging to the user. An error message is usually for fatal events.

tf2onnx/rewriter/dropout_rewriter.py
```diff
@@ -37,21 +40,29 @@ def rewrite_dropout(g, ops):
...
-                    OpTypePattern(None, name="input3")
+                    OpTypePattern("*", name="input3")
                 ])
             ]),
             OpTypePattern("Mul", name="input2"),
         ]),
     ]
     for pattern in patterns:
-        matcher = GraphMatcher(pattern, allow_reorder=True)
+        matcher = GraphMatcher(pattern, allow_reorder=False)
```
guschmue, line 50: There is a Add in the pattern ... tensorflow could switch the inputs

TomWildenhain-Microsoft: There is also a GreaterEqual for which switching the inputs would be incorrect.  Finding such a pattern is unlikely but possible, and would not be equivalent to a Dropout node.  I'll switch allow_reorder to True for now, and fix it with #998

guschmue: there is a pattern as input to GreaterEqual which prevents it in this case from being switched. Basically tensorflow can't switch the Random* with the 2nd input and we are looking for it so we can't switch the inputs either.
Not saying that the reorder can't be an issue in other cases.

tf2onnx/rewriter/dropout_rewriter.py
TomWildenhain-Microsoft: I think such a graph is technically possible if the pattern we detect is not actually from a dropout node (i.e. the user crafted the graph shape by hand from RandomUniform/floor/etc.).  Also, Keras graphs seem to sometimes output the intermediate values.

guschmue: if a tensorflow method like tf.nn.dropout() emits multiple nodes its kind of sealed, aka the nodes in the middle are not known outside.
If we mistakenly identify something that is not a dropout but handle it as dropout, no point of doing anything other than raising an exception.
Keras might put all nodes as resource types into the output but we will filter those out.

TomWildenhain-Microsoft: I agree regarding Keras.  If we mistakenly match against non-dropout nodes, I believe the correct behavior is to not replace the nodes at all.  We should not create a dropout node, not raise an exception, and just directly convert all the component nodes.  I think it also makes sense to display a warning, since we expect such cases to be rare and there may be a different problem (Keras filtering may have failed).  I'll make a new commit with that behavior and see what you think.

guschmue: sure, that works as well and we do this in other rewriters.

TomWildenhain-Microsoft: Also, if we can't get the ratio (if it isn't constant), we may want to similarly skip replacement (and show a warning) instead of setting it to 0.  If the user wants the ratio to be an argument, we can actually simulate that behavior by directly converting the component nodes.  What do you think?

guschmue: But than the model does not convert anymore. If you are the user - what would you prefer ?
For inference - it is ok to remove dropout regardless of rate, keras does it also this way if you run predict().
For training - sure, we should pass on the rate correctly, most likely that requires a onnx change.

TomWildenhain-Microsoft: As I understand it, if we skip replacing the dropout node, the model will still successfully convert.  The constituent nodes (RandomUniform, Mul, etc.) will just be converted individually.  The result is a functioning model which perfectly matches the behavior of the original.

guschmue: we'd take a performance hit on this since now there are 2+ ops where we had 0 ops before and depending on the model we'll take that hit multiple times.
What is wrong by just removing the dropout related ops ?

PR #1020, opened 2020-07-24
Implemented MaxPool3D and AvgPool3D

TomWildenhain-Microsoft:


ralovich:
In my limited testing the 3d pooling and convolutions appears to work. Thank you!

PR #1045, opened 2020-08-07
Conv3DBackpropInputV2

ralovich:
add support of NDHWC Conv3DBackpropInputV2


guschmue:
We might have just fixed the same here:
https://github.com/onnx/tensorflow-onnx/pull/1041

PR #1046, opened 2020-08-07
Fixed bug in padding calculation for padding='SAME' when dilations>1

TomWildenhain-Microsoft:


TomWildenhain-Microsoft:
The added test fails under the old code and passes with this fix.

PR #1056, opened 2020-08-12
Use the same syntax to replace an node input

xadupre:
Preliminary work to build a forward index to keep track of all nodes ingesting a specific input.

### DIFF
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -1262,18 +1262,6 @@ class Graph(object):
     def replace_inputs(self, node, new_inputs):
         """Replace node inputs."""
         assert isinstance(node, Node) and isinstance(new_inputs, list)
-
-        for old_input in node.input:
-            to_ops = self._input_to_node_name.get(old_input, None)
-            if to_ops is not None and old_input in to_ops:
-                # To avoid issues when a node
-                # takes twice the same entry.
-                to_ops.remove(old_input)
-
-        for input_name in new_inputs:
-            assert isinstance(input_name, six.text_type)
-            self._register_input_name(input_name, node)
-
         node.input = new_inputs
         return True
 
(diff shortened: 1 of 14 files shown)


tf2onnx/graph.py
```diff
@@ -1134,16 +1139,22 @@ def dump_node_statistics(self):
         return op_cnt
 
     @staticmethod
-    def remove_input(node, to_be_removed):
+    def remove_input(node, to_be_removed, i=None):
```
TomWildenhain-Microsoft, line 1142: Might be cleaner to do remove_input_at_index and replace_input_at_index instead of making i an optional argument.

tf2onnx/optimizer/back_to_back_optimizer.py
```diff
@@ -83,7 +83,7 @@ def _optimize_cast(g, node, consumer_nodes):
             type2 = pnode.get_attr('to').i
             if type1 == type2:
                 for node2 in consumer_nodes:
-                    node2.input[0] = node.input[0]
```
TomWildenhain-Microsoft, line 86: If we want to keep the old syntax, you could just let input be a subclass of list that tracks when its elements are changed and updates the cache.  But this works too.

xadupre: We could do that. I chose to replace this syntax by a method because users usually do not expect this syntax to do anything complex such as calling a method of the graph owning this node and updating some indices. And the documentation of the method is easier to find with an explicit name.

tf2onnx/graph.py
xadupre: Renamed i into input_index. Not sure we should create another function which does almost the same thing.

tf2onnx/optimizer/back_to_back_optimizer.py
TomWildenhain-Microsoft: Ah, makes sense.

tf2onnx/graph.py
TomWildenhain-Microsoft: That works too.

PR #1061, opened 2020-08-12
Faster insertion of operator cast, missing replacement in #1059

xadupre:


guschmue:
I do not understand what this change is about.

Issue #1072, opened 2020-08-21
Transpose Optimization error when working less than 4 dims


NikolasMarkou:
**Describe the bug**
At many points in the transpose optimization it is assumed that the shape is of size 4. 

**Urgency**
This is a non blocker, if the optimization fails it just continues with the next optimization

**System information**
- Linux Mint 18.04
- Tensorflow Version: 1.15
- Python version: 3.6.9

**To Reproduce**
Add a Transpose optimization with input shape less than 4
```
2020-08-21 13:44:13,975 - WARNING - Failed to apply optimize_transpose
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/tf2onnx/optimizer/__init__.py", line 52, in optimize_graph
    graph = opt.optimize(current) or graph
  File "/usr/local/lib/python3.6/dist-packages/tf2onnx/optimizer/optimizer_base.py", line 41, in optimize
    graph = self._optimize(graph)
  File "/usr/local/lib/python3.6/dist-packages/tf2onnx/optimizer/transpose_optimizer.py", line 135, in _optimize
    return self._apply_optimization(graph, self._optimize_at_current_graph_level)
  File "/usr/local/lib/python3.6/dist-packages/tf2onnx/optimizer/optimizer_base.py", line 62, in _apply_optimization
    graph = optimize_func(graph)
  File "/usr/local/lib/python3.6/dist-packages/tf2onnx/optimizer/transpose_optimizer.py", line 148, in _optimize_at_current_graph_level
    if self._handle_nhwc_tranpose(n):
  File "/usr/local/lib/python3.6/dist-packages/tf2onnx/optimizer/transpose_optimizer.py", line 281, in _handle_nhwc_tranpose
    return op_handler(trans, p)
  File "/usr/local/lib/python3.6/dist-packages/tf2onnx/optimizer/transpose_optimizer.py", line 660, in _slice_handler
    return self._switch_transpose_and_node(node, trans)
  File "/usr/local/lib/python3.6/dist-packages/tf2onnx/optimizer/transpose_optimizer.py", line 262, in _switch_transpose_and_node
    new_shape = [shape[i] for i in NHWC_TO_NCHW]
  File "/usr/local/lib/python3.6/dist-packages/tf2onnx/optimizer/transpose_optimizer.py", line 262, in <listcomp>
    new_shape = [shape[i] for i in NHWC_TO_NCHW]
IndexError: list index out of range
```

**Expected behavior**
[...]

PR #1073, opened 2020-08-21
#1072 Fixed transpose_optimizer shape assumptions

NikolasMarkou:
By adding additional checks before entering the change phase. There is no change in logic or optimization

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -101,10 +101,8 @@ class TransposeOptimizer(GraphOptimizerBase):
                 input_shape = self._g.get_shape(op.input[0])
                 if not input_shape:
                     continue
-
                 if (is_nchw_transpose(op) and len(input_shape) == 4 and
                         (input_shape[3] == 1 or (input_shape[1:3] == [1, 1]))) \
-
                         or (is_nhwc_transpose(op) and (input_shape[1] == 1 or (input_shape[2:4] == [1, 1]))):
                     new_shape = _calculate_new_shape(self._g, op)
                     # replace transpose with reshape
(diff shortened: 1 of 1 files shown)


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -101,8 +101,8 @@ def _calculate_new_shape(graph, op):
                 input_shape = self._g.get_shape(op.input[0])
                 if not input_shape:
                     continue
-
-                if (is_nchw_transpose(op) and (input_shape[3] == 1 or (input_shape[1:3] == [1, 1]))) \
+                if (is_nchw_transpose(op) and len(input_shape) == 4 and
```
TomWildenhain-Microsoft, line 104: Doesn't `is_nchw_transpose` already ensure the length of the input shape is 4?

NikolasMarkou: You are right it does

TomWildenhain-Microsoft: I take that back.  It seems like the whole issue with #1072 is that somehow a transpose node has perm and shape properties with different dimensions.

NikolasMarkou: That somehow was the bug i saw, but when reading the code the check seemed redundant, i put it there again.

NikolasMarkou:
@TomWildenhain-Microsoft what is the requested change here ?

NikolasMarkou:
@guschmue

TomWildenhain-Microsoft:
In #1072 I mentioned that this check is only needed if the perms attribute of the transpose is longer than its shape would allow.  We could merge this as a temporary fix but I'm still wondering why this situation is occurring.  It suggests the true bug is somewhere else.

TomWildenhain-Microsoft (approved):

PR #1074, opened 2020-08-21
Feature: Added optimization step to remove upsample layers with all ones in scale

NikolasMarkou:
Upsample operations with all ones in scale are redundant and can be safely removed. In this commit I replace them with an Identity node which is then removed by the Identity optimization step.


guschmue:
Thanks a ton for the contribution!

PR #1060, opened 2020-08-12
Use the same syntax to replace an node input (2) + optimize replace_all_inputs

xadupre:
Measurss for **EfficientNetB2**:

Before:

````
21.492 profile  profile_conversion_time.py:58
`- 21.492 convert  profile_conversion_time.py:53
   `- 21.492 spy_convert  profile_conversion_time.py:30
      |- 20.699 spy_convert_in  profile_conversion_time.py:35
      |  `- 20.698 process_tf_graph  tf2onnx\tfonnx.py:335
      |     |- 10.522 tensorflow_onnx_mapping  tf2onnx\tfonnx.py:211
      |     |  |- 3.105 version_1  tf2onnx\onnx_opset\nn.py:329
      |     |  |  `- 3.077 conv_convert_inputs  tf2onnx\onnx_opset\nn.py:64
      |     |  |     |- 1.601 insert_new_node_on_output  tf2onnx\graph.py:1191
      |     |  |     |  `- 1.522 replace_all_inputs  tf2onnx\graph.py:1228
      |     |  |     |     |- 0.824 replace_all_inputs  tf2onnx\graph.py:1228
      |     |  |     |     |  |- 0.483 get_body_graphs  tf2onnx\graph.py:324
      |     |  |     |     |  |  `- 0.268 _graph_check  tf2onnx\graph.py:386
      |     |  |     |     |  `- 0.255 [self]  
      |     |  |     |     `- 0.411 get_body_graphs  tf2onnx\graph.py:324
      |     |  |     |        `- 0.233 _graph_check  tf2onnx\graph.py:386
      |     |  |     `- 1.234 find_output_consumers  tf2onnx\graph.py:1214
      |     |  |        |- 0.672 find_output_consumers  tf2onnx\graph.py:1214
      |     |  |        |  `- 0.486 get_body_graphs  tf2onnx\graph.py:324
      |     |  |        |     `- 0.269 _graph_check  tf2onnx\graph.py:386
      |     |  |        `- 0.396 get_body_graphs  tf2onnx\graph.py:324
      |     |  |           `- 0.226 _graph_check  tf2onnx\graph.py:386
      |     |  |- 2.523 version_1  tf2onnx\onnx_opset\tensor.py:124
      |     |  |  `- 2.518 replace_all_inputs  tf2onnx\graph.py:1228
      |     |  |     |- 1.469 replace_all_inputs  tf2onnx\graph.py:1228
[...]

PR #1077, opened 2020-08-26
Add a script to profile the conversion of a model

xadupre:


benchmarks/profile_conversion_time.py
```diff
@@ -0,0 +1,101 @@
+# coding: utf-8
```
guschmue, line 1: We want to limit top level directories - can we stick this under tools?

ci_build/azure_pipelines/templates/unit_test.yml
```diff
@@ -14,3 +14,14 @@ steps:
     condition: succeededOrFailed()
     env:
         CI_ONNX_OPSET: '${{ onnx_opset }}'
+
+  - bash: |
+      export TF2ONNX_TEST_BACKEND=$CI_ONNX_BACKEND
```
guschmue, line 19: Our ci pipleline is already slow, don't think we want to add benchmarks to it.

benchmarks/profile_conversion_time.py
```diff
@@ -0,0 +1,101 @@
...
+"""
+Profiles the conversion of a Keras model.
+"""
+import fire
+import tensorflow as tf
+from tf2onnx import tfonnx
+from tensorflow.keras.applications import MobileNet
```
guschmue, line 8: Iffy to depend on tf.keras since we must support older tf-1.x versions.

xadupre: We can remove the script from CI, no issue with that, it takes 30 seconds. My plan was to only run it with the latest version of tensorflow. However, the script is failing and it was working a month ago. So I was wondering what change could have caused that. Either my script is wrong, either there is something not covered by the unit tests...

benchmarks/profile_conversion_time.py
xadupre: Done.

benchmarks/profile_conversion_time.py
guschmue: can we not start using fire ? We don't use it anywhere else and want to limit the dependencies.

guschmue:
Isn't this PR seems to be included in here https://github.com/onnx/tensorflow-onnx/pull/1060 ?

sdpython:
I merged it to check it was working. I was about to remove the merge.

ci_build/azure_pipelines/templates/unit_test.yml
xadupre: Fixed

benchmarks/profile_conversion_time.py
xadupre: Dependency removed.

guschmue (approved):

Issue #679, opened 2019-08-28
LogSoftMax should handle axis


zhanghuanrong:
Currently in xlnet  squad model, there is one logSoftmax working on the third dimension(2, in fact -1 in tf) rather than default (1). Please fix this.

PR #1084, opened 2020-09-01
Fix the missing axis when LogSoftmax is in graph

peterjc123:
Fixes https://github.com/onnx/tensorflow-onnx/issues/679.


guschmue:
@peterjc123, nice catch! Thanks for the contribution.

PR #1089, opened 2020-09-02
Add cast to same type before equal operator

bedapisl:
- Operands of Equal must be of the same type ([onnx documentation](https://github.com/onnx/onnx/blob/master/docs/Changelog.md#Equal-11)). This PR adds necessary casts before Equal.
- Output of Size operation should be Int64 ([documentation](https://github.com/onnx/onnx/blob/master/docs/Changelog.md#Size-1)). This PR fixes a problem when the output of Size was internally Int32.

This allows for example this model: https://drive.google.com/file/d/105xs9j1PjzKk45PkM5-hGNKxGTdElsCL/view?usp=sharing to be correctly converted and loaded with onnxruntime. Without the proposed changes the converted model is not valid (https://github.com/microsoft/onnxruntime/issues/4997).


tests/test_equal.py
```diff
@@ -0,0 +1,39 @@
...
+
+from backend_test_base import Tf2OnnxBackendTestBase
+
+
+# pylint: disable=missing-docstring,invalid-name,unused-argument,using-constant-test
+# pylint: disable=abstract-method,arguments-differ
+
+class EqualTests(Tf2OnnxBackendTestBase):
+
+    def test_equal_with_different_parameters(self):
```
TomWildenhain-Microsoft, line 20: Put this test in test_backend.py under test_equal.

TomWildenhain-Microsoft (changes requested):
Thanks for your contribution!  Please move the test to test_backend.py

tests/test_equal.py
bedapisl: Done :)

TomWildenhain-Microsoft (approved):
Thanks for the contribution!

PR #1090, opened 2020-09-02
Added support for converting large models

TomWildenhain-Microsoft:


tf2onnx/tf_loader.py
```diff
@@ -96,7 +96,19 @@ def inputs_without_resource(sess, input_names):
     return input_names
 
 
-def from_function(func, input_names, output_names):
+def from_function(func, input_names, output_names, large_model):
```
guschmue, line 99: maybe use large_model=None since one might call this from some random python code

guschmue (changes requested):

tf2onnx/tf_loader.py
guschmue: other than that super lgtm.

guschmue (approved):

OpUs-Nebula:
Are there any plans for this support to be added to TF 1.X aswell?

TomWildenhain-Microsoft:
I believe converting saved models for TF 1.X should already support large models though I haven't tested it.  Do you have a specific model you would like to convert?  If so, please open a GitHub issue.

OpUs-Nebula:
I've tried but i get the error: "ValueError: You passed in an iterable attribute but I cannot figure out its applicable type". It's odd because it seems to compile fine until i get the protobuf size error when i don't pass --saved_model. I do have a specific model in mind. I'm fairly new to ML, so i suppose it could be user error aswell.

OpUs-Nebula:
Just added it

dgoldenberg-audiomack:
If this is fixed, why am I seeing it? see https://github.com/tensorflow/recommenders/issues/274#issuecomment-848071286

TomWildenhain-Microsoft:
Hi @dgoldenberg-audiomack, are you trying to convert your model from tensorflow to onnx?  This repo is the tf2onnx repo.  Our fix is a workaround/hack that lets us convert large TF models.  It isn't a fix to tensorflow itself.

PR #1103, opened 2020-09-15
Add an end2end example with tf.keras

xadupre:


guschmue (approved):

examples/end2end_tfkeras.py
```diff
@@ -0,0 +1,119 @@
...
+print(np.abs(got[0] - expected).max())
+
+########################################
+# Measures processing time.
+print('tf:', timeit.timeit('model.predict(input)',
+                           number=100, globals=globals()))
+print('ort:', timeit.timeit("session.run(None, {'input_1:0': input})",
+                            number=100, globals=globals()))
+
+########################################
```
guschmue, line 72: tflite does not contribute much to a example how to convert tf->onnx.
We should remove it.

examples/end2end_tfhub.py
```diff
@@ -0,0 +1,115 @@
...
+interpreter = tf.lite.Interpreter(model_path='efficientnetb0clas.tflite')
+interpreter.allocate_tensors()
+input_details = interpreter.get_input_details()
+output_details = interpreter.get_output_details()
+print("input_details", input_details)
+print("output_details", output_details)
+index = input_details[0]['index']
+
+
+def tflite_predict(input, interpreter=interpreter, index=index):
```
guschmue, line 94: tflite does not contribute much to a example how to convert tf->onnx.
We should remove it.

guschmue (changes requested):

examples/end2end_tfhub.py
xadupre: Ok, I'll remove it from the example. Do I keep an example about tflite?

guschmue: I think it would confuse users if we have tflite in examples since we don't support it.
Best to remove it completely.

xadupre: fixed

examples/end2end_tfkeras.py
xadupre: fixed

guschmue (approved):

PR #1117, opened 2020-09-29
Fixes #595, support operator InvertPermutation

### LINKED ISSUE #595: Missing op InvertPermutation
Hi. I would like to add the [InvertPermutation](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/invert-permutation) op to this project.
How do I get started?

xadupre:


tf2onnx/onnx_opset/math.py
```diff
@@ -697,3 +697,38 @@ def atan2(y, x):
             op_name_scope=node.name + 'all',
             shapes=[shape], dtypes=[onnx_dtype])
         ctx.replace_all_inputs(node.output[0], last_node.output[0])  # ops=ctx.get_nodes()
+
+
+@tf_op("InvertPermutation")
+class InvertPermutationOp:
+    supported_dtypes = [
```
TomWildenhain-Microsoft, line 704: This seems unused.  What is it for?

tf2onnx/onnx_opset/math.py
```diff
@@ -697,3 +697,38 @@ def atan2(y, x):
...
+
+        neg_node = ctx.make_node(
+            "Neg", inputs=node.input, name=utils.make_name(node.name + '_neg'))
+
+        topk_unused = utils.make_name(node.name + '_topk')
+        topk_indices = utils.make_name(node.name + '_indices')
+        outputs = [topk_unused, utils.port_name(topk_indices, 1)]
+        topk_node = ctx.make_node(
+            "TopK", inputs=[neg_node.output[0], shape_node.output[0]],
+            name=utils.make_name(node.name + '_topk'), outputs=outputs)
```
TomWildenhain-Microsoft, line 726: The output names should be node_name:0 and node_name:1 which I think make_node will assign automatically.  make_name is for making node names, not output names, I think.

TomWildenhain-Microsoft (changes requested):

tf2onnx/onnx_opset/math.py
xadupre: The operator only supports these types. I was that on other operators. I assumed some verification would automatically happen. Based on your remark, I assume not. I'll do that then.

TomWildenhain-Microsoft: Just searched the repo for `supported_dtypes`.  It is usually declared in a function and then iterated through for casting purposes.  I think you might have seen it and mistakenly thought it should be assigned in a class.  You defined it in RFFT, ComplexAbs, and Atan2 but I don't think it should be included.  Sorry I didn't catch that sooner.

TomWildenhain-Microsoft (approved):

PR #1119, opened 2020-09-29
Add cast after Size if expected type is not int64

TomWildenhain-Microsoft:


guschmue (approved):

guschmue:
I saw that one too yesterday :)

PR #1164, opened 2020-11-05
Fix parsing of Sparse ops

TomWildenhain-Microsoft:


guschmue (approved):

guschmue:
ci passed but notify to github is missing.

PR #1165, opened 2020-11-05
fix shape inference for onnx-1.8

guschmue:

we used to set subgraphs as attributes after calling make_node() but called shape inference in make_node() which makes it hard for shape inference to do something useful. Pass the subgraphs to make_node() so they are set before calling shape inference.

### DIFF
--- a/ci_build/azure_pipelines/unit_test.yml
+++ b/ci_build/azure_pipelines/unit_test.yml
@@ -16,7 +16,7 @@ stages:
     - template: 'templates/job_generator.yml'
       parameters:
         python_versions: ['3.7']
-        tf_versions: ['1.14.0','1.15.2','2.3.0']
+        tf_versions: ['1.14.0','1.15.2','2.2.0','2.3.0']
         onnx_opsets: ['']
         job:
           steps:
--- a/ci_build/azure_pipelines/unit_test.yml
+++ b/ci_build/azure_pipelines/unit_test.yml
@@ -16,7 +16,7 @@ stages:
     - template: 'templates/job_generator.yml'
       parameters:
         python_versions: ['3.7']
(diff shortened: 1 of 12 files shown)


tf2onnx/schemas.py
```diff
@@ -156,7 +156,7 @@ def build_onnx_op(node):
     try:
         inferred_model = shape_inference.infer_shapes(model_proto)
     except Exception:  # pylint: disable=broad-except
-        logger.warning(
+        logger.info(
```
TomWildenhain-Microsoft, line 159: I have liked that this is a warning since it often indicates a problem with freezing.

ci_build/azure_pipelines/unit_test.yml
```diff
@@ -16,7 +16,7 @@ stages:
     - template: 'templates/job_generator.yml'
       parameters:
         python_versions: ['3.7']
-        tf_versions: ['1.14.0','1.15.2','2.1.0','2.2.0']
+        tf_versions: ['1.14.0','1.15.2','2.3.0']
```
TomWildenhain-Microsoft, line 19: Add 2.2.0
```suggestion
        tf_versions: ['1.14.0','1.15.2','2.2.0','2.3.0']
```

TomWildenhain-Microsoft (changes requested):

ci_build/azure_pipelines/unit_test.yml
guschmue: done

tf2onnx/schemas.py
guschmue: done

TomWildenhain-Microsoft (approved):

guschmue:
ci pipeline completed but notify to github seems missing ... forced the merge.

PR #1237, opened 2020-12-16
Bug fixes for opset 13 support

TomWildenhain-Microsoft:

### DIFF
--- a/tests/test_optimizers.py
+++ b/tests/test_optimizers.py
@@ -11,7 +11,7 @@ import unittest
 import numpy as np
 from onnx import helper, TensorProto, OperatorSetIdProto
 from backend_test_base import Tf2OnnxBackendTestBase
-from common import unittest_main, group_nodes_by_type, check_opset_min_version, check_opset_max_version
+from common import unittest_main, group_nodes_by_type, check_opset_min_version, check_opset_max_version, get_test_config
 from tf2onnx import utils, constants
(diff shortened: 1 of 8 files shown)


tests/test_optimizers.py
```diff
@@ -973,9 +984,9 @@ def test_duplicated_duplicated_input(self):
 
     def test_duplicated_duplicated_attributes(self):
         # same attr or not
-        node0 = helper.make_node('ReduceSum', inputs=["X"], outputs=["value0"], axes=[0], keepdims=0)
```
TomWildenhain-Microsoft, line 976: ReduceMin doesn't change in opset 13 and is equally effective for testing.

tf2onnx/graph_builder.py
```diff
@@ -55,8 +55,7 @@ def make_slice(self, kwargs, name=None, shapes=None, dtypes=None, return_node=Fa
             inputs = [data, starts, ends, axes, steps]
 
         # pro-process inputs and attr
-        if kwargs:
-            logger.warning("kwargs contains un-used key")
+        utils.make_sure(not kwargs, "kwargs contains un-used key")
```
TomWildenhain-Microsoft, line 58: If the call uses the wrong kwargs, there is a problem with the tf2onnx code.  This should be an error.

guschmue (approved):

PR #1248, opened 2020-12-25
fix sparse_softmax_cross_entropy bug when logits is negative infinity

liangkaihuan:
When all elements of logits are negative infinity, sparse_softmax_cross_entropy will output NAN. Sub the maximum first can fix the problem.


liangkaihuan (commented):
sub the maxmium

TomWildenhain-Microsoft:
Thanks so much for the contribution!  To confirm, you are avoiding the negative infinities entirely by shrinking the inputs to the exp.  Is that correct?

TomWildenhain-Microsoft (approved):

PR #1270, opened 2021-01-13
Added handlers for tflite

TomWildenhain-Microsoft:


tf2onnx/tflite_handlers/__init__.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT license.
+"""tf2onnx.tflite_handlers module"""
```
guschmue, line 3: I think all of the tflite code should not start at the root dir and be under tf2onnx.
The structure at the root is something like src, tests, tools and in our case src=tf2onnx and all tflite code should go there.

guschmue (changes requested):
lgtm but we should move all of tflite code under tf2onnx

tf2onnx/tflite_handlers/__init__.py
TomWildenhain-Microsoft: When flatbuffer generates the interface scripts, they all import each other with lines like `from tflite.OperatorCode import OperatorCode` and will fail if tflite isn't in the root, but I can find/replace them to fix this.

guschmue (approved):

PR #1271, opened 2021-01-14
[LICENSE] add ALv2 licenses

szha:
Signed-off-by: Sheng Zha <zhasheng@amazon.com>


guschmue (approved):

szha:
Thanks for the quick review. It looks like some existing MIT licenses were missed that should be removed. I'm still looking into it

Issue #1277, opened 2021-01-19
Failed to apply optimize_transpose - converting faster_rcnn_inception_v2 to ONNX


vilmara:
**Describe the bug**
The process crashed even though the  script ends stating the model was successfully converted and saved as ONNX

**Urgency**
End of January

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Ubuntu 18.04
- Tensorflow Version: 1.15.2
- Python version: 3.6
-  Onnx=1.7.0
- tf2onnx=1.9.0/72fb20
- I`NFO - Using tensorflow=1.15.2, onnx=1.7.0, tf2onnx=1.9.0/72fb20`

**To Reproduce**
Model: 
faster_rcnn_inception_v2_coco_2018_01_28 downloaded from TensorFlow 1 Detection Model Zoo

Script used: 
```
python3 -m tf2onnx.convert --graphdef faster_rcnn_inception_v2_coco_2018_01_28/frozen_inference_graph.pb \
      --output model.onnx \
      --inputs image_tensor:0 \
      --outputs num_detections:0,detection_boxes:0,detection_scores:0,detection_classes:0 \
      --tag server \ 
      --fold_const \
     --opset 12
```

**Expected behavior**
graphdef converted to onnx but got a bunch on warnings and errors:
```
2021-01-19 01:42:14.297029: I tensorflow/stream_executor/platform/default/dso_loader.cc:44] Successfully opened dynamic library libcudart.so.10.2
WARNING:tensorflow:From /usr/local/lib/python3.6/dist-packages/tf2onnx/verbose_logging.py:76: The name tf.logging.set_verbosity is deprecated. Please use tf.compat.v1.logging.set_verbosity instead.

2021-01-19 01:42:15,433 - WARNING - From /usr/local/lib/python3.6/dist-packages/tf2onnx/verbose_logging.py:76: The name tf.logging.set_verbosity is deprecated. Please use tf.compat.v1.logging.set_verbosity instead.

2021-01-19 01:42:15.434277: I tensorflow/stream_executor/platform/default/dso_loader.cc:44] Successfully opened dynamic library libcuda.so.1
2021-01-19 01:42:15.469844: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1639] Found device 0 with properties:
name: Tesla T4 major: 7 minor: 5 memoryClockRate(GHz): 1.59
pciBusID: 0000:03:00.0
[...]

PR #1288, opened 2021-01-22
transpose optimizer wrongly removes a transpose

guschmue:
Signed-off-by: guschmue <guschmue@microsoft.com>

fix for https://github.com/onnx/tensorflow-onnx/issues/1277

transpose optimizer wrongly removes a transpose when trying to push to none existing subgraph.

### DIFF
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -462,7 +462,7 @@ class Graph(object):
         self._output_to_consumers = {}
         self._input_to_graph = {}
         self.shapes = {}
-        self.graph_name = graph_name or "tf2onnx"
+        self.graph_name = graph_name or utils.make_name("tf2onnx")
         self._is_subgraph = is_subgraph
         self.ta_reads = []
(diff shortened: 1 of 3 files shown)


tf2onnx/graph.py
```diff
@@ -462,7 +462,7 @@ def __init__(self, nodes, output_shapes=None, dtypes=None, target=None, opset=No
         self._output_to_consumers = {}
         self._input_to_graph = {}
         self.shapes = {}
-        self.graph_name = graph_name or "tf2onnx"
+        self.graph_name = graph_name or utils.make_name("tf2onnx")
```
TomWildenhain-Microsoft, line 465: Did you determine the cause of the identical names issue?

guschmue: yes, its the main graph that is set to 'tf2onnx' when writing the onnx file, which is actually what we want.

TomWildenhain-Microsoft (approved):

PR #1297, opened 2021-01-26
Enabled tflite support and tests

TomWildenhain-Microsoft:

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -18,12 +18,15 @@ from onnx import onnx_pb
 
 import tf2onnx
 import tf2onnx.onnx_opset  # pylint: disable=unused-import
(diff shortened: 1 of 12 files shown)


tf2onnx/tfonnx.py
```diff
@@ -18,12 +18,15 @@
 
 import tf2onnx
 import tf2onnx.onnx_opset  # pylint: disable=unused-import
+import tf2onnx.tflite_handlers  # pylint: disable=unused-import
```
guschmue, line 21: can we move the import of the tflite_handlers into 
if tflite
so we pull those only in for tflite models? From the code way below it looks like that should be possible.

TomWildenhain-Microsoft: Can do

guschmue (approved):

PR #1305, opened 2021-02-01
test_unsorted_segment_ops needs tf-2.3 or better

guschmue:
Signed-off-by: guschmue <guschmue@microsoft.com>


TomWildenhain-Microsoft (approved):
Alternatively, pin numpy 1.19

PR #1320, opened 2021-02-05
Fix bug in depthwise conv dilations with padding=same

TomWildenhain-Microsoft:

### DIFF
--- a/tests/backend_test_base.py
+++ b/tests/backend_test_base.py
@@ -177,10 +177,9 @@ class Tf2OnnxBackendTestBase(unittest.TestCase):
                 tf.import_graph_def(graph_def, name='')
                 graph_def = tf_optimize(list(feed_dict.keys()), outputs, graph_def, fold_constant=constant_fold)
 
-        if True or self.config.is_debug_mode:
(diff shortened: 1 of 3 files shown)


tests/backend_test_base.py
```diff
@@ -177,10 +177,9 @@ def freeze_and_run_tf(self, func, feed_dict, outputs, as_session, premade_placeh
                 tf.import_graph_def(graph_def, name='')
                 graph_def = tf_optimize(list(feed_dict.keys()), outputs, graph_def, fold_constant=constant_fold)
 
-        if True or self.config.is_debug_mode:
-            model_path = os.path.join(self.test_data_directory, self._testMethodName + "_after_tf_optimize.pb")
-            utils.save_protobuf(model_path, graph_def)
-            self.logger.debug("created file  %s", model_path)
+        model_path = os.path.join(self.test_data_directory, self._testMethodName + "_after_tf_optimize.pb")
```
TomWildenhain-Microsoft, line 180: Must have previously slide in a change with `if True or ...`, but I like this always enabled so I'll keep it since I don't think it has any UT perf impact.

guschmue (approved):

guschmue (approved):

PR #1322, opened 2021-02-06
Tom/tflite dequantize

TomWildenhain-Microsoft:


TomWildenhain-Microsoft:
NVM disregard this review request until after #1321 merges.  It contains changes from both.

tests/run_pretrained_models.py
```diff
@@ -190,6 +202,8 @@ def __init__(self, url, local, input_func, input_names, output_names,
         self.structured_outputs = structured_outputs  # Needed to determine output order for tf_function
         self.rtol = rtol
         self.atol = atol
+        self.ptol = ptol
```
TomWildenhain-Microsoft, line 205: The new models in this commit contain post-processing with an ArgMax op so it makes more sense to allow p percent of the output tensor entries to differ from the TF model rather than allowing for a large atol or rtol.  The converted model has a small number of differing entries but they are not necessarily close to to their corresponding tf entries.

tests/run_pretrained_models.yaml
```diff
@@ -454,3 +454,30 @@ ssd_mobilenet_v2_300_float_tflite:
...
+    - ArgMax
+
+deeplabv3_mnv2_ade20k_uint8_tflite:
+  tf_min_version: 2.1
+  disabled: false
+  url: https://github.com/mlcommons/mobile_models/raw/main/v0_7/tflite/deeplabv3_mnv2_ade20k_uint8.tflite
+  model: "deeplabv3_mnv2_ade20k_uint8.tflite"
+  model_type: tflite
+  input_get: get_ade20k_uint8
+  ptol: 1.0
```
TomWildenhain-Microsoft, line 478: This is a 1% error rate which is surprisingly high.  I don't know why it is so high for the quantizized model, probably because ORT doesn't have quantization implemented for some ops so the ORT result may be slower but actually more accurate than (but different from) the tf results.

guschmue (approved):

PR #1330, opened 2021-02-09
Optimize Transpose->Slice regardless of Slice's axes

mateusztabaka:

### DIFF
--- a/tests/test_optimizers.py
+++ b/tests/test_optimizers.py
@@ -8,9 +8,11 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import unittest
+import itertools
 import numpy as np
 from onnx import helper, numpy_helper, TensorProto, OperatorSetIdProto
 from parameterized import parameterized
+
 from backend_test_base import Tf2OnnxBackendTestBase
 from common import unittest_main, group_nodes_by_type, check_opset_min_version, check_opset_max_version, get_test_config
 from tf2onnx import utils, constants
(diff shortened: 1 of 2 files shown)


TomWildenhain-Microsoft:
Feel free to request a review from me when you are ready to have this merged.

mateusztabaka:
I'm ready, so I'd appreciate your comments on this (seems like I can't add reviewers from github GUI, so I guess I can mark my readiness in the comments section).

tests/test_optimizers.py
```diff
@@ -309,33 +311,84 @@ def test_transpose_dequantize_with_axis(self, shape, perm_input, perm_output):
...
+            model_proto = self.make_model(graph, producer_name="onnx-tests")
+            self.run_transpose_compare(["Z1"], {"X": np.random.randn(*input_shape).astype(np.float32)},
+                                       model_proto, remaining_transpose_num=0)
+
+    @parameterized.expand([
+        ([2, 3, 4, 5], [1, 2, 1, 2], [0, 2, 3, 1], [0, 3, 1, 2]),
+        ([2, 3, 4, 5, 6], [1, 2, 1, 2, 1], [0, 2, 3, 4, 1], [0, 4, 1, 2, 3]),
+    ])
+    @check_opset_min_version(10, "Slice in opset 10 can accept dynamic 'start' and 'ends'")
+    def test_transpose_slice_10(self, input_shape, slice_size, perm_input, perm_output):
```
TomWildenhain-Microsoft, line 358:
```suggestion
    def test_transpose_slice_int32(self, input_shape, slice_size, perm_input, perm_output):
```
Nice catch on the type issue.  I know the other tests just have a number but it is generally nice to have a better test description.

tests/test_optimizers.py
```diff
@@ -309,33 +311,84 @@ def test_transpose_dequantize_with_axis(self, shape, perm_input, perm_output):
...
-        axes = np.array(list(range(len(input_shape))), dtype=np.int64)
-        node1 = helper.make_node("Transpose", ["X"], ["Y"], perm=perm_input, name="trans_1")
-        node2 = helper.make_node("Slice", ["Y", "starts", "ends", "axes"], ["Z"], name="relu")
-        node3 = helper.make_node("Transpose", ["Z"], ["Z1"], perm=perm_output, name="trans_2")
+    @check_opset_max_version(9, "Slice in opset 9 and takes 'axes, 'start' and 'ends' as attributes")
+    def test_transpose_slice(self, input_shape, slice_size, perm_input, perm_output):
+        axes_combinations = []
+        axes = list(range(len(input_shape)))
+        for i in range(1, len(input_shape) + 1):
+            axes_combinations.extend(list(itertools.combinations(axes, i)))
```
TomWildenhain-Microsoft, line 322: Do we really need to test every single combination?  That's like 32 runs per test.  Glad to know they all pass, but I think now we can safely just add an arg for parameterized "axes" and fill in a few tests, get rid of the loop here.

TomWildenhain-Microsoft (changes requested):
Thanks for the contribution!  Some small changes with the tests.  Now that it's reviewed, please submit updates to a new commit instead of amending this one.

tests/test_optimizers.py
mateusztabaka: this test was primarly designed to test Slice from opset 10 (hence the name similarly to test_const_fold_unsqueeze_with_const_13 test case we have). After pretrained_model_test blew up, I decided to change the axes type to int32 in this test case.
Maybe this case should be named like test_transpose_slice_opset_10 - let me know what you think.

tests/test_optimizers.py
mateusztabaka: sure, I will add parametrized axes.

tests/test_optimizers.py
TomWildenhain-Microsoft: Ah, my mistake.  Then yes, test_transpose_slice_opset_10 is great.

TomWildenhain-Microsoft (approved):

PR #1331, opened 2021-02-10
Optimize Transpose->Pad when 'pads' input is not a constant

mateusztabaka:
When the second input is not a constant, let's shuffle it with Split followed by Concat.
There are examples of models, where this non-constant input gets constant folded anyway by a framework.
Even if that's not the case, Split+Concat of 8 (or 10) element tensor should be a good trade for Transpose pair.

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -676,7 +676,21 @@ class TransposeOptimizer(GraphOptimizerBase):
                 input1.set_tensor_value(new_pads)
                 input1.data_format = "NCHW"
             return self._switch_transpose_and_node(node, trans)
-        return False
+        # when the second input is not a constant, let's shuffle it with Split followed by Concat
+        # there are examples of models, where this non-constant input
+        # gets constant folded anyway by a framework.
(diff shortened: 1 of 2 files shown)


mateusztabaka:
@TomWildenhain-Microsoft this one is ready for review. Thanks.

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -676,7 +676,21 @@ def _pad_handler(self, trans, node):
                 input1.set_tensor_value(new_pads)
                 input1.data_format = "NCHW"
             return self._switch_transpose_and_node(node, trans)
-        return False
+        # when the second input is not a constant, let's shuffle it with Split followed by Concat
+        # there are examples of models, where this non-constant input
+        # gets constant folded anyway by a framework.
+        split = self._g.make_node("Split", inputs=[node.input[1]], attr={}, output_count=trans_rank * 2)
```
TomWildenhain-Microsoft, line 682: Nice.  When the pad is non-const, it is often constructed by concating some tensors together.  This optimization might introduce a concat -> split -> concat sequence that can be further optimized.  Still, it is worth it for removing a transpose and the concat -> split -> concat will be operating on very small tensors.  We might want to add a new optimizer for it in the future though.

TomWildenhain-Microsoft (approved):

PR #1380, opened 2021-03-05
Added rewriter for specific pattern in a 1P model

TomWildenhain-Microsoft:


guschmue (approved):

tf2onnx/rewriter/layer_normalization_rewriter.py
```diff
@@ -0,0 +1,122 @@
+# SPDX-License-Identifier: Apache-2.0
+
+
+"""
+tf2onnx.rewrite - Rewrites a specific pattern that appears in some models from the tf layer_norm contrib op
```
xadupre, line 5: We should have a more specific comment. This one does not say which optimization is done here.

TomWildenhain-Microsoft: Fair point.

PR #1386, opened 2021-03-08
Enhance BatchToSpaceND and SpaceToBatchND handlers

mateusztabaka:
- add explicit NHWC_TO_NCHW and NCHW_TO_NHWC transposes so they can be
  eliminated with its counterpart in TransposeOptimizer
- handle nonconst pads and crops


TomWildenhain-Microsoft:
Is this ready for review + merge?

mateusztabaka:
> Is this ready for review + merge?

Yes

tests/test_backend.py
```diff
@@ -3323,7 +3323,7 @@ def func(x, z):
 
     @check_opset_min_version(11, "SpaceToBatchND")
     def test_space_to_batchnd_non_const_7d(self):
-        x_type, y_type, z_type = np.int64, np.int64, np.int64
+        x_type, y_type, z_type = np.float32, np.int64, np.int64
```
TomWildenhain-Microsoft, line 3326: The black_shapes include `[2, 2]` in this case, so it appears the code is properly tested.

tests/test_backend.py
```diff
@@ -3304,7 +3304,7 @@ def func(input_x, block_size, pad):
 
     @check_opset_min_version(11, "BatchToSpaceND")
     def test_batch_to_spacend_non_const_7d(self):
-        x_type, y_type, z_type = np.int64, np.int64, np.int64
+        x_type, y_type, z_type = np.float32, np.int64, np.int64
```
TomWildenhain-Microsoft, line 3307: I believe we need `block_shape[0] == block_shape[1]` and `len(block_shape) == 2` to ensure the code you added to BatchToSpaceND is tested.  The block shapes here are only `[2, 3]`, and `[2]`.  Are you sure your code is being tested?

TomWildenhain-Microsoft (changes requested):
Thanks for the contribution.  I'm not entirely sure that your code is being tested.  Can you double-check?

tests/test_backend.py
mateusztabaka: good catch. Thanks.

PR #1402, opened 2021-03-17
Added conversion for WordpieceTokenizer

xadupre:
Signed-off-by: xavier dupré <xavier.dupre@gmail.com>

### DIFF
--- a/tf2onnx/custom_opsets/string_ops.py
+++ b/tf2onnx/custom_opsets/string_ops.py
@@ -1,10 +1,12 @@
 # SPDX-License-Identifier: Apache-2.0
 
 """ tf2onnx mapping functions for string ops using contrib ops domain. """
+import io
+import json
 import logging
 import numpy as np
 from onnx.onnx_pb import TensorProto
-
+from onnx.helper import make_attribute
 from tf2onnx import constants, handler
 from tf2onnx.handler import tf_op
 from tf2onnx import utils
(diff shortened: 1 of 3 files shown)


tf2onnx/graph.py
```diff
@@ -1343,6 +1343,12 @@ def find_output_consumers(self, output_name):
                 nodes.extend(g.find_output_consumers(output_name))
         return nodes
 
+    def remove_output(self, output_name):
```
TomWildenhain-Microsoft, line 1346: Looks like this is never used.

tf2onnx/custom_opsets/string_ops.py
```diff
@@ -168,3 +170,41 @@ def version_1(cls, ctx, node, **kwargs):
...
+        table = initialized_tables[ressource]
+        if isinstance(table, tuple):
+            table = table[0]
+        mapping = {}
+        for i, word in enumerate(table):
+            if isinstance(word, bytes):
+                word = word.decode('utf-8')
+            mapping[word] = i
+        st = io.StringIO()
+        json.dump(mapping, st, separators=(',', ':'))
```
TomWildenhain-Microsoft, line 206: I think you can use `json.dumps` instead if you want a string.

TomWildenhain-Microsoft (changes requested):
A few comments, otherwise looks good!

tf2onnx/graph.py
xadupre: removed

TomWildenhain-Microsoft (approved):

guschmue:
@xadupre , what this merged ?

PR #1409, opened 2021-03-17
Only load specified signatures when loading saved models in tf1

TomWildenhain-Microsoft:


tf2onnx/tf_loader.py
```diff
@@ -231,29 +231,32 @@ def from_checkpoint(model_path, input_names, output_names):
...
 
     if tag is None:
         tag = [tf.saved_model.tag_constants.SERVING]
         logger.warning(wrn_no_tag)
 
+    if not signature_names:
+        logger.warning(wrn_empty_sig)
+
     if tag == '':
-        tag = [[]]
```
TomWildenhain-Microsoft, line 245: TF1 complains that list is an unhashable type.  We actually want to specify an empty set, which is `set([])` not `set([[]])`

guschmue (approved):

PR #1433, opened 2021-03-30
Prepare to enable Transpose optimizer for transposes of any rank

TomWildenhain-Microsoft:


tests/test_optimizers.py
```diff
@@ -455,6 +455,38 @@ def test_transpose_max_input_non_const(self, input_shape1, input_shape2, perm_in
                                             "non_const": np.random.randn(*input_shape2).astype(np.float32)},
                                    model_proto, remaining_transpose_num=1)
 
+    @parameterized.expand([
+        ((2, 3, 4, 5), (2, 4, 5, 3), [0, 2, 3, 1], [0, 3, 1, 2]),
+        ((2, 3, 4, 5, 6), (2, 4, 5, 6, 3), [0, 2, 3, 4, 1], [0, 4, 1, 2, 3]),
+    ])
+    @check_opset_min_version(8, "Max in opset 10 supports broadcasting")
+    def test_transpose_max_no_cancel(self, input_shape1, input_shape2, perm_input, perm_output):
```
TomWildenhain-Microsoft, line 463: Add a test for when the number of remaining transposes is still 2 (transposes don't cancel).  Once arbitrary transposes can be pushed, termination is not guaranteed so the non-canceling case requires special care.

guschmue (approved):

PR #1450, opened 2021-04-09
Enable string tests in CI pipeline

TomWildenhain-Microsoft:

### DIFF
--- a/ci_build/azure_pipelines/templates/setup.yml
+++ b/ci_build/azure_pipelines/templates/setup.yml
@@ -25,6 +25,15 @@ steps:
       pip install --index-url https://test.pypi.org/simple/ ort-nightly
     fi
 
+    if [[ $CI_PLATFORM == "windows" ]] ;
+    then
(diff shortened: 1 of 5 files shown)


ci_build/azure_pipelines/templates/setup.yml
```diff
@@ -25,6 +25,15 @@ steps:
       pip install --index-url https://test.pypi.org/simple/ ort-nightly
     fi
 
+    if [[ $CI_PLATFORM == "windows" ]] ;
+    then
+      pip install -i https://test.pypi.org/simple/ onnxruntime-customops==0.0.1
```
xadupre, line 30: I would put the version number of ortcustomops at the top of the file. It will be easier to test multiple configurations.

TomWildenhain-Microsoft: Cool.  I'll do that when we have multiple versions of ort-customops that we want to test.

guschmue (approved):

PR #1463, opened 2021-04-20
Improve perf of SegmentSum

TomWildenhain-Microsoft:

### DIFF
--- a/tests/backend_test_base.py
+++ b/tests/backend_test_base.py
@@ -250,6 +250,8 @@ class Tf2OnnxBackendTestBase(unittest.TestCase):
                 return None
             return [d.dim_value if d.HasField('dim_value') else -1 for d in info.type.tensor_type.shape.dim]
         for info in model_shapes.graph.value_info:
+            if info.name == "":
+                continue
             onnx_shape = get_shape(info)
             tf2onnx_shape = graph.get_shape(info.name)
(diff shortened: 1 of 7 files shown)


tests/backend_test_base.py
```diff
@@ -250,6 +250,8 @@ def get_shape(info):
                 return None
             return [d.dim_value if d.HasField('dim_value') else -1 for d in info.type.tensor_type.shape.dim]
         for info in model_shapes.graph.value_info:
+            if info.name == "":
```
TomWildenhain-Microsoft, line 253: ONNX shouldn't be adding shape info for empty strings used to denote optional outputs, but it has a bug https://github.com/onnx/onnx/issues/3439

tests/test_cond.py
```diff
@@ -233,6 +233,7 @@ def case_graph():
 
     @check_tf_min_version("1.8", "shape inference for Reshape op screws up")
     @check_opset_min_version(9, "ConstantOfShape")
+    @allow_missing_shapes("ONNX shape inference still determines if/else shape for unknown reason")
```
TomWildenhain-Microsoft, line 236: This PR also includes fixes for the UT's breaking with onnx 1.9

guschmue (approved):

PR #1478, opened 2021-04-27
Fix bug that renamed subgraph i/o twice

TomWildenhain-Microsoft:

### DIFF
--- a/tf2onnx/tfonnx.py
+++ b/tf2onnx/tfonnx.py
@@ -511,8 +511,6 @@ def process_tf_graph(tf_graph, continue_on_error=False, verbose=False, target=No
         for func in ordered_func:
             f_inputs_names = [t.name for t in func.inputs]
             f_output_names = [t.name for t in func.outputs]
-            f_inputs_names = rename_tensors_in_list(f_inputs_names)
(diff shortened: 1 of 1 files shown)


tf2onnx/tfonnx.py
```diff
@@ -524,12 +522,13 @@ def rename_tensors_in_nodes(onnx_nodes):
...
     check_io(input_names, output_names, output_shapes)
 
-    rename_tensors_in_nodes(onnx_nodes)
-    input_names = rename_tensors_in_list(input_names)
-    output_names = rename_tensors_in_list(output_names)
-    output_shapes = rename_tensors_in_dict(output_shapes)
-    dtypes = rename_tensors_in_dict(dtypes)
-    inputs_as_nchw = rename_tensors_in_list(inputs_as_nchw)
+    if not is_subgraph:
+        rename_tensors_in_nodes(onnx_nodes)
```
TomWildenhain-Microsoft, line 526: We actually don't need to rename tensors in subgraphs at all.

guschmue (approved):
interesting that we did not notice this. 
maybe we need to find a nice tf-2 model for integration tests.

PR #1480, opened 2021-04-27
Reorder keras outputs if possible

TomWildenhain-Microsoft:


tf2onnx/convert.py
```diff
@@ -322,6 +322,11 @@ def from_keras(model, input_signature=None, opset=None, custom_ops=None, custom_
     initialized_tables = None
     tensors_to_rename = tensor_names_from_structed(concrete_func, input_names, output_names)
 
+    if model.output_names:
+        # model.output_names is an optional field of Keras models indicating output order. It is None if unused.
```
TomWildenhain-Microsoft, line 326: Interestingly for the huggingface models this is None.  But on some other models it is present.

guschmue (approved):

PR #1497, opened 2021-05-04
Add optimizer to push q and dq ops into place

TomWildenhain-Microsoft:

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -915,6 +915,13 @@ class TransposeOptimizer(GraphOptimizerBase):
 
     def _quantize_handler(self, trans, node):
         # Used for QuantizeLinear and DequantizeLinear
+        if node.type == "DequantizeLinear":
+            # Only push through if we will be able to push through consumers too.
+            cons = self._g.find_output_consumers(node.output[0])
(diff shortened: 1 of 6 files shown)


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -916,6 +916,13 @@ def _slice_handler(self, trans, node):
 
     def _quantize_handler(self, trans, node):
         # Used for QuantizeLinear and DequantizeLinear
+        if node.type == "DequantizeLinear":
+            # Only push through if we will be able to push through consumers too.
+            cons = self._g.find_output_consumers(node.output[0])
+            # If there is a false positive in the handler map, the q_dq and transpose optimizers might fight.
+            # Give up after 3 iterations. The q_dq optimizer should win so the dq hugs the op.
+            if not all(n.type in self._handler_map for n in cons) or self.opt_iteration >= 3:
```
xadupre, line 924: I wonder about the impact on the conversion time. Does it increase when this optimizer is used?

TomWildenhain-Microsoft: Good question.  I expect this condition to occur very rarely.  Should only be an issue when a DQ is above an op that the transpose optimizer thinks it can handle but can't.  3 iterations shouldn't be too bad.  I think that's actually a very common number of iterations for the optimizers.

guschmue (approved):

Issue #1501, opened 2021-05-05
Feature request: explicitly-defined padding in ConvTranspose


calvinmccarter-at-lightmatter:
Currently [(here in onnx_opset/nn.py)](https://github.com/onnx/tensorflow-onnx/blob/9c7b56c7321927666eb2c00d15e6db16a1b7de1f/tf2onnx/onnx_opset/nn.py#L440), whenever possible a ConvTranspose node is created using `output_shape`-based padding (which is dynamic) rather than explicitly-defined padding.  In my opinion, it would be preferable to set `output_shape=None` and instead compute the corresponding `pads` from this `output_shape`. It's a lot more straightforward to convert ONNX models to PyTorch models when they use `pads` rather than `output_shape`, since `pads` is only computable if you also have an input datum (as one does with ONNX). And in my opinion, it's also more intuitive when examining the resulting ONNX model (in say, Netron).

Alternatively, the current condition `if output_shape is not None` could be replaced with something like `if FLAG_NAME and output_shape is not None`.

Any thoughts before I start working on a PR?

TomWildenhain-Microsoft:
Sure, using pads instead of output shape when possible is fine.  If you make a PR, I'll review and merge it.

calvinmccarter-at-lightmatter:
@TomWildenhain-Microsoft -- just made a PR, thanks!

PR #1513, opened 2021-05-12
Explicit padding for const shape ConvTranspose

calvinmccarter-at-lightmatter:
This creates a ConvTranspose node with constant shape using explicitly-defined padding rather than output_shape-based padding  (issue #1501).

### DIFF
--- a/tf2onnx/onnx_opset/nn.py
+++ b/tf2onnx/onnx_opset/nn.py
@@ -430,8 +430,6 @@ class ConvTranspose:
         conv_kernel_shape(ctx, node, 1, spatial=spatial)
         input_shape = ctx.get_shape(node.input[2])
         input_batch_dim = input_shape[0]
-        print("\nstarting")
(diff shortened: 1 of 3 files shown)


TomWildenhain-Microsoft (approved):

tf2onnx/onnx_opset/nn.py
```diff
@@ -435,12 +435,12 @@ def version_1(cls, ctx, node, **kwargs):
             input_dims = input_shape[1:1+spatial]
         else:
             input_dims = input_shape[2:2+spatial]
+        input_dims_known = -1 not in input_dims
```
TomWildenhain-Microsoft, line 438: Unused variable.  Please remove this line or use it in line 456.

TomWildenhain-Microsoft (changes requested):

TomWildenhain-Microsoft (approved):

PR #1524, opened 2021-05-18
Fix Conv for unknown input rank

TomWildenhain-Microsoft:

### DIFF
--- a/tf2onnx/onnx_opset/nn.py
+++ b/tf2onnx/onnx_opset/nn.py
@@ -349,7 +349,7 @@ class ConvOp:
 
         # prefix with batch dim of [1] to satisfy rank requirements
         input_shape = ctx.get_shape(node.input[0])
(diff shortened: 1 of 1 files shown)


tf2onnx/onnx_opset/nn.py
```diff
@@ -349,7 +349,7 @@ def any_version(cls, opset, ctx, node, **kwargs):
 
         # prefix with batch dim of [1] to satisfy rank requirements
         input_shape = ctx.get_shape(node.input[0])
-        if len(input_shape) == spatial + 1:
+        if input_shape is not None and len(input_shape) == spatial + 1:
             gb = GraphBuilder(ctx)
```
TomWildenhain-Microsoft, line 353: Honestly I'm not sure why this condition is here and if it ever triggers.  The unit tests I tried don't run it.

guschmue (approved):

PR #1529, opened 2021-05-21
Made from_keras work for tf1

TomWildenhain-Microsoft:


guschmue (approved):

guschmue:
I think you need to limit it to tf-1.15.

PR #1538, opened 2021-05-26
add script to run optimizer on onnx files

guschmue:


TomWildenhain-Microsoft (approved):
At some point we might want to move this inside tf2onnx so it gets pip installed (something like tf2onnx.optimize(model)) but for now looks good.

guschmue:
yeap, agree.

PR #1542, opened 2021-05-28
Fix subtle bug in ConvTranspose

TomWildenhain-Microsoft:

### DIFF
--- a/tf2onnx/onnx_opset/nn.py
+++ b/tf2onnx/onnx_opset/nn.py
@@ -263,8 +263,9 @@ def parse_dims_attr(node, dims, spatial):
         if len(dims) != spatial:
             dims = dims[1:-1]
     else:
-        # We have (N, C, ...).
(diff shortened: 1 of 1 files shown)


tf2onnx/onnx_opset/nn.py
```diff
@@ -263,8 +263,9 @@ def parse_dims_attr(node, dims, spatial):
         if len(dims) != spatial:
             dims = dims[1:-1]
     else:
-        # We have (N, C, ...).
-        dims = dims[2:]
+        # We have (N, C, ...) or (...).
+        if len(dims) != spatial:
+            dims = dims[2:]
```
xadupre, line 268: Is len(dim) == spatial after this line?

TomWildenhain-Microsoft: Yes it should be

xadupre (approved):

PR #1551, opened 2021-06-07
Tom/ort 1.8

TomWildenhain-Microsoft:


guschmue (approved):

guschmue:
wired, why did it start to fail?

TomWildenhain-Microsoft:
> wired, why did it start to fail?

I'm not completely sure.  For some reason it is installing flatbuffers 2.0 and that's causing the failure.

guschmue:
onnxruntime installs flatbuffers without pinning. tf2onnx wants flatbuffers~=1.12 - I guess if flatbuffers-2.x is compatible to flatbuffers==1.12 it would explain it.

TomWildenhain-Microsoft:
Hmm, flatbuffers 2.x is compatible with flatbuffers>=1.12 but not flatbuffers~=1.12 so I'm not sure.

PR #1568, opened 2021-06-14
Fix nightly CI

TomWildenhain-Microsoft:


guschmue (approved):

guschmue:
I think you could also move to python-3.9 with tf-2.5

TomWildenhain-Microsoft:
> I think you could also move to python-3.9 with tf-2.5

Good idea.  Let's try that.

PR #1580, opened 2021-06-28
Adds scripts to benchmark tfhub models and check discrepencies

xadupre:


guschmue (approved):
Not sure if it belongs in examples, maybe consider moving it into a subdir under tests.
There is also tests/run_pretrained_models.py that does similar things ... you could for example add a source 'tfhub' there.

xadupre:
> Not sure if it belongs in examples, maybe consider moving it into a subdir under tests.
> There is also tests/run_pretrained_models.py that does similar things ... you could for example add a source 'tfhub' there.

So I rename example/benchmark into tests/tfhub?

guschmue:
yes, something like that. If you can make it fairly generic, run_pretrained_models.py would work good I think.

xadupre:
I hesitated. I wanted to create something people could copy paste and reuse. run_pretrained_models is very long.

guschmue:
I did the same for tests/huggingface.py but there some other team wanted to run it outside tf2onnx :(
With tfhub I fear it might not be generic enough ... sometimes one needs to do extra stuff to make it work which might not work well with run_pretrained_models.

guschmue (approved):

PR #1581, opened 2021-06-28
Add target for converting to channels last format

TomWildenhain-Microsoft:


tf2onnx/graph.py
```diff
@@ -1124,13 +1124,29 @@ def make_graph(self, doc, graph_name=None, external_tensor_storage=None):
         # create output_tensor_values
         output_tensor_values = self.make_onnx_graph_io(self.outputs)
 
+        tensor_value_info = []
+
+        for op in ops:
+            if op.domain in [constants.ONNX_DOMAIN, constants.AI_ONNX_ML_DOMAIN]:
+                continue
+            # We still don't 100% trust the accuracy of all the shapes in graph.py, but for custom ops they are
+            # almost certainly accurate and onnx has no other way of knowing them.
```
xadupre, line 1133: Maybe one example would help here. Typical case: Reshape operator, `Z = Reshape(X, Y)` where X and Y are both inputs. There is no way to infer the shape of Z.

tools/onnx-optimize.py
```diff
@@ -28,23 +29,32 @@ def get_args():
...
-    g = GraphUtil.create_graph_from_onnx_model(model_proto)
+    g = GraphUtil.create_graph_from_onnx_model(model_proto, target)
     return g, model_proto
 
 
 def main():
     args = get_args()
 
-    g, org_model_proto = load_graph(args.input)
+    g, org_model_proto = load_graph(args.input, args.target)
```
xadupre, line 52: Maybe it would be worth adding an example in the documentation? Is it available through all APIs?

guschmue: I think we could but right now onnx does not support this so it would be a little iffy.

guschmue (approved):

tf2onnx/graph.py
TomWildenhain-Microsoft: This case wasn't actually what I was referring to.  Having unknown dims added is fine, the bigger issue would be actually incorrect dims.  While manipulating the graph object I have in the past seen shape data become inaccurate (after it has been used).  We now have tests for this and ideally it should never happen but I'm not 100% sure all bugs have been fixed.

PR #1583, opened 2021-06-29
add opset-14, move ci pipeline to test opset-8 -> opset-14

guschmue:


TomWildenhain-Microsoft (approved):
Opset 14 also adds Trilu and HardSwish which we may want to add at some point.

guschmue:
Yes, sure.
I think Trilu and HardSwish might not be directly supported in tf, at least search results don't show results. Need to look a little closer.

PR #1590, opened 2021-07-01
Tom/keras2onnx application tests

TomWildenhain-Microsoft:


guschmue:
our root dir gets full with keras* ... could we move this to tests/ ?

TomWildenhain-Microsoft:
> our root dir gets full with keras* ... could we move this to tests/ ?

Sure.  Should I move `keras2onnx_tests` as well?

guschmue:
yes, that would be awesome

guschmue (approved):

PR #1599, opened 2021-07-07
Refactor tflite/tf logic out of tfonnx into tf_utils and tflite_utils

TomWildenhain-Microsoft:


TomWildenhain-Microsoft:
This change will make it easy to add tfjs support.

guschmue (approved):

PR #1631, opened 2021-07-29
Move check earlier to fix loop in transpose optimizer

TomWildenhain-Microsoft:

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -236,14 +236,14 @@ class TransposeOptimizer(GraphOptimizerBase):
         }
 
     def _handle_node_having_branches(self, trans, node):
-        if not self._should_push_transpose(trans, node):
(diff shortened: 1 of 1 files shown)


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -236,14 +236,14 @@ def _initialize_handlers(self):
         }
 
     def _handle_node_having_branches(self, trans, node):
-        if not self._should_push_transpose(trans, node):
+        if not self._should_push_transpose(trans, node) or len(node.output) != 1:
```
TomWildenhain-Microsoft, line 239: This check needs to be moved before the transpose pair creation.  Otherwise, pairs can be created unnecessarily which need to be removed.  The removal actually causes an infinite loop in the optimizer.  Technically some of the other checks need to be moved too, or the pairs destroyed on failure.

guschmue (approved):

PR #1637, opened 2021-08-02
Fix tfhub scripts (enformer, ...)

xadupre:


TomWildenhain-Microsoft:
The unit tests were broken by a new onnx release.  I'm fixing it here #1638

guschmue (approved):

PR #1638, opened 2021-08-02
Skip onnx checker for new onnx version for IR3 bug

TomWildenhain-Microsoft:

### DIFF
--- a/tests/backend_test_base.py
+++ b/tests/backend_test_base.py
@@ -9,6 +9,7 @@
 import logging
 import os
 import unittest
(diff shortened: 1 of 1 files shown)


guschmue (approved):

guschmue:
I guess alternatively you could also run the checker with False for opsets < 8

TomWildenhain-Microsoft:
Yeah that would work too.  Hopefully we can remove it entirely when it is fixed in the next ONNX version.

PR #1657, opened 2021-08-09
Implement reduce all/any for non-const axes

TomWildenhain-Microsoft:

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -3557,14 +3557,14 @@ class BackendTests(Tf2OnnxBackendTestBase):
         self._run_test_case(func, [_OUTPUT], {_INPUT: input_val})
 
     def test_reduce_all(self):
-        input_val = np.random.randint(0, 2, (10, 20)).astype(np.bool)
+        input_val = np.random.randint(0, 2, (2, 20)).astype(np.bool)
         def func(x):
             res = tf.reduce_all(input_tensor=x, keepdims=False)
             res1 = tf.reduce_all(input_tensor=x, axis=[0], keepdims=False)
             return tf.identity(res, name=_TFOUTPUT), tf.identity(res1, name=_TFOUTPUT1)
         self._run_test_case(func, [_OUTPUT, _OUTPUT1], {_INPUT: input_val})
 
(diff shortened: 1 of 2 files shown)


tests/test_backend.py
```diff
@@ -3557,29 +3557,29 @@ def func(x):
         self._run_test_case(func, [_OUTPUT], {_INPUT: input_val})
 
     def test_reduce_all(self):
-        input_val = np.random.randint(0, 2, (10, 20)).astype(np.bool)
+        input_val = np.random.randint(0, 2, (2, 20)).astype(np.bool)
```
TomWildenhain-Microsoft, line 3560: 10 isn't great since there's a only a 1/1024 chance of them all matching.  20 sets of size 2 is good.

tf2onnx/onnx_opset/reduction.py
```diff
@@ -144,6 +144,29 @@ def version_6(cls, ctx, node, **kwargs):
         ctx.make_node(op_type="Greater", inputs=[reduce_node_output, zero_node.output[0]],
                       name=node.name, outputs=node.output, shapes=shapes, dtypes=dtypes)
 
+    @classmethod
+    def version_13(cls, ctx, node, **kwargs):
+        keepdims = node.get_attr_value('keep_dims')
+        reduce_input = node.input[0]
+        if node.type == "All":
+            reduce_input = ctx.make_node("Not", [reduce_input]).output[0]
+        cast = ctx.make_node("Cast", inputs=[reduce_input], attr={"to": onnx_pb.TensorProto.FLOAT}).output[0]
```
xadupre, line 153: I was wondering why it is necessary to cast into float?

TomWildenhain-Microsoft: Hmm, good question.  I copied over the existing code and just modified it to do dynamic axes.  I think one of the previous ops maybe didn't support non-float values.  In any case, float is technically good because it doesn't roll over and is guaranteed to preserve monotonicity of addition.  Since we just want to detect if any of the values are non-zero, this is good.

guschmue (approved):

PR #1722, opened 2021-09-15
Create end2end_tfkeras_serialize.py

microprediction:
When I read the Keras example I was initially unsure how to avoid the use of subprocess/saving the keras model to file. This is a minor modification to the existing example as I suspect some users would prefer not to go via disk, or may not have entitlements to do so.


examples/end2end_tfkeras_serialize.py
```diff
@@ -0,0 +1,63 @@
...
+
+"""
+This example builds a simple model without training.
+It is converted into ONNX directly, by serializing to byte string.
+Predictions are compared to the predictions from tensorflow to check there is no
+discrepencies. Inferencing time is also compared between
+*onnxruntime*, *tensorflow* and *tensorflow.lite*.
+"""
+
+# from https://github.com/microprediction/tensorflow-onnx/blob/master/examples/end2end_tfkeras.py
```
TomWildenhain-Microsoft, line 11: I don't think we want this reference to an external repo.

TomWildenhain-Microsoft (changes requested):
Hi, thanks for the contribution. I agree that having an example using `from_keras` would be great. I'd favor renaming the existing tutorial to be called `end2end_tf_saved_model` and edit `end2end_tf_keras` accordingly.

microprediction:
A renaming & cut and pate error, but hopefully the end result is okay. 
I added a few lines for the M1 folks.

examples/end2end_tfkeras_serialize.py
microprediction: shall delete

guschmue:
inactive - closing

PR #1725, opened 2021-09-20
Try both zipped and unzipped formats for tfjs weights

TomWildenhain-Microsoft:


TomWildenhain-Microsoft:
NVM don't merge.  Have to confirm something first.

PR #1727, opened 2021-09-23
Transpose optimization for Softmax and LogSoftmax (fixes #1716)

### LINKED ISSUE #1716: Feature: Transpose optimization for Softmax
Thanks to the optimizations by tf2onnx, most Transpose operations in converted models are eliminated.
However, Softmax does not seem to be accounted for yet.

Take the following simple SavedModel:
```py
import tensorflow as tf
from keras import Model
from keras.layers import Input, Softmax
import keras.backend as K
K.set_image_data_format('channels_first')

inputs = Input((1, None, None), name='input')
outputs = Softmax(axis=1, name='output')(inputs)
m = Model(inputs, outputs)

tf.saved_model.save(
    obj=m,
    export_dir='test_softmax'
)
```

When exporting it via `python -m tf2onnx.convert --saved-model test_softmax --output test_softmax/model.onnx --tag serve --signature_def serving_default --opset 14`, the result looks like this:

Same for every other channels_first model I export - the Softmax operation is always surrounded by Transposes.

fthielke:
In lower opsets, Softmax always coerces its inputs to a 2D tensor, making Transpose operations necessary if the permutation moves axes between the coerced batch and feature dimensions.
While one could find and optimize away Transposes that only permute axes in a way that keeps the batch/feature split, I would not consider that a common use case; optimizing only for opset >= 13 seems good enough for now.


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -822,6 +823,13 @@ def permute_pads(pads):
 
     def _prelu_handler(self, trans, node):
         return self._handle_node_having_branches(trans, node)
+    
+    def _softmax_handler(self, trans, node):
+        # Softmax only operates on an arbitrary axis since opset 13
+        if self._g.opset >= 13:
+            return self._arg_min_max_handler(trans, node)
```
TomWildenhain-Microsoft, line 830: Good to case on the opset.  I think all opsets should be able to be optimized, but the axis attribute has a different default value (1 in opsets 1 and 11, -1 in opsets 13 onward).

https://github.com/onnx/onnx/blob/master/docs/Changelog.md#Softmax-1

arg_min_max handler probably isn't the right thing to use.  Combine simple_through_handler with an update to the axis attribute.  The output shape of Softmax equals the input shape so you don't need the fancy shape updating from _arg_min_max_handler (inherited from _reduce_handler)

tests/test_optimizers.py
```diff
@@ -1302,6 +1302,24 @@ def test_transpose_argmax(self):
         self.run_transpose_compare(["res"], {"X": np.random.randn(*input_shape).astype(np.float32)},
                                    model_proto, remaining_transpose_num=0)
 
+    @check_opset_min_version(13, "Softmax can only be optimized since opset 13")
+    def test_transpose_softmax(self):
```
TomWildenhain-Microsoft, line 1306: Thanks for adding a test along with your change.

TomWildenhain-Microsoft (changes requested):
Really appreciate the PR.  Fix these issues and hopefully the tests will pass.

tf2onnx/optimizer/transpose_optimizer.py
TomWildenhain-Microsoft: Ah I see why you have the `self._g.opset >= 13` check.  The 2D coercion will be an issue if the transpose permutation crosses the coercion boundary.  If it doesn't though, you can handle pre-opset 13 cases by **not** updating the axis and using the simple through handler.

TomWildenhain-Microsoft:
I'm actually working on a comprehensive transpose optimizer for ORT currently. It might be use din tf2onnx at some point. Here is the logic for Softmax optimization: https://github.com/microsoft/onnxruntime/blob/27ec2cd488d516a39abb41473d2284e0bc9732dc/onnxruntime/core/optimizer/transpose_optimizer/transpose_optimizer.cc#L709

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -822,6 +824,28 @@ def permute_pads(pads):
...
+    
+    def _softmax_handler(self, trans, node):
+        trans_rank = get_transpose_rank(trans)
+        perm = trans.get_attr("perm").ints
+
+        if self._g.opset >= 13:
+            # Softmax operates on an arbitrary axis since opset 13
+            axis = node.get_attr_value("axis", -1)
+            new_axis = perm[axis + trans_rank if axis < 0 else axis]
+            if not self._switch_transpose_and_node(node, trans):
```
fatcat-z, line 836:
```suggestion
            if not self._simple_through_handler(trans, node):
```

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -822,6 +824,28 @@ def permute_pads(pads):
...
+            return True
+        else:
+            # For older opsets, the "axis" attribute determines the coercion point for coercing the input tensor to 2D.
+            # We can safely switch transpose and node if the permutation does not make any axes cross that boundary.
+            coercion_axis = node.get_attr_value("axis", 1)
+            for from_axis, to_axis in enumerate(perm):
+                if (from_axis < coercion_axis and to_axis >= coercion_axis) or (from_axis >= coercion_axis and to_axis < coercion_axis):
+                    return False
+            
+            return self._switch_transpose_and_node(node, trans)
```
fatcat-z, line 848:
```suggestion
            return self._simple_through_handler(trans, node)
```

tf2onnx/optimizer/transpose_optimizer.py
TomWildenhain-Microsoft: I think those are identical

TomWildenhain-Microsoft:
Some pylint style failures:
```
************* Module tf2onnx.optimizer.transpose_optimizer
tf2onnx/optimizer/transpose_optimizer.py:827:0: C0303: Trailing whitespace (trailing-whitespace)
tf2onnx/optimizer/transpose_optimizer.py:845:0: C0301: Line too long (136/120) (line-too-long)
tf2onnx/optimizer/transpose_optimizer.py:847:0: C0303: Trailing whitespace (trailing-whitespace)
tf2onnx/optimizer/transpose_optimizer.py:832:8: R1705: Unnecessary "else" after "return" (no-else-return)
tf2onnx/optimizer/transpose_optimizer.py:845:20: R1716: Simplify chained comparison between the operands (chained-comparison)
tf2onnx/optimizer/transpose_optimizer.py:845:80: R1716: Simplify chained comparison between the operands (chained-comparison)
************* Module test_optimizers
tests/test_optimizers.py:1305:0: C0301: Line too long (141/120) (line-too-long)
tests/test_optimizers.py:1323:0: C0301: Line too long (141/120) (line-too-long)
tests/test_optimizers.py:1359:0: C0301: Line too long (144/120) (line-too-long)
tests/test_optimizers.py:1377:0: C0301: Line too long (144/120) (line-too-long)
```
Otherwise, looks good to me.

tf2onnx/optimizer/transpose_optimizer.py
fatcat-z: Yes, the 2 function is identical but _simple_through_handler seems like a wrapper for such case. Just like we call this method for some simple ops instead of calling _switch_transpose_and_node().

fthielke: _simple_through_handler seems to only be used in _handler_map, the other handler functions always use _switch_transpose_and_node.

Thus I would prefer to leave this as is.

fatcat-z:
@fthielke , just a kindly reminder to see if you might have a chance to update this PR for a merge.

fatcat-z:
Closing this one since https://github.com/onnx/tensorflow-onnx/pull/1964 has been merged with latest changes.

PR #1732, opened 2021-10-01
Rename end2end_tfkeras.py to end2end_tf_keras.py

microprediction:


TomWildenhain-Microsoft:
You need to commit with the -s flag to make the DCO bot happy.  The other unit test failures are because onnxruntime-extensions changed/broke something.

hwangdeyu:
> You need to commit with the -s flag to make the DCO bot happy. The other unit test failures are because onnxruntime-extensions changed/broke something.

Hi @TomWildenhain-Microsoft , does it's a necessary change? Seems https://github.com/onnx/tensorflow-onnx/blob/master/examples/end2end_tfhub.py named the same way with `end2end_tfkeras.py` file.

garymm:
I suggest renaming the files to:

* end_to_end_keras.py
* end_to_end_tf_hub.py

As much as possible, separate words with underscores, and omit unnecessary words.

guschmue:
inactive - closing

Issue #1757, opened 2021-10-29
Bert2Bert concat axis dimension error at serving time


LoicDagnas:
**Describe the bug**
Converting a Bert2Bert model from [TensorFlow model official](https://github.com/tensorflow/models/blob/master/official/nlp/nhnet/models.py), I get the following errors at serving time:

```
Traceback (most recent call last):
  File "C:/dev/ml/QueryGenerator/query_generator/models/bert2bert/save_model.py", line 245, in <module>
    output = session.run(output_names=None, input_feed=input_feed)
  File "C:\dev\ml\QueryGenerator\venv\lib\site-packages\onnxruntime\capi\onnxruntime_inference_collection.py", line 188, in run
    return self._sess.run(output_names, input_feed, run_options)
onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : Non-zero status code returned while running Loop node. Name:'bert2_bert/while_loop' Status Message: Non-zero status code returned while running Concat node. Name:'bert2_bert/while/decoder/decoder/layer_0/self_attention/concat' Status Message: concat.cc:159 onnxruntime::ConcatBase::PrepareForCompute Non concat axis dimensions must match: Axis 0 has mismatched dimensions of 10 and 6
```

It is quite close to the error I had at the end of [this issue](https://github.com/onnx/tensorflow-onnx/issues/1634), but:
- the model is different
- the minimal code is simpler
- the error is different

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Windows 10.0.19042
- Tensorflow Version: 2.5.0
- Python version: 3.7.6

**To Reproduce**
It is quite easy to reproduce using the following code:

```python
 import tensorflow as tf
 import onnxruntime
 import tf2onnx
 from official.nlp.nhnet.configs import UNITTEST_CONFIG, BERT2BERTConfig
 from official.nlp.nhnet.models import Bert2Bert, get_bert2bert_layers
 
 MAX_SEQ_LENGTH = 10
 MAX_OUTPUT_LENGTH = 4
 
 # Create the Bert2Bert model
 bert2bert_config_dict = UNITTEST_CONFIG.copy()
 bert2bert_config_dict["max_position_embeddings"] = MAX_SEQ_LENGTH
 bert2bert_config_dict["len_title"] = MAX_OUTPUT_LENGTH
[...]

PR #1771, opened 2021-11-15
Add support for one hot encoding of 0-rank tensor

LoicDagnas:
Trying to solve my self this issue https://github.com/onnx/tensorflow-onnx/issues/1757 I have been lead to add the support of one hot encoding of 0-rank tensor as it is supported by TensorFlow.


TomWildenhain-Microsoft (approved):

q-ycong-p:
This commit caused `python setup.py test` to throw segmentation fault on the newly added `test_onehot_rank0` with `onnx==1.7.0`. Confirmed that latest `onnx==1.10.2` is fine. Can we investigate if it's an easy fix, or only latest onnx is supported?

PR #1699, opened 2021-09-02
Handle Abs in TransposeOptimizer

mateusztabaka:
Signed-off-by: Mateusz Tabaka <mateusz.tabaka@intel.com>

PR #1785, opened 2021-11-23
Handle 'Neg' nodes in TransposeOptimizer

jan-haug:
`Neg` nodes are also element-wise operations, so we can handle them like similar nodes, e.g. `Abs` (see #1699).


hwangdeyu (approved):
Looks great to me. Any other comments? @TomWildenhain-Microsoft  @fatcat-z

TomWildenhain-Microsoft (approved):

Issue #1792, opened 2021-12-01
Different conv2d with bias behavior between python script and test_backend.py


hwangdeyu:
**Describe the bug**
Different conv2d with bias behavior between python script and [test_backend.py](https://github.com/onnx/tensorflow-onnx/blob/master/tests/test_backend.py)

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 18.04):
- Tensorflow Version: 1.15.4
- Python version: Python 3.8

**To Reproduce**
- Python script
```python
import onnx
import tf2onnx
import tensorflow as tf

inputs = tf.keras.Input(shape=(64, 256, 256))
middles = tf.keras.layers.ZeroPadding2D(
    padding=(0, 4),
    data_format="channels_first",
    name="padding")(inputs)
outputs = tf.keras.layers.Conv2D(
    filters=1,
    kernel_size=(9, 9),
    strides=(1, 1),
    use_bias=True,
    data_format="channels_first",
    name="conv2d",
)(middles)
model = tf.keras.Model(inputs=inputs, outputs=outputs)

onnx_model, _ = tf2onnx.convert.from_keras(
    model=model,
    opset=13,
    large_model=False,
)
onnx.save_model(onnx_model, "conv2d.onnx")
```

- test_backend test
```python
    def test_conv2d_biasadd_rewriter(self):
        x_shape = [1, 64, 256, 256]
        x_val = make_xval(x_shape)
        def func(x):
            middles = tf.keras.layers.ZeroPadding2D(
                padding=(0, 4),
                data_format="channels_first",
                name="padding")(x)
            t = tf.keras.layers.Conv2D(
                filters=1,
                kernel_size=(9, 9),
                strides=(1, 1),
                use_bias=True,
                data_format="channels_first",
                name="conv2d",
            )(middles)
            return tf.identity(t, name=_TFOUTPUT)
        self._run_test_case(func, [_OUTPUT], {_INPUT: x_val}, rtol=1e-04, atol=1e-2, as_session=True,
                                graph_validator=lambda g: check_op_count(g, "Add", 0, disabled=False))
```

**Screenshots tf graph**
Different node.
- Python script result
![image](https://user-images.githubusercontent.com/10047193/144208530-75b0c09a-4d7a-45e3-8338-cc3c3d9476f5.png)

- test_backend test result
[...]

PR #1794, opened 2021-12-02
Enable constfold variable in optimize

hwangdeyu:
fixes https://github.com/onnx/tensorflow-onnx/issues/1792

### DIFF
--- a/tf2onnx/tf_loader.py
+++ b/tf2onnx/tf_loader.py
@@ -684,8 +684,10 @@ def tf_optimize_grappler(input_names, output_names, graph_def, fold_constant=Non
     #   depends on so for now don't turn this on.
     rewrite_options.optimizers[:] = [
         # 'pruning', 'constfold', 'arithmetic', 'dependency', 'function',
-        'constfold', 'function'
+        'function'
     ]
+    if fold_constant:
+        rewrite_options.optimizers.append("constfold")
     meta_graph = tf.compat.v1.train.export_meta_graph(graph_def=graph_def)
     fetch_collection = meta_graph_pb2.CollectionDef()
     for t in input_names + output_names:


hwangdeyu:
Hi @guschmue , 
I found the `constfold` is no used in tf_optimize() after this change https://github.com/onnx/tensorflow-onnx/commit/5c11ecee6f619dfc60fef956a76c31fe90249037. 

And I find some unit test cases is not needed the const fold. Could you help review this change, thanks!

guschmue:
Oh, this is a pretty old change. The rational was that we don't want to ever run with fold_constant=False since it have a pretty hard dependency on constant folding. 
We want to get rid of fold_constant completely but because tf_optimize() was part of the old api we did not want to break it and just ignore fold_constant.

Basically without fold_constant we can hit all kinds of issues - not worth it.

hwangdeyu:
> Oh, this is a pretty old change. The rational was that we don't want to ever run with fold_constant=False since it have a pretty hard dependency on constant folding. We want to get rid of fold_constant completely but because tf_optimize() was part of the old api we did not want to break it and just ignore fold_constant.
> 
> Basically without fold_constant we can hit all kinds of issues - not worth the trouble.
> 
> If this is related to #1792 - the graph that comes out of the unit test is the one we want since the BiasAdd is all 0 (we don't want the back to back Identity - should have been optimized unless both are in the output list) so the question is more why we did not optimize the BiasAdd from the python script - should have taken the same code path.

I get it. We always need `fold_constant` in `tf_optimize()`.
However, the python script runs the different path to ONNX. It deals with the constat folding in  https://github.com/onnx/tensorflow-onnx/blob/72d64606d9aff4f93bb064ba901af49377266150/tf2onnx/tfonnx.py#L468

And calls `fold_constants_using_tf()` instead unit test `rewrite_options.optimizers` to do constfold.
Is it necessary to add the `rewrite_options.optimizers` it in python script path too?

tf2onnx/tf_loader.py
```diff
@@ -684,8 +684,10 @@ def tf_optimize_grappler(input_names, output_names, graph_def, fold_constant=Non
     #   depends on so for now don't turn this on.
     rewrite_options.optimizers[:] = [
         # 'pruning', 'constfold', 'arithmetic', 'dependency', 'function',
-        'constfold', 'function'
+        'function'
     ]
+    if fold_constant:
```
hwangdeyu, line 689: Old version api. `fold_constant` should always be opened.

Issue #1741, opened 2021-10-13
TF1-Keras ZeroPadding2D + Conv2D Bias Term Not Fused.


leimao:
**Describe the bug**

ZeroPadding2D + Conv2D with bias term cannot be exported as single Conv2D ONNX operator. It is exported as  Conv2D + Add instead.

**Urgency**

We wish it could be fixed asap.

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): 20.04.2 LTS
- Tensorflow Version: 1.15.5
- Python version: 3.8.10

**To Reproduce**

In NVIDIA TensorFlow-1 Docker Container:

```
docker run -it --rm --gpus all -v $(pwd):/mnt nvcr.io/nvidia/tensorflow:21.08-tf1-py3
```

Install TF2ONNX (tried both the master branch and the latest release 1.9.2)

```
pip install git+https://github.com/onnx/tensorflow-onnx
```

Run the following script to generate ONNX file for Conv2DTranspose.

```python
import onnx
import tf2onnx
import tensorflow as tf

inputs = tf.keras.Input(shape=(64, 256, 256))
middles = tf.keras.layers.ZeroPadding2D(
    padding=(0, 4),
    data_format="channels_first",
    name="padding")(inputs)
outputs = tf.keras.layers.Conv2D(
    filters=1,
    kernel_size=(9, 9),
    strides=(1, 1),
    use_bias=True,
    data_format="channels_first",
    name="conv2d",
)(middles)
model = tf.keras.Model(inputs=inputs, outputs=outputs)

onnx_model, _ = tf2onnx.convert.from_keras(
    model=model,
    opset=13,
    large_model=False,
)

onnx.save_model(onnx_model, "conv2d.onnx")
```

**Screenshots**

![image](https://user-images.githubusercontent.com/17606112/137186770-dd287163-6bf8-4d3a-9fb5-c00ffce20f2c.png)

**Additional context**

N/A

PR #1796, opened 2021-12-06
Fix keras Conv2D BiasAdd fuse

hwangdeyu:
fixes https://github.com/onnx/tensorflow-onnx/issues/1741.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -2376,7 +2376,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
             return tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case(func, [_OUTPUT], {_INPUT: x_val})
 
-    @skip_tflite("skip_tflite is not support unit32 when tf version <= 2.3.0")
+    @skip_tflite("tflite does not support uint32 if tf version <= 2.3.0")
     @check_opset_min_version(6, "cast")
     def test_cast_unit32(self):
         x_val = np.array([1, 2, 3, 4], dtype=np.uint32).reshape((2, 2))
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -2376,6 +2376,9 @@ class BackendTests(Tf2OnnxBackendTestBase):
             return tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case(func, [_OUTPUT], {_INPUT: x_val})
 
+    @skip_tflite("skip_tflite is not support unit32 when tf version <= 2.3.0")
+    @check_opset_min_version(6, "cast")
+    def test_cast_unit32(self):
         x_val = np.array([1, 2, 3, 4], dtype=np.uint32).reshape((2, 2))
         def func(x):
             x_ = tf.cast(x, tf.uint64)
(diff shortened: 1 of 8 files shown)


tf2onnx/rewriter/__init__.py
```diff
@@ -49,6 +48,7 @@
     "rewrite_quantize_and_dequantize",
     "rewrite_layer_normalization",
     "rewrite_conv_dilations",
+    "rewrite_conv2d_with_pad",
```
hwangdeyu, line 51: `rewrite_biasadd_with_conv2d` fuses change the tf op `Conv2D` to ONNX op 'Conv'.
So `rewrite_conv2d_with_pad` should be set after `rewrite_biasadd_with_conv2d`.

tf2onnx/convert.py
```diff
@@ -373,6 +373,10 @@ def _from_keras_tf1(model, input_signature=None, opset=None, custom_ops=None, cu
 
     with tf.device("/cpu:0"):
         frozen_graph, initialized_tables = tf_loader.freeze_session(sess, input_names, output_names, get_tables=True)
+        tf_loader.tf_reset_default_graph()
+        tf.import_graph_def(frozen_graph, name='')
+        sess = tf.keras.backend.get_session(model.outputs)
+        frozen_graph = tf_loader.tf_optimize(input_names, output_names, frozen_graph)
```
hwangdeyu, line 379: fixes https://github.com/onnx/tensorflow-onnx/issues/1792.

hwangdeyu:
Hi @guschmue, 
after adding the `tf_optimize()` in `_from_keras_tf1', the optimize() may make it model change. So all [the keras test](https://dev.azure.com/tensorflow-onnx/tensorflow-onnx/_build/results?buildId=16516&view=logs&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=841356e0-85bb-57d8-dbbc-852e683d1642) models using model predict are failed. 
Do you have any idea to fix this cases?

guschmue:
tf_optimize() sure will change the graph but should not change numerical results. We call it for every path (just not the one in this issue) - must be some problem how we call it in this case.

hwangdeyu:
> tf_optimize() sure will change the graph but should not change numerical results. We call it for every path (just not the one in this issue) - must be some problem how we call it in this case.

Yeah, there may have been a misunderstanding in my comment before.

I mean that after adding the optimize change code, some tests are failed in [ `model.predict test_api.py#L68` ](https://github.com/hwangdeyu/tensorflow-onnx/blob/23b67120e510842da4658ebcae90856b8c49dcd6/tests/test_api.py#L68) and throw the error:
```
tensorflow.python.framework.errors_impl.NotFoundError: PruneForTargets: Some target nodes not found: predict/group_deps 
```
But [test_api.py #L65 L66](https://github.com/hwangdeyu/tensorflow-onnx/blob/23b67120e510842da4658ebcae90856b8c49dcd6/tests/test_api.py#L65) can be passed. So the onnx graph is ok and onnxruntime can be run successfully and get the same numerical results with original keras model.
Now I'm tring to figure out how the `tf_optimize()` change the graph and make model predict failed.

tests/test_backend.py
```diff
@@ -2353,6 +2376,9 @@ def func(x):
             return tf.identity(x_, name=_TFOUTPUT)
         self._run_test_case(func, [_OUTPUT], {_INPUT: x_val})
 
+    @skip_tflite("skip_tflite is not support unit32 when tf version <= 2.3.0")
```
fatcat-z, line 2379: ... does not support unit32 if tf version <= 2.3.0.

hwangdeyu: Fixed!

PR #1797, opened 2021-12-14
Fix tfjs CI failure and add tfjs convert log

hwangdeyu:


tests/backend_test_base.py
```diff
@@ -203,15 +203,18 @@ def convert_to_tfjs(self, graph_def_path, output_names):
         try:
             from tensorflowjs.converters import converter
         except ImportError:
+            self.logger.debug("Tensorflowjs.converters package import failed.")
```
fatcat-z, line 206: We need a warning for such case.

fatcat-z (approved):

PR #1806, opened 2021-12-20
Skip the blob failed model tests

hwangdeyu:
Co-authored-by: fatcat-z <jiz@microsoft.com>

### DIFF
--- a/tests/keras2onnx_applications/nightly_build/test_transformers.py
+++ b/tests/keras2onnx_applications/nightly_build/test_transformers.py
@@ -12,6 +12,7 @@ from mock_keras2onnx.proto import keras
 import numpy as np
 import tensorflow as tf
 from onnxconverter_common.onnx_ex import get_maximum_opset_supported
+from test_utils import check_bloburl_access
 
 sys.path.insert(0, os.path.join(dirname(abspath(__file__)), '../../keras2onnx_tests/'))
 from test_utils import run_onnx_runtime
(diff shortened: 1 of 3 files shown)


tests/keras2onnx_applications/nightly_build/test_transformers.py
```diff
@@ -21,9 +21,20 @@
 if os.environ.get('ENABLE_FULL_TRANSFORMER_TEST', '0') != '0':
     enable_transformer_test = True
 
+def check_bloburl_access():
```
fatcat-z, line 24: I think we can add a common.py to save such common methods and avoid having duplicated ones in different tests files.

tests/keras2onnx_applications/nightly_build/test_transformers.py
```diff
@@ -21,9 +21,20 @@
...
+    url = r'https://lotus.blob.core.windows.net/converter-models/transformer_tokenizer/'
+    try:
+        response = urllib.request.urlopen(url)
+        if response.getcode() != 200:
```
fatcat-z, line 28: This could be replaced with:

return response.getcode() == 200

tests/keras2onnx_applications/nightly_build/test_transformers.py
hwangdeyu: Good point.

fatcat-z (approved):

PR #1839, opened 2022-02-08
feat: For QDQ models, disable constant folding and ensure the conv weights are transposed without using an explicit op

peri044:
For QDQ models, disable constant folding and ensure the conv weights are transposed without using an explicit op


srihari-humbarwadi (commented):
LGTM

tf2onnx/tf_loader.py
```diff
@@ -684,8 +684,11 @@ def tf_optimize_grappler(input_names, output_names, graph_def, fold_constant=Non
     # depends on so for now don't turn this on, fold_constant is always enabled now.
     rewrite_options.optimizers[:] = [
         # 'pruning', 'constfold', 'arithmetic', 'dependency', 'function',
-        'constfold', 'function'
+        'function', 'dependency'
     ]
+    # This flag disables folding QDQ nodes around constants in the network (eg: around conv/FC weights)
+    rewrite_options.experimental_disable_folding_quantization_emulation = True
```
fatcat-z, line 690: This does not exist in tensorflow v1.x versions which is also need to be supported by us.

Please add a validation(is_tf2()) here to make sure this will only be called when tf version is 2.x so the CI could pass.

peri044: Thank you @fatcat-z Added it now.

tf2onnx/onnx_opset/nn.py
```diff
@@ -139,12 +139,25 @@ def conv_convert_inputs(ctx, node, with_kernel=False, new_kernel_shape=None,
...
+            if kernel_node.op.op_type == 'DequantizeLinear':
+                # Assuming the model was trained in NHWC in TF,
+                # the weights would be in [fH, fW, C_in, C_out].
+                # orig_conv_weights -> Q -> DQ -> new_conv_weights -> conv
+                weights_node = kernel_node.inputs[0].inputs[0]
+                val = weights_node.get_tensor_value(as_list=False)
+                val = np.transpose(val, permutation)
+                weights_node.set_tensor_value(val)
+                need_transpose = False
+                #Change the quantization axis for Q and DQ node accordingly
```
fatcat-z, line 153: Please update it to "# Change ..." for the format.

peri044: @fatcat-z  Done. There are some tflite errors which are unrelated to my changes. Are those supposed to pass ?

tf2onnx/tf_loader.py
```diff
@@ -684,8 +684,12 @@ def tf_optimize_grappler(input_names, output_names, graph_def, fold_constant=Non
     # depends on so for now don't turn this on, fold_constant is always enabled now.
     rewrite_options.optimizers[:] = [
         # 'pruning', 'constfold', 'arithmetic', 'dependency', 'function',
-        'constfold', 'function'
+        'function', 'dependency'
     ]
+    if is_tf2():
```
fatcat-z, line 689: Looks like this flag was effective only after v2.4.0, so this is not good enough and needs to be changed like below:

if LooseVersion(tf.__version__) > "2.4":

tf2onnx/tf_loader.py
```diff
@@ -684,8 +684,12 @@ def tf_optimize_grappler(input_names, output_names, graph_def, fold_constant=Non
     # depends on so for now don't turn this on, fold_constant is always enabled now.
     rewrite_options.optimizers[:] = [
         # 'pruning', 'constfold', 'arithmetic', 'dependency', 'function',
-        'constfold', 'function'
+        'function', 'dependency'
```
fatcat-z, line 687: Since experimental_disable_folding_quantization_emulation is not available for some versions, we still need to bring "constfold" optimizer back in those cases.

fatcat-z:
@peri044 

There is a PR(https://github.com/onnx/tensorflow-onnx/pull/1843) has been merged into master branch to resolve some CI issues which has no relationship with your code. Please sync this change and check the results again.

gcunhase:
@fatcat-z 

It seems the last issue is with my username (it's `Gwena-Workstation` instead of `gcunhase`), so the sign-off is not working. How do you recommend we fix this issue? On the [DCO details](https://github.com/onnx/tensorflow-onnx/pull/1839/checks?check_run_id=5225724520), it says to only follow the steps there if I'm the only author of the commits in this branch, which is not true. TIA.

fatcat-z:
> @fatcat-z
> 
> It seems the last issue is with my username (it's `Gwena-Workstation` instead of `gcunhase`), so the sign-off is not working. How do you recommend we fix this issue? On the [DCO details](https://github.com/onnx/tensorflow-onnx/pull/1839/checks?check_run_id=5225724520), it says to only follow the steps there if I'm the only author of the commits in this branch, which is not true. TIA.

There is no too much experience for such case yet. Is it possible you can sync with @peri044 to create a new PR with all latest changes of this one and submitted by @peri044 so that your user name won't bother any more?

peri044:
@fatcat-z closing this in favor of https://github.com/onnx/tensorflow-onnx/pull/1856. Hopefully everything should pass now and ready to merge. Thanks !!

PR #1855, opened 2022-02-17
Add used opset into opset_imports argument to create an onnx model.

fatcat-z:
In latest ONNX version, there is an enhancement that if the used domain of a node doesn't exist in the model.opset_import, it will result in an exception and stop processing.

To align with this enhancement, we add current used domain and opset version into model.opset_import property by appending them into "opset_imports" argument in make_model() method.

### DIFF
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -1193,10 +1193,8 @@ class Graph(object):
                       "producer_version": __version__}
 
         if "opset_imports" not in kwargs:
-            opsets = []
-            imp = OperatorSetIdProto()
-            imp.version = self._opset
-            opsets.append(imp)
+            opsets = [helper.make_opsetid(constants.ONNX_DOMAIN, self._opset)]
+            opsets.append(constants.AI_ONNX_ML_OPSET)
             if self.extra_opset is not None:
                 opsets.extend(self.extra_opset)
             kwargs["opset_imports"] = opsets
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -11,7 +11,7 @@ import logging
 import six
 import numpy as np
 
-from onnx import helper, numpy_helper, shape_inference, OperatorSetIdProto, AttributeProto, TensorProto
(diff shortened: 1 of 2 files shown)


tf2onnx/graph.py
```diff
@@ -1193,10 +1193,8 @@ def make_model(self, graph_doc, optimize=False, graph_name="tf2onnx", external_t
                       "producer_version": __version__}
 
         if "opset_imports" not in kwargs:
-            opsets = []
-            imp = OperatorSetIdProto()
-            imp.version = self._opset
-            opsets.append(imp)
+            opsets = [helper.make_opsetid(constants.ONNX_DOMAIN, self._opset)]
+            opsets.append(constants.AI_ONNX_ML_OPSET)
```
hwangdeyu, line 1197: Add a condition to check if `AI_ONNX_MLDOMAIN opset 2` in list `opsets`?
And the latest `ai.onnx.ml` is `3`, should we also add it? https://github.com/onnx/onnx/blob/main/docs/Versioning.md#released-versions

fatcat-z: ORT doesn't support 3 for this domain yet, so 2 is good enough.

hwangdeyu (approved):

guschmue:
why add onnx.ml as domain if no op in the graph uses it ? It would be super rare for tf2onnx to use a onnx.ml op

guschmue:
yes, only add the domain when we use onnx.ml. The idea behind the domains is that you tell the runtime what is needed to run the model (ie. one can load the support for domains on demand).

fatcat-z:
> yes, only add the domain when we use onnx.ml. The idea behind the domains is that you tell the runtime what is needed to run the model (ie. one can load the support for domains on demand).

Adding this built-in domain in model.opset_import field is to align with a latest check enhancement of onnx. It won't impact the way how runtime runs the model because the runtime will only consider the domain of each op node. 
For the pair of domain and version in model.opset_import field, runtime will only check if they are supported.

PR #1869, opened 2022-02-26
Fix keras bidirectional merge failures

kota-row:
Fixed an issue where the Bidirectional layers generated by Keras were not correctly combined into a single LSTM/GRU.

currently biltm_rewriter/bigru_rewriter only supports the following structure
```
        /-------------(optional: Transpose) -- forward RNN -- Squeeze -- (optional: Transpose)
shared input
        \-- Reverse---(optional: Transpose) -- backward RNN -- Squeeze -- (optional: Transpose) -- Reverse
```

but, tf2 keras Bidirectional Layer generates different graph
* There are several Identity layers in between.
* shared_input -> Transpose -> Reverse -> backward (Unlike the above figure, Transpose and Reverse have been swapped)
* return_sequence=False, which has not Reverse after Squeeze, but has StridedSlice

I also added a test to check that all Bidirectional recurrent layers are combined into one. The first commit will fail the test.


tf2onnx/rewriter/rnn_utils.py
```diff
@@ -554,27 +570,41 @@ def belong_to_birnn(g, fw_rnn, bw_rnn, rnn_type):
...
-            are_all_reverse = all([utils.is_tf_reverse_op(r_op) for r_op in reverse_nodes])
-            if are_all_reverse:
-                return reverse_nodes
+        reverse_or_slice_nodes = g.find_output_consumers(s.output[0])
+        if len(reverse_or_slice_nodes) == 1:
+            if reverse_or_slice_nodes[0].type == "Transpose":
+                reverse_or_slice_nodes = g.find_output_consumers(reverse_or_slice_nodes[0].output[0])
+
+            if len(reverse_or_slice_nodes) == 1 and reverse_or_slice_nodes[0].type == "Identity":
+                reverse_or_slice_nodes = g.find_output_consumers(reverse_or_slice_nodes[0].output[0])
```
fatcat-z, line 597: Should we do this twice here? Or update it to :

            if len(reverse_or_slice_nodes) == 1 and reverse_or_slice_nodes[0].type == "Identity":
                reverse_or_slice_nodes = g.find_output_consumers(reverse_or_slice_nodes[0].output[0])
                if len(reverse_or_slice_nodes) == 1 and reverse_or_slice_nodes[0].type == "Identity":
                    reverse_or_slice_nodes = g.find_output_consumers(reverse_or_slice_nodes[0].output[0])

tf2onnx/rewriter/rnn_utils.py
```diff
@@ -554,27 +570,41 @@ def belong_to_birnn(g, fw_rnn, bw_rnn, rnn_type):
...
+
+            if len(reverse_or_slice_nodes) == 1 and reverse_or_slice_nodes[0].type == "Identity":
+                reverse_or_slice_nodes = g.find_output_consumers(reverse_or_slice_nodes[0].output[0])
+
```
fatcat-z, line 601: Do we need to do it twice here? Or update it like:

            if len(reverse_or_slice_nodes) == 1 and reverse_or_slice_nodes[0].type == "Identity":
                reverse_or_slice_nodes = g.find_output_consumers(reverse_or_slice_nodes[0].output[0])
                if len(reverse_or_slice_nodes) == 1 and reverse_or_slice_nodes[0].type == "Identity":
                    reverse_or_slice_nodes = g.find_output_consumers(reverse_or_slice_nodes[0].output[0])

tf2onnx/rewriter/rnn_utils.py
```diff
@@ -620,12 +650,27 @@ def slice_birnn_for_original_rnn_consumers(g, rnn_fw, rnn_bw, bi_rnn, rnn_output
     if rnn_output_index == 0:
         axis = 1
         # remove reverse op for rnn_bw
```
fatcat-z, line 652: Please update the comments accordingly to involve tailed slice node.

tf2onnx/rewriter/rnn_utils.py
kota-row: yes, tensorflow makes one or two Identity, it depends on Layer arguments.
I changed the code to nested if.

fatcat-z (approved):
Thanks for your contribution!

PR #1873, opened 2022-03-02
Force to upgrade typing-extensions to a version above 4.0.0.

fatcat-z:
For onnx-weekly CI job, upgrade the typing-extensions to a version above 4.0.0 to avoid unexpected crash on opset 15.


ci_build/azure_pipelines/templates/setup.yml
```diff
@@ -62,6 +62,8 @@ steps:
       # onnx-weekly won't satisfy onnx requirement so uninstallation must happen here
       pip uninstall -y onnx
       pip install --index-url https://test.pypi.org/simple/ ${PIP_ONNX_NAME:-onnx}
+      pip uninstall typing-extensions
```
hwangdeyu, line 65: I guess it's related to `onnx-weekly-1.10.1.dev20210830`,
The latest one is `1.11.0.dev20220228`, but it will install fail.

hwangdeyu (approved):

PR #1896, opened 2022-03-27
Upgrade some keras jobs to use ort 1.11.0.

fatcat-z:
Upgrade some keras jobs to use ort 1.11.0.


hwangdeyu (approved):
LGTM!

PR #1907, opened 2022-04-07
Allow passing optimizers to from_keras conversion function

shelkesagar29:
`_convert_common` function in `convert.py` has `optimizers` parameter which allows user to select a subset of `_optimizers` from `tf2onnx.optimizer` while conversion is being done.

One way of using it to convert a TF saved model can be as follows.
```
from tf2onnx.optimizer import _optimizers
# Get all optimizers
updated_optimizers = copy.deepcopy(_optimizers)
# Remove two optimizers and then apply remaining optimizers during conversion
del updated_optimizers["q_dq_optimizer"]
del updated_optimizers["const_dequantize_optimizer"]

# 1. Load TensorFlow saved model
graph_def, inputs, outputs = tf_loader.from_saved_model(
    model_path=saved_model_path,
    input_names=None,
    output_names=None,
    tag="serve",
    signatures=["serving_default"]
)

# 2. Convert tf2onnx and save onnx file
model_proto, _ = convert._convert_common(
    graph_def,
    name=saved_model,
    opset=13,
    input_names=inputs,
    output_names=outputs,
    output_path=onnx_output_path,
    optimizers=updated_optimizers
)
```
However, currently there is no way to pass optimizers when using `from_keras` function to directly convert a keras model (without saving and loading). Converting from Keras directly is important since TF gives error many times for saving custom layers (this is especially true for QAT).

This PR adds `optimizers` parameter to `from_keras` function and passes it to `_convert_common`.


tf2onnx/convert.py
```diff
@@ -417,6 +417,7 @@ def from_keras(model, input_signature=None, opset=None, custom_ops=None, custom_
         inputs_as_nchw: transpose inputs in list from nchw to nhwc
         large_model: use the ONNX external tensor storage format
         output_path: save model to output_path
+        optimizers: list (subset) of tf2onnx optimizers if applying all optimizers is not desired.
```
hwangdeyu, line 420: Please remove the Trailing whitespace as `pylint` complainting.

hwangdeyu:
Thanks for contribution!
Could you sign off every commit message following DCO required?
```
Signed-off-by: Author Name <authoremail@example.com>
```

hwangdeyu (approved):
LGTM.

PR #1909, opened 2022-04-10
Add the absolute path condition when loads the tf custom op library

hwangdeyu:
The path in `tf.load_op_library()` must be absolute path.


fatcat-z (approved):
LGTM, thanks!

PR #1933, opened 2022-05-07
Bugfix for concatenating node instead of str.

FrankD412:
There are two exceptions that attempt to concatenate a node to the exception string. This PR simply calls into the name property to get the node name out.


fatcat-z (approved):
LGTM, thanks for your contribution.

Issue #1848, opened 2022-02-15
Bug with conversion of tf function with no inputs


iolkhovsky:
**Describe the bug**
Conversion of a tf.function fails ([convert.from_function](https://github.com/onnx/tensorflow-onnx/blob/master/tf2onnx/convert.py#L507)  throws unexpected exception [here](https://github.com/onnx/tensorflow-onnx/blob/master/tf2onnx/convert.py#L535)) in case the graph has no any inputs (i.e. when signature is an empty list). It seems that a model with no inputs is essentially a correct use-case, so the suggestion is to replace [existing check](https://github.com/onnx/tensorflow-onnx/blob/master/tf2onnx/convert.py#L534) with something like:
`if input_signature is None:`

**System information**
- Ubuntu 20.04.3 LTS
- Tensorflow 2.8.0:
- Python 3.8.10 [GCC 9.3.0] on linux

**To Reproduce**
```
import tensorflow as tf
import tf2onnx


def no_inputs_graph():
    return tf.constant(1.0, dtype=tf.float32)


tf_func = tf.function(func=no_inputs_graph, input_signature=[])


class ProxyModule(tf.Module):
    def __init__(self, tf_func):
        super().__init__()
        self.apply = tf_func
        

SAVED_MODEL_PATH = "graph.SavedModel"
proxy = ProxyModule(tf_func)
tf.saved_model.save(proxy, SAVED_MODEL_PATH, signatures={"apply": proxy.apply})
restored = tf.saved_model.load(SAVED_MODEL_PATH)
print("Tensorflow inference result: ", restored.apply())

onnx_proto, _ = tf2onnx.convert.from_function(function=tf_func, input_signature=[], output_path=None, opset=14)
```

**Additional context**
The example above generates the next output:
```
INFO:tensorflow:Assets written to: graph.SavedModel/assets
Tensorflow inference result:  tf.Tensor(1.0, shape=(), dtype=float32)

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [10], in <module>
     21 restored = tf.saved_model.load(SAVED_MODEL_PATH)
     22 print("Tensorflow inference result: ", restored.apply())
---> 24 onnx_proto, _ = tf2onnx.convert.from_function(function=tf_func, input_signature=[], output_path=None, opset=14)

[...]

PR #1947, opened 2022-05-24
Update the way to check input_signature in from_function().

fatcat-z:
When a function was converted to an ONNX model, the input_signature will be checked. The logic is we want to check if the input_signature is None, but the current way may make a mistake for some inputs which are not TensorSpec but still valid, so correct it.


fatcat-z:
This PR will resolve one issue in https://github.com/onnx/tensorflow-onnx/issues/1848.

hwangdeyu (approved):
LGTM!

Issue #1938, opened 2022-05-16
Feature request: tf2onnx.convert.from_tflite


josephrocca:
Wondering if this is on the road map, since the other formats all have a Python API for programmatic conversion?

PR #1954, opened 2022-05-30
Add a new api from_tflite to improve user experience.

fatcat-z:
1. Add a new api from_tflite() so users could call it to import a tflite model.
2. Add relative test code as well.

Fix https://github.com/onnx/tensorflow-onnx/issues/1938


hwangdeyu (commented):
Add the description of new api `from_tflite` in https://github.com/onnx/tensorflow-onnx#python-api-reference?

tests/test_api.py
```diff
@@ -232,5 +232,18 @@ def test_graphdef(self):
         self.assertAllClose([2.1193342], oy[0], rtol=0.1, atol=0.1)
 
 
+    def test_tflite(self):
```
hwangdeyu, line 235: Should we also check the test `@check_tf_min_version("1.15")`, like the other API test?

fatcat-z:
> Add the description of new api `from_tflite` in https://github.com/onnx/tensorflow-onnx#python-api-reference at the same time?

Good catch!

hwangdeyu (approved):
LGTM!

PR #1960, opened 2022-06-08
Add shape inference as an arg for tools/onnx-optimize.py

hwangdeyu:
In https://github.com/onnx/tensorflow-onnx/blob/1915fb52270856debc74ba5d28e332e46ccc1e1a/tools/onnx-optimize.py#L59 the `optimizer.optimize_graph()` method would eliminate infered shape.
This PR adds the shape inference after model optimized.


tools/onnx-optimize.py
```diff
@@ -64,10 +70,12 @@ def main():
 
     model_proto = helper.make_model(onnx_graph, **kwargs)
 
+    model_proto_inferred = model_shape_inference(model_proto)
```
hwangdeyu, line 73: change it as a optional argument.

xadupre (approved):

PR #1964, opened 2022-06-10
Transpose optimization for Softmax and LogSoftmax (fixes #1716)

### LINKED ISSUE #1716: Feature: Transpose optimization for Softmax

janbernloehr:
This is reopening #1727 by @fthielke to fix the remaining issues to get it merged.

**Original Description**
In opsets 13 and higher, the axis of the operation is arbitrary and can simply be changed according to the permutation of the Transpose.
In lower opsets, Softmax always coerces its inputs to a 2D tensor, making Transpose operations necessary if the permutation moves axes between the coerced batch and feature dimensions.

### DIFF
--- a/tests/test_optimizers.py
+++ b/tests/test_optimizers.py
@@ -1369,7 +1369,9 @@ class OptimizerTests(Tf2OnnxBackendTestBase):
         self.run_transpose_compare(["res"], {"X": np.random.randn(*input_shape).astype(np.float32)},
                                    model_proto, remaining_transpose_num=0)
 
-    @check_opset_max_version(12, "Before opset 13, Softmax coerced its inputs to 2D and can thus only be optimized for certain permutations")
+    @check_opset_max_version(
+        12, "Before opset 13, Softmax coerced its inputs to 2D and can thus only be optimized for certain permutations"
+    )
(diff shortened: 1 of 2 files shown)


janbernloehr:
@fatcat-z , @TomWildenhain-Microsoft : Maybe you can take another look. pylint issues have been fixed.

fatcat-z (approved):

PR #1980, opened 2022-06-24
Update Keras related tests to support latest TF version.

fatcat-z:
Since TF 2.6.0, tensorflow.python.keras namespace will be retired. Some modules and functions are valid after that. So we make some changes to make the test code supports all of TF versions, no matter they are older than 2.6.0, or later than 2.6.0.


hwangdeyu (approved):
LGTM！

Issue #1941, opened 2022-05-17
Using TransposeOptimizer breaks model (again)


CarlPoirier:
**Describe the bug**
I have an onnx model on which I want to run the [onnx-optimize.py](https://github.com/onnx/tensorflow-onnx/blob/master/tools/onnx-optimize.py) script. All the optimizers work fine except the TransposeOptimizer. When I use it, my onnx model is not working anymore, with onnxruntime at least. I get this error :

`[ONNXRuntimeError] : 6 : RUNTIME_EXCEPTION : Non-zero status code returned while running Split node. Name:'Split_12' Status Message: D:\a\_work\1\s\onnxruntime\core\providers\cpu\tensor\split.cc:160 onnxruntime::Split::ComputeImpl split_tensor->Shape().NumDimensions() == 1 was false. An split tensor must be a vector tensor.`

**Urgency**
None, for now I deactivated this optimizer and I'm using all the others.

**System information**
- OS Platform and Distribution: Windows 10
- Tf2onnx version: 1.10.1
- Tensorflow Version: 2.5.2
- Python version: 3.9.9
- ONNXRuntime: 1.9.3

**To Reproduce**
[test_models.zip](https://github.com/onnx/tensorflow-onnx/files/8710649/test_models.zip)

The above zip contains four files:

- test-run.py : script to run the model using onnxruntime
- test_model2_small.onnx : starting model which works
- test_model2_small_broken.onnx : optimized model which does not work anymore
- test_model2_small_ok.onnx : optimized model for which I have deactivated only the transpose optimizer and it still works

I'm running the onnx-optimize script this way:

`python onnx-optimize-orig.py --input test_model2_small.onnx --output test_model2_small_broken.onnx
2022-05-17 19:13:24.283642: I tensorflow/stream_executor/platform/default/dso_loader.cc:53] Successfully opened dynamic library cudart64_110.dll
2022-05-17 19:13:27,704 - INFO - Optimizing ONNX model
2022-05-17 19:13:27,732 - INFO - After optimization: Const -1 (7->6), Identity -1 (1->0)`

Then I run the model with any image, this way:

`python test-run.py`

If you need any more information, let me know!

hwangdeyu:
@CarlPoirier  The latest tf2onnx verison 1.12 has been released. I think your model can be optimized as expected. Thanks!

PR #1996, opened 2022-07-12
Fix transpose split optimize attr when opset >=13

hwangdeyu:
split is an optional input not attr since opset 13.
transpose optimizer should adapt to ONNX version >= 13 simultaneously.

fixes https://github.com/onnx/tensorflow-onnx/issues/1941

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -672,15 +672,15 @@ class TransposeOptimizer(GraphOptimizerBase):
     def _split_handler(self, trans, node):
         # Todo: need handle cases where Split node has more than 1 outputs.
         split = None
-        if len(node.input) > 1 and node.inputs[1].is_const():
+        if self._g.opset >= 13 and len(node.input) > 1 and node.inputs[1].is_const():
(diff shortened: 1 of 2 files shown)


tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -671,11 +671,19 @@ def _concat_handler(self, trans, node):
 
     def _split_handler(self, trans, node):
         # Todo: need handle cases where Split node has more than 1 outputs.
+        split = None
```
hwangdeyu, line 674: This `split` represents the second input of `Split` op when opset >= 13.

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -671,11 +671,19 @@ def _concat_handler(self, trans, node):
...
+        if len(node.input) > 1 and node.inputs[1].is_const():
```
fatcat-z, line 675: Please add opset version check here.

hwangdeyu: Added it.

fatcat-z (approved):
LGTM, thanks!

PR #2000, opened 2022-07-15
Skip existing const node in _parse_graph_input

q-ycong-p:
If the Graph object being constructed already contains a node of
the name of a Const node in orginal graph, do not add as input.

Signed-off-by: congyc <congyc@amazon.com>


q-ycong-p:
There're two points I'd like to raise for conversion:
1. **Reason for the PR change**: ONNX < IR-4 has initializers added as graph inputs. This leads to models such as `non_quantized.onnx` (attached within repro_issue.tar.gz) to not be optimized as intended. See `optimized_wrong.onnx` (in repro_issue.tar.gz) for the incorrect optimization prior to PR change, and `optimized_correct.onnx` (in repro_issue.tar.gz) for the expected optimization post to PR change.

2. **Why change a unit test**: This PR change exposes a potential issue within the unit test. `test_duplicated_duplicated_constant_and_initializer` expects optimized model to still contain 2 initializer. But looking at the non_optimized.onnx (in unit_test.tar.gz), only 1 initializer is actually needed. See the `optimized_2_inits_left.onnx` (in unit_test.tar.gz) for model produced prior to PR change, and `optimized_1_inits_left` (in unit_test.tar.gz) for model post to PR change. As far as I can see, the later model where only `value0` is retained as initializer appear to be what we'd expect.

Please correct me if my understanding on initializer and Const node appear wrong.

hwangdeyu (approved):

hwangdeyu:
> Thanks @hwangdeyu! When attempting merge, it says "Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests". Could you or other contributor of this community merge it?

Sure☺, I was waitting for CI passed on yesterday.

PR #2001, opened 2022-07-15
Add handling of HardSigmoid recurrent activation for Keras LSTM

q-ycong-p:
This commit adds pattern matching and parsing where Keras LSTM uses
HardSigmoid as the recurrent activation. Refactored to allow addition
of other activations when needed.

Signed-off-by: Yu Cong <congyc@amazon.com>

### DIFF
--- a/tests/test_lstm.py
+++ b/tests/test_lstm.py
@@ -751,6 +751,28 @@ class LSTMTests(Tf2OnnxBackendTestBase):
             return tf.identity(y[0], name="output"), tf.identity(y[1], name="output1")
         self.run_test_case(func, {"input:0": x_val}, [], ["output:0", "output1:0"], rtol=1e-05, atol=1e-06)
 
+    @check_tf_min_version("2.0")
+    @skip_tf_versions("2.1", "Bug in TF 2.1")
+    def test_keras_lstm_recurrent_activation_is_hard_sigmoid(self):
+        in_shape = [10, 3]
+        x_val = np.random.uniform(size=[2, 10, 3]).astype(np.float32)
+
+        model_in = tf.keras.layers.Input(tuple(in_shape), batch_size=2)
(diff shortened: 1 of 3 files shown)


q-ycong-p:
The existing tf2 lstm rewriter exclusively handles {sigmoid,tanh,relu} as {recurrent,}activations. There are [other activations](https://www.tensorflow.org/api_docs/python/tf/keras/activations) defined in `tf.keras.activations`. For example, a `tf.keras.layers.LSTM` layer's recurrent activation may be `hard_sigmoid`. tf2onnx frozen graph alters patterns to represent the hard sigmoid computation as low-level ops (see attached screenshot). This messes existing pattern matching and fails tf2 lstm rewriter. This PR extends support to handle this variation, and refactors pattern matching to allow easy plug-in of other activation patterns in the future (i.e. ad hoc as needed).

See below for minimal example to reproduce the original issue. Prior to this PR, loop op is produced from Keras LSTM layer. This PR extends rewriter pattern matching to convert LSTM node correctly.
```
import tensorflow as tf
import tf2onnx

input = tf.keras.Input(shape=(10, 8))
lstm_output = tf.keras.layers.LSTM(4, recurrent_activation="hard_sigmoid")(input)
model = tf.keras.Model(input, lstm_output)
onnx_model, _ = tf2onnx.from_keras(model) # onnx model contains loop instead of LSTM op
```

Screen Shot 2022-07-15 at 2 23 29 PM

hwangdeyu (approved):
LGTM, thanks for your contribution! 

The `HardSigmoid` as the recurrent activation would be very helpful. Could you also tell us which model you are using to come up with this issue?

q-ycong-p:
Hi @hwangdeyu , thank you for taking a look. The original model is internal. The general architecture is a simple LSTM encoder. If all is good, would you mind merging this change? Thanks!

Issue #2002, opened 2022-07-18
Failed to convert LSTM cell with activation and dropout settings


Andreas-Krebbel:
**Describe the bug**
Keras LSTM model setting activation function and dropout is decomposed instead of using ONNX LSTM cell.

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Fedora 35, x86
- Tensorflow Version: 2.9.1
- Python version: 3.9.13
- tf2onnx: e896723 (2022-07-15)

**To Reproduce**
```
from keras.layers import Input, LSTM
from keras.models import Sequential
import tf2onnx

model = Sequential()
model.add(Input(shape=(16, 16)))
model.add(LSTM(16, activation='sigmoid',  dropout=0.1))

model.save('test.h5')
onnx_model = tf2onnx.convert.from_keras(model, opset=16, output_path='test.onnx')
```
Removing either activation='sigmoid' or dropout=0.1 generates an LSTM cell in the output as expected.

**Screenshots**
test.h5:
![Screenshot from 2022-07-18 11-33-49](https://user-images.githubusercontent.com/38103320/179484101-dd04144f-eb05-4ebd-920f-934fd33ca011.png)
test.onnx
![Screenshot from 2022-07-18 11-34-06](https://user-images.githubusercontent.com/38103320/179484114-ca7ce693-a599-40be-a2f5-b815c3dc1593.png)

hwangdeyu:
tf2onnx is not expected supporting converting models for training so far, such as setting dropout.
But in this case, it seems like the root cause is from the `tensorflow OptimizeGraph`. We will consider to add `dependency` to optimization.

PR #2027, opened 2022-08-19
Add lstm sigmoid dropout test for dependency optimization

hwangdeyu:
Related to https://github.com/onnx/tensorflow-onnx/issues/2002.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -454,7 +454,7 @@ def group_nodes_by_type(graph):
 
 
 def check_op_count(graph, op_type, expected_count, disabled=True):
-    # FIXME: after switching to grappler some of the op counts are off. Fix later.
+    # The grappler optimization may effect some of the op counts.
     return disabled or len(group_nodes_by_type(graph)[op_type]) == expected_count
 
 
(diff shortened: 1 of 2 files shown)


tests/common.py
```diff
@@ -454,7 +454,7 @@ def group_nodes_by_type(graph):
 
 
 def check_op_count(graph, op_type, expected_count, disabled=True):
-    # FIXME: after switching to grappler some of the op counts are off. Fix later.
+    # The grappler optimization may effect some of the op counts.
```
fatcat-z, line 457: may effect -> will change

tests/test_lstm.py
```diff
@@ -51,6 +51,7 @@ def new_graph_validator(g):
                 # Skip checks for tflite graphs (no ":" in outputs)
                 return good
             good = good and check_op_count(g, "LSTM", require_lstm_count, disabled=False)
+            # if LSTM op rewriter matchs failed, Loop op will be shown generally.
```
fatcat-z, line 54: Let's change it to:

If LSTM op rewriter failed to work, Loop op will be shown in general.

tests/common.py
hwangdeyu: Somethings, it will change. So I change it to `may`.

fatcat-z (approved):
LGTM, thanks!

Issue #1871, opened 2022-02-28
Keras LSTM not converted correctly when precedent Embedding layer specifies mask_zero=True


q-ycong-p:
**Describe the bug**
When an `tf.keras.layers.Embedding` with attribute `mask_zero=True` attribute precede LSTM layer, the LSTM is **converted into loops instead of LSTM op**.

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Linux
- Tensorflow Version: 2.4.0
- Python version: 3.8.12

**To Reproduce**
```
# minimal example of problematic keras model
x = tf.keras.layers.Input(shape=(4,), dtype="int32")
e = tf.keras.layers.Embedding(5, 5, mask_zero=True)(x)
rnn = tf.keras.layers.LSTM(3, return_sequences=True)(e)[0]
model = tf.keras.Model(inputs=x, outputs=rnn)

# converted onnx will have loops instead of LSTM op
onnx_model, _ = tf2onnx.convert.from_keras(model)
onnx.save(onnx_model, "lstm_masking_zero.onnx")
```

**Screenshots**
<img width="1011" alt="Original keras model to be converted" src="https://user-images.githubusercontent.com/50014064/156077571-5363d262-5580-4acf-9e9b-402effcde6d3.png">

<img width="289" alt="Unsuccessful conversion of lstm into loops" src="https://user-images.githubusercontent.com/50014064/156077719-db2bdb72-381c-421f-be96-75e4d6e08cb9.png">

**Additional context**
I've tried modified `lstm_tf2_rewriter` to accommodate the new pattern in rewriter parsing. Although I can skip the extra `SelectV2` pattern and get LSTM op in final onnx model, I am not able to correctly handle the masking zero information. My attempt will result in incorrect inference result if 0 is contained in input. 

Below is my unsuccessful attempt: masking is ignored and results in incorrect inference result. Any suggestion on how masking should be handled? Thank you!
```
--- a/tf2onnx/rewriter/lstm_tf2_rewriter.py
+++ b/tf2onnx/rewriter/lstm_tf2_rewriter.py
@@ -56,21 +56,22 @@ def rewriter_lstm_tf2(g, ops):
             # extract output h_t
             ht_mul = match_result.get_op("ht")
             final_consumers = g.find_output_consumers(ht_mul.output[0])
-            select_ops = [n for n in final_consumers if n.type == "Select"]
[...]

PR #2030, opened 2022-08-27
Add masked LSTM support

q-ycong-p:
Masking is a intra-layer behavior in TF LSTM [1] but is not a intra-op behavior in ONNX LSTM [2]. When converted to ONNX, masked TF LSTM layer is converted to Loop op. This over-complicates the ONNX model, and has a negative impact on inference performance in ORT without leveraging LSTM optimizations. (issue #1871)

This commit adds support to convert masked LSTM correctly, **under the important assumption that input must be post-padded** - which is the most common use case. The "masking" info is conveyed to ONNX LSTM op as `sequence_lens` which is dynamically computed by summing the number of non-skip timesteps per batch per-LSTM. This behavior is implemented with reference to keras2onnx PR#386 [3]. Additional logic is added for backward LSTM so that the input sequence is reversed correctly given `sequence_lens`.

Note that if mask-enabled, and LSTM input is pre- or randomly padded, the converted ONNX model will behave incorrectly for inference. Unless ONNX add new attribute e.g. `mask_enabled` to RNN ops, converter alone may not be able to handle generic masking while keeping the RNN ops, since masking alters intra-op behavior. With such limitation, I'd like to share this PR for further comment and suggestion.

[1] https://www.tensorflow.org/guide/keras/masking_and_padding#masking
[2] https://github.com/onnx/onnx/blob/main/docs/Operators.md#LSTM
[3] https://github.com/onnx/keras-onnx/pull/386

---
#### Forward LSTM
Here's an minimal example with an embedded LSTM (`mask_zeros=True`):
- H5 model:
Screen Shot 2022-08-26 at 6 18 13 PM

- tf2onnx-converted ONNX model, **before** proposed change:
Screen Shot 2022-08-26 at 6 17 49 PM

- tf2onnx-converted ONNX model, **after** proposed change:
Screen Shot 2022-08-26 at 6 18 57 PM

#### Reverse LSTM
- Need to alter `tf.raw_op.ReverseV2`->`ReverseSequence` behavior to reverse LSTM input correctly:
reverse_masked_lstm

### DIFF
--- a/tests/test_lstm.py
+++ b/tests/test_lstm.py
@@ -793,5 +793,164 @@ class LSTMTests(Tf2OnnxBackendTestBase):
             return tf.identity(y[0], name="output")
         self.run_test_case(func, {"input:0": x_val}, [], ["output:0"], rtol=1e-05, atol=1e-06)
 
+    @check_tf_min_version("2.0")
+    @skip_tf_versions("2.1", "Bug in TF 2.1")
+    def test_keras_masked_lstm_embedding_unidirectional(self):
+        for go_backwards in [True, False]:
+            timesteps = 4
+            # Note: masked LSTM only support post-padded input after conversion
+            # test case sequence_lens = [4, 2, 0]
+            x_val = np.array([
+                [1, 2, 3, 4],
+                [5, 6, 0, 0],
+                [0, 0, 0, 0]
+            ], dtype=np.int32)
+
+            model_in = tf.keras.layers.Input((timesteps,), dtype="int32")
+            x_embedding = tf.keras.layers.Embedding(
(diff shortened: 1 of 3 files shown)


q-ycong-p:
Sorry will address the test failures on TF-2.9 soon.

AndreyOrb:
Hello,
Is there any progress with this issue?

AndreyOrb:
Hi,
Is there any update? Will the proposed code work if pulled?

AndreyOrb:
@xadupre Hello Xavier. I've been waiting for this PR for a few years now. Could you assist, please?

PR #2032, opened 2022-08-30
Fix keras2onnx_unit_test nightly job opset version failures.

fatcat-z:
1. Python 3.10 doesn't support onnxruntime 1.11, so update the target ort version to 1.12.
2. If ONNX version > ONNXRuntime version, there is a problem caused by opset version. Fix this issue.

### DIFF
--- a/tests/keras2onnx_unit_tests/test_utils.py
+++ b/tests/keras2onnx_unit_tests/test_utils.py
@@ -17,7 +17,7 @@ import urllib
 
 # Mapping opset to ONNXRuntime version.
 ORT_OPSET_VERSION = {
-    "1.6.0": 13, "1.8.0": 14, "1.9.0": 15, "1.11.0": 16, "1.12.0": 17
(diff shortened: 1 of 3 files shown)


hwangdeyu (approved):
LGTM. Thanks!

tests/keras2onnx_unit_tests/test_utils.py
```diff
@@ -8,10 +8,18 @@
...
+from packaging.version import Version
+from tf2onnx.keras2onnx_api import convert_keras
 import time
 import json
 import urllib
 
+
+# Mapping opset to ONNXRuntime version.
+ORT_OPSET_VERSION = {
+    "1.6.0": 13, "1.7.0": 13, "1.8.0": 14, "1.9.0": 15, "1.10.0": 15, "1.11.0": 16, "1.12.0": 17
```
hwangdeyu, line 20: Marked, in case we need add new version in the future.

PR #2033, opened 2022-08-31
Update the way to get opset version for running tests.

fatcat-z:
Some tests are using get_maximum_opset_supported() function to get the opset version. Replace it with a new method which will consider the max opset version supported by current onnxruntime.


hwangdeyu (approved):
LGTM.

PR #2034, opened 2022-09-01
Update left tests in test_subclassing for the opset version issue.

fatcat-z:
If ONNX version > ONNXRuntime version, there is a problem caused by opset version. Fix this issue in test_subclassing file.


hwangdeyu (approved):
LGTM！

PR #2090, opened 2022-12-01
Fixed group attribute in convolution op

### LINKED ISSUE #2084: Negative group attribute in generated convolution operations
**Describe the bug**
Generated ONNX convolution nodes may get a negative `group` attribute  value in certain cases. The graph is generated without errors, but trying to run it causes an error in ONNXRuntime.

**Urgency**
NA

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 18.04*): Windows 10
- TensorFlow Version: 2.4.1
- Python version: 3.7
- ONNX version (if applicable, e.g. 1.11*): 1.10.2
- ONNXRuntime version (if applicable, e.g. 1.11*): 1.9.0

**To Reproduce**

```python
import tensorflow as tf
import tf2onnx
import numpy as np
import onnxruntime

# Convolution function
@tf.function
def conv1d(x, y):
    return tf.nn.conv1d(x, y, [1], 'VALID', name='Conv')
# Convert to ONNX model
model_proto, _ = tf2onnx.convert.from_function(
    conv1d,
    # The number of input channels in the filters is left unspecified
    input_signature=[tf.TensorSpec([None, None, 3]), tf.TensorSpec([10, None, 5])])
# Find convolution node
conv_node = next(n for n in model_proto.graph.node if n.name == 'Conv')
# Find group attribute
group_attr = next(a for a in conv_node.attribute if a.name == 'group')
# Attribute has a negative value
print(group_attr)
# name: "group"
# i: -3
# type: INT

# Try to use with ONNX runtime
sess = onnxruntime.InferenceSession(model_proto.SerializeToString())
res = sess.run([sess.get_outputs()[0].name], {'x': np.zeros([10, 100, 3], np.float32), 'y': np.zeros([10, 3, 5], np.float32)})
# 2022-11-18 16:40:50.3568383 [E:onnxruntime:, sequential_executor.cc:346 onnxruntime::SequentialExecutor::Execute] Non-zero status code returned while running Conv node. Name:'Conv' Status Message: Input channels C is not equal to kernel channels * group. C: 3 kernel channels: 3 group: -3
# ---------------------------------------------------------------------------
# Fail                                      Traceback (most recent call last)
# <ipython-input-268-ffc10888f6c3> in <module>
#      25 # Try to use with ONNX runtime
#      26 sess = onnxruntime.InferenceSession(model_proto.SerializeToString())
# ---> 27 res = sess.run([sess.get_outputs()[0].name], {'x': np.zeros([10, 100, 3], np.float32), 'y': np.zeros([10, 3, 5], np.float32)})
# 
# ~\Anaconda3\envs\tf\lib\site-packages\onnxruntime\capi\onnxruntime_inference_collection.py in run(self, output_names, input_feed, run_options)
#     186             output_names = [output.name for output in self._outputs_meta]
#     187         try:
# --> 188             return self._sess.run(output_names, input_feed, run_options)
#     189         except C.EPFail as err:
#     190             if self._enable_fallback:
# 
# Fail: [ONNXRuntimeError] : 1 : FAIL : Non-zero status code returned while running Conv node. Name:'Conv' Status Message: Input channels C is not equal to kernel channels * group. C: 3 kernel channels: 3 group: -3
```

**Screenshots**
NA

**Additional context**
The error seems to be in [`tf2onnx.onnx_opset.nn`](https://github.com/onnx/tensorflow-onnx/blob/v1.13.0/tf2onnx/onnx_opset/nn.py#L378-L384). The calculation of the `groups` attribute is done as follows:

```python
shape_dim = -1
if data_format == "NHWC":
    shape_dim = ctx.get_shape(node.input[0])[3]
elif data_format == "NCHW":
    shape_dim = ctx.get_shape(node.input[0])[1]
if shape_dim != -1:
    groups = int(shape_dim / ctx.get_shape(node.input[1])[2])
```

First the number of channels in `node.input[0]` is checked, depending on `data_format`, and then, only if that does not result in `-1`, it is divided by the number of input channels in `node.input[1]`. However, there is no check to verify that number of channels is actually valid, and not `-1`. I suppose the simple fix would be to just check that and leave the attribute as `1` if either of the two values is `-1`, but I thought I'd leave that decision to someone from the team instead of directly sending a PR.

javidcf:
Also minor change reading shape dimensions so it works for different dimensionalities.
Fixes #2084


javidcf:
I wrote this test case for this:

```python
@check_opset_min_version(11, "Pad")
def test_conv_unknown_kernel_channels(self):
    x_shape = [2, 10, 3]
    x_val = make_xval(x_shape)
    kernel_shape = [4, 3, 5]
    kernel_val = make_xval(kernel_shape)
    pad_val = np.array([[0, 0], [0, 0], [0, 0]], np.int64)
    def func(x, kernel, pad):
        # Make kernel dimensions unknown
        kernel = tf.pad(kernel, pad)
        conv = tf.nn.conv1d(x, kernel, stride=[1], padding='VALID')
        return tf.identity(conv, name='output')
    self._run_test_case(func, [_OUTPUT], {_INPUT: x_val, _INPUT1: kernel_val, _INPUT2: pad_val})
```

Unfortunately, in the process of doing so I found _another_ bug (already present before the fix) with convolution kernels with unknown shape. My fix solves the issue of the wrong value in the op `group` attribute, but attempting to run the above test function with ONNX (after my fix) results in this error:

```none
[ONNXRuntimeError] : 1 : FAIL : Non-zero status code returned while running Conv node. Name:'conv1d' Status Message: kernel_shape is not compatible with W shape. kernel_shape: {1,-1} W: {5,3,1,4}
```

The error can also be reproduced before my fix by simply adding the following line before the call to `tf.nn.conv1d`:

```python
x = tf.pad(x, pad)
```

The problem is I think this is actually an onnxruntime issue. [This line](https://github.com/microsoft/onnxruntime/blob/v1.14.0/onnxruntime/core/providers/cpu/nn/conv_attributes.h#L85) from onnxruntime seems to suggest the code is not properly checking that the dimensions are known (not -1), or maybe `kernel_shape_specified` should be `false` in that case...

Either way, I'm not sure if a different test should be added which somehow just checks the attribute value with a `graph_validator` without trying to run the operation (which I'm not sure if it's possible, since ops not used by the result get pruned) or leave the test as it is and keep this "on hold" until the other issue is resolved.

tf2onnx/onnx_opset/nn.py
```diff
@@ -311,8 +311,9 @@ def conv_kernel_shape(ctx, node, input_idx, spatial=2):
     # Get spatial part.
     kernel_shape = kernel_shape[:spatial]
 
-    # Set new value and return it.
-    node.set_attr("kernel_shape", kernel_shape)
+    # Set attribute value only if all dimensions are known.
+    if all(d > 1 for d in kernel_shape):
```
fatcat-z, line 315: Should it be "d > 0"?

tf2onnx/onnx_opset/nn.py
```diff
@@ -649,7 +652,8 @@ def version_1(cls, ctx, node, **kwargs):
             raise ValueError("input channel must be positive")
         k_output_channels = k_input_channels * k_channel_multiplier
 
-        node.set_attr("kernel_shape", [k_h, k_w])
+        if k_h >= 0 and k_w >= 0:
```
fatcat-z, line 655: Is zero a valid value for k_h or k_w? Would it be better to change to 'k_h > 0 and k_w > 0'?

tf2onnx/onnx_opset/nn.py
javidcf: Yes, thanks, that should be `d > 0` or even `d >= 0`. Depending on answer to next comment.

tf2onnx/onnx_opset/nn.py
javidcf: Good point, I'm not really sure. Zero-sized dimensions are valid in general, but I'm not sure it makes sense for convolution kernels. Interestingly, I just tried it in my TensorFlow environment (2.4.1) and a zero-sized convolution kernel makes hang (probably in an infinite loop) with CPU and log and error and crash with GPU.

fatcat-z: So should we change >= to > here?

javidcf: Yes, I have made both suggested changes.

fatcat-z:
Could you please resolve the conflict in test_backend.py file?

fatcat-z (approved):

PR #2117, opened 2023-02-09
Replace deprecated np.str with np.str_

jcwchen:
To make tf2onnx compatible with numpy>=1.24, replace deprecated np.str with np.str_


jcwchen:
Ah just saw there is PR already: https://github.com/onnx/tensorflow-onnx/pull/2117... Close this one.

Issue #2068, opened 2022-10-20
"compute_const_folding_using_tf" can result in infinite loop when converting well-formed model


christopherbate:
**Describe the bug**

When converting a model that has dynamic strided slice operations, the function `compute_const_folding_using_tf` can result in an infinite loop.

**System information**
- TensorFlow Version: Built from souce a couple days ago
- Python version: 3.8

**To Reproduce**

Unfortunately I cannot share the reproducer model, but it is a well-formed tensorflow saved model. The conversion succeeds if I manually disable the constant folding for the StridedSlice operation here https://github.com/onnx/tensorflow-onnx/blob/main/tf2onnx/tf_utils.py#L212. 

The original PR for this feature (https://github.com/onnx/tensorflow-onnx/pull/1145/files#diff-31e088e0e9d455d2164fd6251e7c23a77afc5561b3746cdf83e08d6a05142154) had very limited tests and did not include the dynamic bounds case.

fatcat-z:
It is totally understood that some models are not allowed to be shared. Is it possible to cut your scenario so that you can provide a small case which could possibly reproduce this issue?

christopherbate:
My workaround is to disable the strided slice constant folding and handle it elsewhere in my workflow. If I have additional time, I'll probably spend it directly troubleshooting. I think the best place to start is to upgrade the unit tests I mentioned above. Otherwise, I'll let you know if I find out a better root cause.

fatcat-z:
> My workaround is to disable the strided slice constant folding and handle it elsewhere in my workflow. If I have additional time, I'll probably spend it directly troubleshooting. I think the best place to start is to upgrade the unit tests I mentioned above. Otherwise, I'll let you know if I find out a better root cause.

Thanks for your efforts and share. Will definitely take a look to review related test case to see if there is a chance to repro your issue.

hgong-snap:
https://github.com/onnx/tensorflow-onnx/issues/1965

@fatcat-z   this is a similar issue  and it has a reproduce file.  can you also take a look?  thanks !

christopherbate:
Can you remove "pending user response"? I'm not planning on working on this anymore as I went with a different route.

hgong-snap:
@christopherbate   wondering any other workaround for this infinite loop?

fatcat-z:
> Can you remove "pending user response"? I'm not planning on working on this anymore as I went with a different route.

[...]

Issue #1965, opened 2022-06-13
Model conversion stuck in tf_utils.compute_const_folding_using_tf


hgong-snap:
**Describe the bug**
I tried to convert my model(graphdef) with the tf2onnx tool, but it takes forever to do the conversion. After debugging, it looks like the code stuck at `tf2onnx.tf_utils.compute_const_folding_using_tf`, particularly [this while-loop](https://github.com/onnx/tensorflow-onnx/blob/main/tf2onnx/tf_utils.py#L206)

**Urgency**
> If there are particular important use cases blocked by this or strict project-related timelines, please share more information and dates. If there are no hard deadlines, please specify none.

We are trying to experiment/benchmark `onnxruntime` in our service. So prefer to get it solved sooner.

**System information**
- MacOS
- Tensorflow Version: 2.7.0
- Python version:  3.8.13

**To Reproduce**
a [zip file](https://drive.google.com/file/d/1HAlfCtGkTUDwMpwdXrppZohqGyQK4s8_/view?usp=sharing).  run with `python convert.py`   (have to use google drive because 37MB > github max file size is 25MB)

**Screenshots**
```
Cannot infer shape for LookupTableExportV2_11: LookupTableExportV2_11:0,LookupTableExportV2_11:1
Cannot infer shape for LookupTableExportV2_12: LookupTableExportV2_12:0,LookupTableExportV2_12:1
Cannot infer shape for LookupTableExportV2_13: LookupTableExportV2_13:0,LookupTableExportV2_13:1
Cannot infer shape for LookupTableExportV2_14: LookupTableExportV2_14:0,LookupTableExportV2_14:1
Cannot infer shape for LookupTableExportV2_15: LookupTableExportV2_15:0,LookupTableExportV2_15:1
Cannot infer shape for LookupTableExportV2_16: LookupTableExportV2_16:0,LookupTableExportV2_16:1
Cannot infer shape for LookupTableExportV2_17: LookupTableExportV2_17:0,LookupTableExportV2_17:1
Cannot infer shape for LookupTableExportV2_18: LookupTableExportV2_18:0,LookupTableExportV2_18:1
```

after these warnings it takes forever.

hgong-snap:
[...]

PR #2119, opened 2023-02-15
Fix Infinite Loop in constant folding

ilyasher:
This commit fixes a bug in the function compute_const_folding_using_tf where an already-folded StridedSlice node could set the 'update' flag to True, causing an infinite loop.

A regression test was also added. The test checks for the infinite loop by setting a timeout using the timeout-decorator module from pypi.

This should fix issues #2068 and #1965 .


ilyasher:
@fatcat-z Could you please review when you get the chance? Thanks!

fatcat-z (approved):
Thanks for your contributions, this looks good to me!

PR #2126, opened 2023-02-23
A fix for seed attribute in the keras random normal generator

inonbe:
The seed attribute was not passing to the converted onnx model due to that it exists in "seed2" attribute instead of "seed" and its internal type is Integer and not float.
Currently there is no impact on existing unit_tests ("test_auto_encoder" for example) or need for new unit test
since still the random generator behaves differently on various engines.


tf2onnx/onnx_opset/generator.py
```diff
@@ -88,8 +88,8 @@ def version_9(cls, ctx, node, **kwargs):
         if node.inputs[0].is_const():
             cls.version_1(ctx, node, **kwargs)
         else:
-            seed = node.get_attr("seed")
-            node.set_attr("seed", float(seed.f))
+            seed = node.get_attr("seed2")
```
fatcat-z, line 91: Do we need to also update this for version_1(line 63)?

inonbe: thanks @fatcat-z you're quite right, ill update version_1 as well.

fatcat-z: I'm a little confused: is this behavior (using seed2 instead of seed) same among different TF versions? Did it change after one of tf version?

inonbe: As far as I've seen and tested, version_1 behaviour is also relevant for TF version.2 
when the random generator dim/shape are constant and not derived from the data 
batch size or any other dependent size.

I'm checking with TF V1 whether it was changed.

fatcat-z (approved):

PR #2133, opened 2023-03-09
support cumprod op

f-salvetti:
Supports [cumprod TF op](https://www.tensorflow.org/api_docs/python/tf/raw_ops/Cumprod). Both reverse and exclusive flags are supported.


tf2onnx/onnx_opset/math.py
```diff
@@ -555,6 +557,207 @@ def version_11(cls, ctx, node, **kwargs):
...
+                                     inputs=[graph.input_names[2], start_slice, end_slice, axis_node.output[0]])
+        if graph.get_dtype(slice_node.output[0]) != graph.get_dtype(one_node.output[0]):
+            pad_const_node = graph.make_node("Cast", inputs=[one_node.output[0]],
+                                             attr={"to": graph.get_dtype(slice_node.output[0])},
+                                             op_name_scope=loop_name, outputs=[utils.make_name("pad_const")])
+        else:
+            pad_const_node = one_node
+        pad_node = graph.make_node("Pad", inputs=[slice_node.output[0], pads_node.output[0], pad_const_node.output[0]],
+                                   op_name_scope=loop_name, outputs=[utils.make_name("pad")])
+        mul_node = graph.make_node("Mul", inputs=[graph.input_names[4], pad_node.output[0]],
```
fatcat-z, line 693: Is it possible to leverage reduce_prod to avoid multiplying those values one by one?

fatcat-z:
The 2 CI failures are not supposed to be relative to your changes. I will figure them out.

tf2onnx/onnx_opset/math.py
f-salvetti: sure, didn't think about that. I will refactor the code in order to use that operator

f-salvetti: @fatcat-z Actually, using ReduceProd, would require to loop over slices of the input tensor to be reduced and then concatenate results to create the output tensor. However, in order to support a non-constant axis for the CumProd op as in TF, I don't think it is possible to use the Concat op, since it requires the axis as an attribute. Due to this limitation, I don't see how I can use ReduceProd instead of Mul.

PR #2134, opened 2023-03-09
Fix the axis of inserted QDQ for ConvTranspose

sdua-nv:
Hi, I noticed a problem in the tf2onnx converter. This is regarding ConvTranspose converter (https://github.com/onnx/tensorflow-onnx/blob/main/tf2onnx/onnx_opset/nn.py#L428). When ConvTranspose is quantized, the converter should take a ConvTranspose layer and convert it as follows.
```
X --->  ConvTranspose
W -----^

now becomes

X ----Q_x -> DQ_x ----->  ConvTranspose
W --> Q_w -> DQ_w -> ----------^
```

These Q_w/DQ_w nodes that are inserted have the wrong axis set when they are modified by the tf2onnx converter. For ConvTranspose, the axis should be set to 1. This is the number of filters, which is Cout. (The kernel shape here is Cin x Cout x kH x kW)

For Conv, the axis should be (and is correctly) set to 0. (The kernel shape here is Cout x Cin x kH x kW)

Instead, this axis is unconditionally set to 0 for all conv layers that use the `conv_convert_inputs()` routine. 

I've provided a fix in this PR. I had some trouble with the tests I added so I needed to add a mechanism to skip checking shapes in `Tf2OnnxBackendTestBase.run_test_case`; it seems that no other test sets `check_shape=False` so this shouldn't change the behavior of anything else.

### DIFF
--- a/tests/backend_test_base.py
+++ b/tests/backend_test_base.py
@@ -284,7 +284,10 @@ class Tf2OnnxBackendTestBase(unittest.TestCase):
             # tflite sometimes converts from tf but produces an invalid model
             return None, None
 
-    def assert_shapes_correct(self, graph, allow_missing=False, run_checker=True):
+    def assert_shapes_correct(self, graph, allow_missing=False, run_checker=True, check_shape=True):
+        if not check_shape:
+            return None
+
         model_proto = graph.make_model("test")
 
         if run_checker and not any(graph.get_shape(out) is None for out in graph.outputs + graph.input_names):
@@ -402,8 +405,10 @@ class Tf2OnnxBackendTestBase(unittest.TestCase):
                     i = output_names_with_port.index(output_name)
                     actual[i] = np.transpose(actual[i], constants.NCHW_TO_NHWC)
 
-            self.assert_results_equal(expected, actual, rtol, atol, mtol, check_value, check_shape, check_dtype)
-            self.assert_shapes_correct(g, self.config.allow_missing_shapes, not self.config.skip_onnx_checker)
+            self.assert_results_equal(expected, actual, rtol, atol, mtol, check_value, check_shape,
+                                      check_dtype)
+            self.assert_shapes_correct(g, self.config.allow_missing_shapes, not self.config.skip_onnx_checker,
+                                       check_shape)
 
             if graph_validator:
                 self.assertTrue(graph_validator(g))
@@ -441,7 +446,8 @@ class Tf2OnnxBackendTestBase(unittest.TestCase):
                     onnx_tfl_res[i] = np.transpose(onnx_tfl_res[i], constants.NCHW_TO_NHWC)
 
             self.assert_results_equal(tfl_res, onnx_tfl_res, rtol, atol, mtol, check_value, check_shape, check_dtype)
-            self.assert_shapes_correct(g, self.config.allow_missing_shapes, not self.config.skip_onnx_checker)
+            self.assert_shapes_correct(g, self.config.allow_missing_shapes, not self.config.skip_onnx_checker,
+                                       check_shape)
 
             if graph_validator:
                 self.assertTrue(graph_validator(g))
@@ -475,7 +481,8 @@ class Tf2OnnxBackendTestBase(unittest.TestCase):
 
             self.assert_results_equal(tfjs_res, onnx_tfjs_res, rtol, atol, mtol, check_value, check_shape,
                                       check_dtype=False)
-            self.assert_shapes_correct(g, self.config.allow_missing_shapes, not self.config.skip_onnx_checker)
+            self.assert_shapes_correct(g, self.config.allow_missing_shapes, not self.config.skip_onnx_checker,
+                                       check_shape)
 
             if graph_validator:
                 self.assertTrue(graph_validator(g))
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -607,6 +607,44 @@ class BackendTests(Tf2OnnxBackendTestBase):
         self._run_test_case(func, [_OUTPUT], {_INPUT: x_val, _INPUT1: output_shape},
                             rtol=1e-05, process_args=process_args)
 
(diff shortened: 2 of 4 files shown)


fatcat-z:
The unit_test CI failure was not caused by your change, will fix it soon.

tests/test_backend.py
```diff
@@ -607,6 +607,42 @@ def func(x, output_shape_placeholder):
...
+    def test_conv2d_quantization_axis(self):
+        x_shape = [1, 1, 5, 5]
+        kernel_shape = _KERNEL3x3
+        strides = [1, 1, 1, 1]
+        x_val = make_xval(x_shape).transpose(NCHW_TO_NHWC)
+        kernel_val = make_xval(_KERNEL3x3)
+
+        def func(x):
+            f = tf.constant(kernel_val, name="kernel", dtype=tf.float32)
+            kernel_dq = tf.quantization.quantize_and_dequantize(f, 0, np.prod(kernel_shape))
```
fatcat-z, line 619: Please replace tf.quantization.quantize_and_dequantize with quantize_and_dequantize which was defined in the beginning.

tests/test_backend.py
```diff
@@ -607,6 +607,42 @@ def func(x, output_shape_placeholder):
...
+    def test_conv2d_transpose_quantization_axis(self):
+        x_shape = [2, 6, 4, 3]
+        output_shape = [2, 13, 9, 2]
+        kernel_shape = [3, 3, 2, 3]
+        strides = [1, 2, 2, 1]
+        x_val = make_xval(x_shape)
+        kernel_val = make_xval(kernel_shape)
+        def func(x):
+            f = tf.constant(kernel_val, name="kernel", dtype=tf.float32)
+            kernel_dq = tf.quantization.quantize_and_dequantize(f, 0, np.prod(kernel_shape))
```
fatcat-z, line 637: Please replace tf.quantization.quantize_and_dequantize with quantize_and_dequantize which was defined in the beginning.

tests/test_backend.py
```diff
@@ -607,6 +607,42 @@ def func(x, output_shape_placeholder):
         self._run_test_case(func, [_OUTPUT], {_INPUT: x_val, _INPUT1: output_shape},
                             rtol=1e-05, process_args=process_args)
 
+    def test_conv2d_quantization_axis(self):
```
fatcat-z, line 610: Please add an annotation for this function:

@check_opset_min_version(10, "quantize_and_dequantize")

tests/backend_test_base.py
```diff
@@ -284,7 +284,10 @@ def run_tflite(self, tflite_path, feed_dict):
             # tflite sometimes converts from tf but produces an invalid model
             return None, None
 
-    def assert_shapes_correct(self, graph, allow_missing=False, run_checker=True):
+    def assert_shapes_correct(self, graph, allow_missing=False, run_checker=True, check_shape=True):
+        if not check_shape:
```
fatcat-z, line 288: Could you please finish address my comments below and then disable this to see if the validation could pass?

If still fail, could you please share the error details so we can check if it could be corrected?

sdua-nv: Hi @fatcat-z , thanks for your review. I addressed your previous comments and tried disabling this. The error message is what I saw while trying to run the tests locally.

```
=================================== FAILURES ===================================
__________________ BackendTests.test_conv2d_quantization_axis __________________

self = <test_backend.BackendTests testMethod=test_conv2d_quantization_axis>

    @check_opset_min_version(10, "quantize_and_dequantize")
    def test_conv2d_quantization_axis(self):
        x_shape = [1, 1, 5, 5]
        kernel_shape = _KERNEL3x3
        strides = [1, 1, 1, 1]
        x_val = make_xval(x_shape).transpose(NCHW_TO_NHWC)
        kernel_val = make_xval(_KERNEL3x3)
    
        def func(x):
            f = tf.constant(kernel_val, name="kernel", dtype=tf.float32)
            kernel_dq = quantize_and_dequantize(f, 0, np.prod(kernel_shape))
            conv = tf.nn.conv2d(x, kernel_dq, strides=strides, padding="VALID")
            return tf.identity(conv, name=_TFOUTPUT)
        def graph_validator(g):
            return check_quantization_axis(g, "DequantizeLinear", 0)
    
>       self._run_test_case(func, [_OUTPUT], {_INPUT: x_val}, rtol=1e-05, graph_validator=graph_validator,
                            check_shape=False)

tests/test_backend.py:626: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/test_backend.py:177: in _run_test_case
    return self.run_test_case(func, feed_dict, [], output_names_with_port, **kwargs)
tests/backend_test_base.py:407: in run_test_case
    self.assert_shapes_correct(g, self.config.allow_missing_shapes, not self.config.skip_onnx_checker,
tests/backend_test_base.py:331: in assert_shapes_correct
    self.assertEqual(onnx_shape, tf2onnx_shape)
E   AssertionError: Lists differ: [1, 1, 3, 3] != [3, 3, 1, 1]
E   
E   First differing element 0:
E   1
E   3
E   
E   - [1, 1, 3, 3]
E   + [3, 3, 1, 1]
```

Essentially the reason why the shapes don't match here is that when the tf model is passed through tf2onnx, the kernel weights are transposed. That's part of the intended behavior of the converter. So this kind of shape mismatch is expected. This is why I think it is fine to ignore the shape check in this case. Let me know if that makes sense.

fatcat-z: Thanks for your details. Please re-enable your changes and let's submit the final changes.

sdua-nv: Great, I re-enabled the changes and the branch is now up to date.

fatcat-z (approved):

PR #2135, opened 2023-03-14
Fix None shape error when one input to ConcatV2 has a shape.

klasma:
I tried to convert a network where one of the inputs to a concatenation along dimension -1 had shape None. The other input did however have a shape. The conversion failed because the code only looked att the shape of the first input to determine what positive axis value to use in the concatenation. If the order of the inputs had been reversed, the conversion would have worked.

I have now changed the code to look at the shapes of both input nodes. With the new code, I can convert the network. I have also verified that the resulting onnx-file works.

Let me know if you would like me to change or add something.


fatcat-z:
Could you please add a unit test in test_backend.py to cover this scenario?

klasma:
@fatcat-z 

Sorry for the delay. I had to do some additional debugging in order to reproduce the problem in a test.

Now I have added a test for this scenario to test_backend.py.

In the test, a tensor with none shape is created by applying the slice operator to another tensor. The slice size is obtained by adding two tensors and therefore the shape of the sliced tensor cannot be inferred and is set to None. A tensor with a known shape is then concatenated to the tensor with none shape along the -1 axis. This is the node structure that caused problems in the network that I was trying to convert.

Let me know if you have any questions or if there is something else that I should do.

fatcat-z (approved):
LGTM, thanks for your contribution!

PR #2137, opened 2023-03-15
Update Resize (opset 11) layer to support scales option when dims are…

qant-um:
Resize layer (version 11) will generate automatically multiple layers using Shape, Slice, Concat whereas it supports also the scales and sizes options to realize this operation. In the case where input shape is undefined (-1), this method is interesting to be applied but otherwise, inference will be faster with only one node of Resize. Backend unit test are already efficient to cover this usecase.
reference: https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Resize-11

### DIFF
--- a/tf2onnx/onnx_opset/nn.py
+++ b/tf2onnx/onnx_opset/nn.py
@@ -1391,20 +1391,37 @@ class Resize:
         else:
             mode = "nearest"
         roi = ctx.make_const(utils.make_name("roi"), np.array([]).astype(np.float32))
-        const_zero = ctx.make_const(utils.make_name("const_zero"), np.array([0]).astype(np.int64))
-        const_two = ctx.make_const(utils.make_name("const_two"), np.array([2]).astype(np.int64))
-        const_empty_float = ctx.make_const(utils.make_name("const_empty_float"), np.array([]).astype(np.float32))
         input_nchw = ctx.make_node("Transpose", [node.input[0]], {"perm": constants.NHWC_TO_NCHW})
-        shape_input = ctx.make_node("Shape", [input_nchw.output[0]])
-        sliced_shape = ctx.make_node("Slice", [shape_input.output[0], const_zero.output[0], const_two.output[0]])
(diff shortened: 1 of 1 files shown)


tf2onnx/onnx_opset/nn.py
```diff
@@ -1391,20 +1391,37 @@ def version_11(cls, ctx, node, **kwargs):
...
-        size_int64 = ctx.make_node("Cast", [node.input[1]], attr={"to": onnx_pb.TensorProto.INT64})
-        concat_shape = ctx.make_node("Concat", [sliced_shape.output[0], size_int64.output[0]], {'axis': 0})
-        resize_inputs = [
-            input_nchw.output[0],
-            roi.output[0],
-            const_empty_float.output[0],
-            concat_shape.output[0]
-        ]
+        shape = ctx.get_shape(node.input[0])
+        if shape and shape[2] != -1 and shape[1] != -1 and node.inputs[1].is_const():
```
fatcat-z, line 1396: The shape comes from the input of Resize. Is it possible the rank is less than 3 so shape[2] doesn't exist?

qant-um: Hi,
This line is a copy of a condition of the version 10 ... but you are right, if this appends there will be an issue.
From my first understanding, shape is (B, C, H, W) so you mean the graph could be in another shape dimension. So, a condition should check that shape is on 4 dims before entering in this instruction block like this for example:
`if shape and len(shape) == 4 and and shape[2] != -1 and shape[1] != -1 and node.inputs[1].is_const()`
or "playing" with the array indexes:
`if shape and shape[-2] != -1 and shape[-1] != -1 and node.inputs[1].is_const()`

Thanks for your review again,

qant-um: Do you mean, this is ok for you ? I do not know exactly how the process works. Do you attend an action from myself ?
or the merge will be done by you once tests are passed ?

fatcat-z (approved):
Thanks for your contribution!

Issue #2102, opened 2023-01-08
The output of the model before and after the conversion is inconsistent.


enderdzz:
**Describe the bug**

I wrapped the single operator `AdjustContrastv2` in tensorflow 2.11 as a model and saved it as a frozen pb model file. At the same time I transformed the tf model by tf2onnx to get the onnx model, and the two have inconsistent results with large errors for the same input parameters.

**Urgency**

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 18.04*): Ubuntu 20.04.5 LTS
- TensorFlow Version: 2.11.0.dev
- Python version: 3.8.10
- ONNX version (if applicable, e.g. 1.11*): 1.12.0
- ONNXRuntime version (if applicable, e.g. 1.11*): 1.12.1

**To Reproduce**

[The link](https://drive.google.com/drive/folders/1lIZtT3sUmzWz8WAShXVe0xg1Sjwbf516?usp=sharing) shows the two models before(`tf.raw_ops.AdjustContrastv2_frozen_graph.pb`) and after(`tf.raw_ops.AdjustContrastv2_model.onnx`) the conversion, and the two input parameters of the model(`images.npy` and `contrast_factor.npy`). The `tf_save_model` dir is the model saved using `tf.saved_model.save`.

The following code is running two models separately, feeding them the same input, but with inconsistent results.
```py
import numpy as np
import tensorflow as tf
import onnxruntime as rt

images = np.load("images.npy")
contrast_factor = np.load("contrast_factor.npy")
onnx_model_path = "tf.raw_ops.AdjustContrastv2_model.onnx"
tf_model_path = "tf.raw_ops.AdjustContrastv2_frozen_graph.pb"

class OnnxModel():
    def __init__(self, onnx_path):
        self.onnx_session = rt.InferenceSession(onnx_path)
        self.input_name = self.get_input_name(self.onnx_session)
        self.output_name = self.get_output_name(self.onnx_session)

    def get_output_name(self, onnx_session):
        output_name = []
        for node in onnx_session.get_outputs():
            output_name.append(node.name)
        return output_name
 
    def get_input_name(self, onnx_session):
        input_name = []
        for node in onnx_session.get_inputs():
            input_name.append(node.name)
        return input_name
 
    def get_input_feed(self, input_name, image_numpy):
        i = 0
        input_feed = {}
        for name in input_name:
            input_feed[name] = image_numpy[i]
            i += 1
        return input_feed
 
[...]

PR #2140, opened 2023-03-18
Fix issue #2102: Set reduction axis of mean to height and width for tf adjust_contrast op

cosineFish:
Refer to https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/image/adjust_contrast_op.h

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -3280,13 +3280,13 @@ class BackendTests(Tf2OnnxBackendTestBase):
         self._run_test_case(func, [_OUTPUT], {_INPUT: x_val, _INPUT1: x_new_size})
 
     def test_adjust_contrast(self):
-        x_shape = [4, 3, 2]
-        x_val = np.arange(1, 1 + np.prod(x_shape), dtype=np.float32).reshape(x_shape)
-        y_val = np.array(2.1, np.float32)
         def func(x, y):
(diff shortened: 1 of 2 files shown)


tests/test_backend.py
```diff
@@ -3280,13 +3280,13 @@ def func(x, x_new_size_):
...
 
     def test_adjust_contrast(self):
-        x_shape = [4, 3, 2]
-        x_val = np.arange(1, 1 + np.prod(x_shape), dtype=np.float32).reshape(x_shape)
-        y_val = np.array(2.1, np.float32)
         def func(x, y):
             x_ = tf.image.adjust_contrast(x, y)
             return tf.identity(x_, name=_TFOUTPUT)
-        self._run_test_case(func, [_OUTPUT], {_INPUT: x_val, _INPUT1: y_val})
+        for x_shape in [[4, 3, 2], [2, 3, 4, 5], [3, 4, 2, 4, 3]]:
```
fatcat-z, line 3286: Is [3, 4, 2, 4, 3] a valid input for adjust_contrast() function? Ideally, the rank of image should be 4, right?

cosineFish: It's a valid input. This function only requests input at least 3d, but doesn't require input to be 3d or 4d only.
https://www.tensorflow.org/api_docs/python/tf/image/adjust_contrast

Issue #2148, opened 2023-03-28
Equal 11 problem with input type of string


mikeessen:
**Describe the bug**

ONNX Equal 11 does not support a type of string, but Equal version 11 in tf2onnx/onnx_opset/logical.py is implemented to support all types including string.

See https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Equal-11 for supported types.

Note that the ONNX documentations shows that Equal-19 will support string.  But, Equal version 11 in tf2onnx should still work with ONNX Equal 11.

**Urgency**

An update to tf2onnx/onnx_opset/logical.py is created to work around and a PR with the update will be submitted.

**System information**
- OS Platform and Distribution: macOS 13.2.1
- TensorFlow Version: 2.10.0
- Python version: 3.10.9
- ONNX version : 1.13.1
- ONNXRuntime: 1.13.1

PR #2149, opened 2023-03-28
Change Equal 11 for string input

mikeessen:
This pull request fixes issue #2148 which is that ONNX Equal 11 does not support a type of string, but Equal version 11 in tf2onnx/onnx_opset/logical.py is implemented to support all types including string.

### DIFF
--- a/tf2onnx/onnx_opset/logical.py
+++ b/tf2onnx/onnx_opset/logical.py
@@ -45,6 +45,7 @@ def _add_cast_to_same_type_to_inputs(graph, node, supported_dtypes, target_dtype
                 const_node = graph.get_node_by_output(inp)
                 const_node.set_tensor_value(utils.np.array(-1))
 
+
 @tf_op("LogicalNot", onnx_op="Not")
 class DirectOp:
     @classmethod
--- a/tf2onnx/onnx_opset/logical.py
+++ b/tf2onnx/onnx_opset/logical.py
@@ -29,16 +29,18 @@ def _add_cast_to_inputs(graph, node, supported_dtypes, target_dtype):
             graph.copy_shape(inp, inp_cast.output[0])
             graph.set_dtype(inp_cast.output[0], target_dtype)
 
-
-def _add_cast_to_same_type_to_inputs(graph, node):
-    common_dtype = graph.get_dtype(node.input[0])
-
-    for inp in node.input[1:]:
-        if graph.get_dtype(inp) != common_dtype:
-            inp_cast = graph.insert_new_node_on_input(node, "Cast", inp, to=common_dtype)
-            graph.copy_shape(inp, inp_cast.output[0])
-            graph.set_dtype(inp_cast.output[0], common_dtype)
-
+def _add_cast_to_same_type_to_inputs(graph, node, supported_dtypes, target_dtype):
+    for inp in node.input:
+        if graph.get_dtype(inp) not in supported_dtypes:
+            if graph.is_const(inp) and graph.get_tensor_value(inp) == '':
+                # Convert '' string constant to -1 int
+                # https://github.com/tensorflow/tensorflow/blob/4e7f0185c70faf35e12acbfe381a729d1e6cc38c/tensorflow/python/feature_column/feature_column.py#L2286
+                const_node = graph.get_node_by_output(inp)
+                const_node.set_tensor_value(utils.np.array(-1))
+            else:
+                inp_cast = graph.insert_new_node_on_input(node, "Cast", inp, to=target_dtype)
+                graph.copy_shape(inp, inp_cast.output[0])
+                graph.set_dtype(inp_cast.output[0], target_dtype)
 
 @tf_op("LogicalNot", onnx_op="Not")
 class DirectOp:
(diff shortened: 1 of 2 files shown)


mikeessen:
@fatcat-z Is there a way to restart the pipeline tests?  For the Equal 11 string update PR there are 2 failing pipeline tests which I think may not have to do with the changes in the PR.

mikeessen:
@fatcat-z please let me know if there are any changes  or updates I should make to the PR.  Thank you!

fatcat-z:
Could you please also add a test (like [this](https://github.com/onnx/tensorflow-onnx/blob/8f8d49a88656a334ec1f4c000552cbdce211a8ac/tests/test_backend.py#L5782)) for this change?

tf2onnx/onnx_opset/logical.py
```diff
@@ -29,16 +29,18 @@ def _add_cast_to_inputs(graph, node, supported_dtypes, target_dtype):
...
-
-    for inp in node.input[1:]:
-        if graph.get_dtype(inp) != common_dtype:
-            inp_cast = graph.insert_new_node_on_input(node, "Cast", inp, to=common_dtype)
-            graph.copy_shape(inp, inp_cast.output[0])
-            graph.set_dtype(inp_cast.output[0], common_dtype)
-
+def _add_cast_to_same_type_to_inputs(graph, node, supported_dtypes, target_dtype):
+    for inp in node.input:
+        if graph.get_dtype(inp) not in supported_dtypes:
```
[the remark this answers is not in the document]
mikeessen: Thanks, I will add logic to unify the dtype of all of inputs.

tf2onnx/onnx_opset/logical.py
```diff
@@ -29,16 +29,21 @@ def _add_cast_to_inputs(graph, node, supported_dtypes, target_dtype):
             graph.copy_shape(inp, inp_cast.output[0])
             graph.set_dtype(inp_cast.output[0], target_dtype)
 
-
-def _add_cast_to_same_type_to_inputs(graph, node):
+def _add_cast_to_same_type_to_inputs(graph, node, supported_dtypes, target_dtype):
     common_dtype = graph.get_dtype(node.input[0])
+    if common_dtype not in supported_dtypes:
```
fatcat-z, line 34: My bad, I misunderstood the issue before.

Actually, the issue is: The op Equal doesn't support 'string' with version 11, meaning we should fail the conversion once we detect this case. Even the content is empty, we should not convert it to an integer value which might confuse users.

So it'd better to set up an unsupported op list which may only contain string right now. If we detect current type of input[0] is in it, we fail the conversion with a reasonable message instead of making tricky things to work around it.

Make sense?

cjvolzka: > Actually, the issue is: The op Equal doesn't support 'string' with version 11, meaning we should fail the conversion once we detect this case.

In the OpSet 7 version of this, the Equal Op works fine assuming the strings can be converted to the desired type. So this change makes OpSet11 match the OpSet7 behavior. 

So if we made this an error, then OpSet 7 Equal and Opset 11 would behave differently but both have the same stance on strings. Making them behave different didn't seem right. Rather than remove what was working in OpSet 7, we decided to enable the same setup for OpSet 11. It could be argued to make OpSet7 (and earlier) also automatically error in this case instead. If that's truely preferred we could look into it. However our thought was to expand 11 rather than restrict 7.

> Even the content is empty, we should not convert it to an integer value which might confuse users.

I'll agree that the empty string was a special case that almost lead us to restrict OpSet 7. However after digging into the TF code, we found they explicitly handle this case. When creating a feature column, if an entry is `""` then it's explicitly changed to `-1` (https://github.com/tensorflow/tensorflow/blob/4e7f0185c70faf35e12acbfe381a729d1e6cc38c/tensorflow/python/feature_column/feature_column.py#L2286). Since it's explicitly handled, we matched the TF behavior over just erroring out. Otherwise it is also confusing to have a model work in TF but fail to convert.

For reference we have a model encountering this scenario that indeed works fine in TF but then fails to convert. With these changes, it converts fine. Unfortunately it's not a model we can share but it is a real world scenario.

tf2onnx/onnx_opset/logical.py
```diff
@@ -29,16 +29,21 @@ def _add_cast_to_inputs(graph, node, supported_dtypes, target_dtype):
...
         if graph.get_dtype(inp) != common_dtype:
             inp_cast = graph.insert_new_node_on_input(node, "Cast", inp, to=common_dtype)
             graph.copy_shape(inp, inp_cast.output[0])
             graph.set_dtype(inp_cast.output[0], common_dtype)
-
+            if graph.is_const(inp) and graph.get_tensor_value(inp) == '':
+                # Convert '' string constant to -1 int
+                # https://github.com/tensorflow/tensorflow/blob/4e7f0185c70faf35e12acbfe381a729d1e6cc38c/tensorflow/python/feature_column/feature_column.py#L2286
+                const_node = graph.get_node_by_output(inp)
+                const_node.set_tensor_value(utils.np.array(-1))
```
fatcat-z, line 46: Please insert one more blank line for the format.

fatcat-z, line 46:
```suggestion
                const_node.set_tensor_value(utils.np.array(-1))

```

tf2onnx/onnx_opset/logical.py
fatcat-z: Thanks for your detailed explanations.

fatcat-z (approved):

Issue #2088, opened 2022-11-30
Input channels C is not equal to kernel channels * group when BN layer's axis=-2


maybeLee:
**Describe the bug**
I was constructing a three-layer keras model using the following script
```
import keras
# This is the buggy model built from scratch
layer_stack = [
  keras.layers.DepthwiseConv2D(kernel_size=(3,3), strides=(2,2)),
  keras.layers.BatchNormalization(axis=-2),  # if we set axis=-1, the bug will not be manifested
  keras.layers.ReLU(max_value=6)
]
keras_input = keras.layers.Input((6,6,512))
layer_input = keras_input
for layer in layer_stack:
  y = layer(layer_input)
  layer_input = y
new_model = keras.models.Model(keras_input, y)
new_model.summary()
new_model.save("generated_model.h5")
```

This model can be successfully constructed and do model prediction using keras. But when I convert it to onnx model and making inference with it, it will raise a runtime error.

```
import tensorflow as tf
import tf2onnx
def transform_onnx(model_path: str):
    model = keras.models.load_model(model_path)
    model_name = model_path.split(".h5")[0]
    onnx_path = f"{model_name}.onnx"
    input_shape = model.layers[0].input_shape[0]
    spec = (tf.TensorSpec(input_shape, tf.float32, name="input"),)
    _, _ = tf2onnx.convert.from_keras(model, input_signature=spec, \
        opset=15, output_path=onnx_path)
transform_onnx("generated_model.h5")

import onnxruntime as ort
onnx_path = "generated_model.onnx"
onnx_model = ort.InferenceSession(onnx_path, providers=["CPUExecutionProvider"])
input_name = onnx_model.get_inputs()[0].name
output_name = onnx_model.get_outputs()[0].name
input = input.astype('float32')
onnx_pred = onnx_model.run([output_name], {input_name: input})[0]
```

Error Message:
```
---------------------------------------------------------------------------
Fail                                      Traceback (most recent call last)
[<ipython-input-27-2745ef9ee442>](https://8h48y5r9dcw-496ff2e9c6d22116-0-colab.googleusercontent.com/outputframe.html?vrz=colab-20221128-060047-RC00_491313853#) in <module>
      5 output_name = onnx_model.get_outputs()[0].name
      6 input = input.astype('float32')
----> 7 onnx_pred = onnx_model.run([output_name], {input_name: input})[0]

[...]

PR #2152, opened 2023-04-01
Fix fusing of Conv and Mul in TransposeOptimizer (#2088)

cosineFish:
Should not fuse Conv and Mul if the constant input of Mul is not with shape [N] or [1,..,1,N].

### DIFF
--- a/tests/test_optimizers.py
+++ b/tests/test_optimizers.py
@@ -1187,6 +1187,41 @@ class OptimizerTests(Tf2OnnxBackendTestBase):
                                    model_proto, remaining_transpose_num=0)
 
     @parameterized.expand([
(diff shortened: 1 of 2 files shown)


fatcat-z:

Could you please add some details about what kind of issues will be caused without this change?

cosineFish:
> Thanks for your contributions!
> 
> Could you please add some details about what kind of issues will be caused without this change?

Sure. 
The reason why this change is needed is from the implementation of conv, weights of conv is only linearly related with output channels(bias is required to be not set for _mul_handler):
image

If perm of the transpose node before mul node is NCHW_TO_NHWC(the most possible case), then the multipler can be merged into weights of conv only if its shape is in (1...1, Cout).

Without this change, incorrect fusing will be made and issues like #2088 will be caused. The reason why the test in #2088 fails is that after fusing conv+transpose+mul, weights of conv is incorrectly changed from shape [512,1,3,3] to [512,2,3,3].

fatcat-z (approved):
LGTM, thanks!

PR #2159, opened 2023-04-12
Update tf_loader.py

DLumi:
freeze_session(): removed unnecessary slices which literally do nothing if we take the first element anyway


fatcat-z (approved):

Issue #436, opened 2019-04-02
ValueError: tensorflow op ExtractImagePatches is not supported


ayermolo:
I am trying to convert yolov2-voc generated by DarkFlow in to ONNX, and running in to this error
"ValueError: tensorflow op ExtractImagePatches is not supported"

I am using tensorflow=1.9.0, onnx=1.4.1, opset=7, tfonnx=1.5.0/9fcb06.

Is this tensorflow issue, or the converter? How would I got about fixing this?

Thank You

nbcsm:
currently, for TF ExtractImagePatches op, there is no equivalent op in onnx.
would need to propose and add such op into future opset.

ayermolo:
@nbcsm Thank you for reply. What would be the process for that?

nbcsm:
go to https://github.com/onnx/onnx, and open an issue for ExtractImagePatches

ayermolo:
Great, thanks

prasanthpul:
@nbcsm can it be composed in the converter?

ayermolo:
Along the same lines. Translated in to something else in special cases:
I think ExtractImagePatch with ksizes = [1, blk, blk, 1], stride = [1, blk, blk, 1], rates = [1, 1, 1, 1]  is equivalent to SpaceToDepth with blocksize = blk, blk.

nbcsm:
for that case (ExtractImagePatch with ksizes = [1, blk, blk, 1], stride = [1, blk, blk, 1], rates = [1, 1, 1, 1]), yes, SpaceToDepth seems to work.
but it may not work for other ExtractImagePatch cases.

it feels like we may use MaxPool + SpaceToDepth to convert ExtractImagePatch, but need to double check.

xtremertx:
Hi, I having the same issue here trying to convert Yolov2 full network into ONNX and I really need ExtractImagePatch, I was trying to use tensorflow model instead but it doesnt worked, as ML:NET failed to load it. Please add support for this or at least some workaround code?

xtremertx:
Hi can someone just paste here the ExtractImagePatch handler that is using SpaceToDepth ? I'm new in this area and writing own operations seems quite hard to me, thanks a lot.

ayermolo:
Are you using DarkFlow?
It's pretty straightforward: https://github.com/ayermolo/darkflow/commit/c62d503493d2db6f3d94d106a60b2726a4eda47a

xtremertx:
[...]

PR #2188, opened 2023-06-16
feat: add support for ExtractImagePatches

nanoskript:
Closes: https://github.com/onnx/tensorflow-onnx/issues/436

This rewrite is based on this comment: https://github.com/onnx/tensorflow-onnx/issues/436#issuecomment-993313423 with changes to make it more general and translatable into `tf2onnx`.

<details>
<summary>Equivalent TensorFlow function and automated test script (expand)</summary>

```python3
import tensorflow as tf
import numpy as np
from hypothesis import given, strategies as st, settings, assume


def our_extract_image_patches(sizes, strides, rates, padding):
    # TensorFlow's constraints.
    assert sizes[0] == 1 and sizes[3] == 1
    assert strides[0] == 1 and strides[3] == 1
    assert rates[0] == 1 and rates[3] == 1
    assert padding in ["SAME", "VALID"]

    # Extract size.
    [_, size_rows, size_cols, _] = sizes

    @tf.function
    def function(tensor):
        # Input shape of [N, H, W, C].
        tensor_shape = tensor.shape

        # Transpose and reshape to [N * C, H, W, 1].
        tensor = tf.transpose(tensor, perm=[0, 3, 1, 2])
        tensor = tf.reshape(tensor, [
            tensor_shape[0] * tensor_shape[3],
            tensor_shape[1],
            tensor_shape[2],
            1,
        ])

        # Convolve with identity kernel into [N * C, ?H, ?W, K].
        k = size_rows * size_cols
        kernel = tf.reshape(tf.eye(k), [size_rows, size_cols, 1, k])
        convolution = tf.nn.conv2d(tensor, kernel, strides=strides, padding=padding, dilations=rates)

        # Reshape into [N, C, ?H, ?W, K].
        reshaped = tf.reshape(convolution, [
            tensor_shape[0],
            tensor_shape[3],
            convolution.shape[1],
            convolution.shape[2],
            k,
        ])

        # Transpose and reshape into [N, ?H, ?W, C * K].
        patches = tf.transpose(reshaped, perm=[0, 2, 3, 4, 1])
        return tf.reshape(patches, [
            tensor_shape[0],
            convolution.shape[1],
            convolution.shape[2],
            tensor_shape[3] * k,
        ])

    return function


def tf_extract_image_patches(sizes, strides, rates, padding):
    @tf.function
    def function(tensor):
        return tf.image.extract_patches(
            tensor,
            sizes=sizes,
            strides=strides,
            rates=rates,
            padding=padding,
        )

    return function


@settings(max_examples=5000)
@given(
    st.lists(st.integers(min_value=1, max_value=20), min_size=4, max_size=4),
    st.integers(min_value=1, max_value=20),
    st.integers(min_value=1, max_value=20),
    st.integers(min_value=1, max_value=20),
    st.integers(min_value=1, max_value=20),
    st.integers(min_value=1, max_value=20),
    st.integers(min_value=1, max_value=20),
    st.sampled_from(["VALID", "SAME"]),
)
def test_equal(shape, size_rows, size_cols, stride_rows, stride_cols, dil_rows, dil_cols, padding):
    sizes = [1, size_rows, size_cols, 1]
    strides = [1, stride_rows, stride_cols, 1]
    rates = [1, dil_rows, dil_cols, 1]

    try:
        tensor = tf.cast(tf.reshape(tf.range(np.prod(shape)), shape), dtype=tf.float32)
        tfs = tf_extract_image_patches(sizes, strides, rates, padding)(tensor)

        if 0 in tfs.shape:
            # We cannot handle operations that produce empty outputs.
            assume(False)
    except ValueError:
        # Ignore input if TensorFlow would fail.
        assume(False)
        return

    ours = our_extract_image_patches(sizes, strides, rates, padding)(tensor)
    assert tf.reduce_all(tf.math.equal(tfs, ours)).numpy()
```
</details>

Output from `pytest convolve.py --hypothesis-show-statistics` (no failures):
```
convolve.py::test_equal:

  - during generate phase (70.74 seconds):
    - Typical runtimes: ~ 1-14 ms, of which < 1ms in data generation
    - 5000 passing examples, 0 failing examples, 3913 invalid examples

  - Stopped because settings.max_examples=5000
```

### DIFF
--- a/tf2onnx/onnx_opset/nn.py
+++ b/tf2onnx/onnx_opset/nn.py
@@ -2105,7 +2105,19 @@ class ExtractImagePatches:
         rates = node.get_attr_value("rates")
         padding = node.get_attr_str("padding")
 
-        # Our constraints.
+        # This implementation of ExtractImagePatches does not generalize
+        # to outputs that are empty. For example:
+        #
+        #   tf.image.extract_patches(
+        #     np.random.rand(1, 1, 1, 1), sizes=[1, 2, 2, 1], strides=[1, 1, 1, 1],
+        #     rates=[1, 1, 1, 1], padding="VALID"
+        #   )
+        #
+        # succeeds with the output of:
+        #
+        #   <tf.Tensor: shape=(1, 0, 0, 4), dtype=float64, numpy=array([], shape=(1, 0, 0, 4), dtype=float64)>
+        #
+        # whereas attempting the same here results in an "Invalid input shape" error for the "Conv" node.
         utils.make_sure(0 not in output_shape, "Empty ExtractImagePatches output is unsupported.")
         [_, size_rows, size_cols, _] = sizes
 
--- a/tf2onnx/onnx_opset/nn.py
+++ b/tf2onnx/onnx_opset/nn.py
@@ -2091,3 +2091,81 @@ class CTCGreedyDecoder:
         ctx.replace_all_inputs(node.output[3], sum_max_neg)
 
         ctx.remove_node(node.name)
+
+
+@tf_op("ExtractImagePatches")
+class ExtractImagePatches:
+    @classmethod
+    def version_9(cls, ctx, node, **kwargs):
+        input_shape = ctx.get_shape(node.input[0])
+        output_shape = node.output_shapes[0]
+
+        sizes = node.get_attr_value("ksizes")
+        strides = node.get_attr_value("strides")
+        rates = node.get_attr_value("rates")
+        padding = node.get_attr_str("padding")
+
+        # Our constraints.
+        utils.make_sure(0 not in output_shape, "Empty ExtractImagePatches output is unsupported.")
+        [_, size_rows, size_cols, _] = sizes
+
+        # Transform input into [N * C, H, W, 1].
+        transformed_input = ctx.make_node("Reshape", inputs=[
+            ctx.make_node("Transpose", inputs=node.input, attr=dict(perm=[0, 3, 1, 2])).output[0],
+            ctx.make_const(utils.make_name("new_shape"), np.int64([
+                input_shape[0] * input_shape[3],
+                input_shape[1],
+                input_shape[2],
+                1,
+            ])).output[0],
+        ])
+
+        # Create identity kernel.
+        k = size_rows * size_cols
+        identity_kernel = ctx.make_node("Reshape", inputs=[
+            ctx.make_node("EyeLike", inputs=[
+                ctx.make_node("ConstantOfShape", inputs=[
+                    ctx.make_const(utils.make_name("eye_size"), np.array([k, k], dtype=np.int64)).output[0],
+                ]).output[0],
+            ]).output[0],
+            ctx.make_const(utils.make_name("new_shape"), np.array([
+                size_rows,
+                size_cols,
+                1,
+                k,
+            ], dtype=np.int64)).output[0],
+        ])
+
+        # Construct placeholder convolution node and transform into [N * C, K, ?H, ?W].
+        convolution = ctx.make_node("Conv", inputs=[transformed_input.output[0], identity_kernel.output[0]],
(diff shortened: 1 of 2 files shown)


fatcat-z:
Rewriter is designed to rewrite the ONNX graph after we transform each tf op into the corresponding onnx op. Each rewriter will search the ONNX graph following a given pattern. Once the pattern is matched, those involved onnx ops will be replaced with some other ops for an optimization in further inference.

In this case, ExtractImagePatches is just a tf op which is not supported by tf2onnx yet. So your implementations should be put into nn.py file instead of adding a rewriter. Please add it into nn.py, just like adding a new tf op support.

nanoskript:
Hi @fatcat-z,

> Rewriter is designed to rewrite the ONNX graph after we transform each tf op into the corresponding onnx op.

I'm not entirely sure if this is true. From my understanding, the rewriters are ran before each operation is converted into an ONNX operation:

https://github.com/onnx/tensorflow-onnx/blob/01520291227d46c48615300b0a73436dbe3c6610/tf2onnx/tfonnx.py#L616
where line 622 performs the conversion (?). There do appear to be late rewriters that run after the mapping occurs, but in general, it seems like the rewriting and optimization steps are separate.

I chose to implement this as a rewrite in order to avoid duplicating the construction of the Conv2D node but if you would still prefer for this to be implemented in `nn.py`, please let me know.

fatcat-z:
> Hi @fatcat-z,
> 
> > Rewriter is designed to rewrite the ONNX graph after we transform each tf op into the corresponding onnx op.
> 
> I'm not entirely sure if this is true. From my understanding, the rewriters are ran before each operation is converted into an ONNX operation:
> 
> https://github.com/onnx/tensorflow-onnx/blob/01520291227d46c48615300b0a73436dbe3c6610/tf2onnx/tfonnx.py#L616
> 
> 
> where line 622 performs the conversion (?). There do appear to be late rewriters that run after the mapping occurs, but in general, it seems like the rewriting and optimization steps are separate.
> I chose to implement this as a rewrite in order to avoid duplicating the construction of the Conv2D node but if you would still prefer for this to be implemented in `nn.py`, please let me know.

No, [graphs_from_tf() function](https://github.com/onnx/tensorflow-onnx/blob/01520291227d46c48615300b0a73436dbe3c6610/tf2onnx/tfonnx.py#L459C1-L460C1) will transfer the tf graph to onnx graph meaning each tf op has been converted to onnx op, if possible. Afterwards, process_parsed_graph() will be called to finish those rewriters and optimizations.

Yes, please implement this as an op in nn.py instead of creating a new rewriter. Thanks.

ruihu102:
Is this new operator going to be merged into main?

nanoskript:
> Is this new operator going to be merged into main?

I don't think this operator can be considered to be new but I'm aiming to get the requested changes done sometime within the week.

nanoskript:
@fatcat-z Hi! Apologies for the ping! Do you think this PR could be merged into main at some time?

tf2onnx/onnx_opset/nn.py
```diff
@@ -2091,3 +2091,81 @@ def version_11(cls, ctx, node, **kwargs):
...
+    def version_9(cls, ctx, node, **kwargs):
+        input_shape = ctx.get_shape(node.input[0])
+        output_shape = node.output_shapes[0]
+
+        sizes = node.get_attr_value("ksizes")
+        strides = node.get_attr_value("strides")
+        rates = node.get_attr_value("rates")
+        padding = node.get_attr_str("padding")
+
+        # Our constraints.
```
fatcat-z, line 2108: Could you please provide more details about this constraint so people know how to improve in future?

nanoskript: I've expanded this comment and given an example of a call that succeeds in Tensorflow but fails for this particular implementation.

fatcat-z (approved):
LGTM, thanks!

PR #2195, opened 2023-06-30
Support unique_with_counts

f-salvetti:
Add mapping for tf.unique_with_counts

### DIFF
--- a/tf2onnx/onnx_opset/tensor.py
+++ b/tf2onnx/onnx_opset/tensor.py
@@ -2366,8 +2366,9 @@ class ReverseV2:
 @tf_op("Unique", onnx_op="Unique")
 @tf_op("UniqueWithCounts", onnx_op="Unique")
 class Unique:
-    int_cast =  [TensorProto.BOOL, TensorProto.INT32, TensorProto.INT16, TensorProto.UINT8, TensorProto.UINT16, TensorProto.UINT32, TensorProto.UINT64]
-    dtype_map = {k:TensorProto.INT64 for  k in int_cast}
+    int_cast = [TensorProto.BOOL, TensorProto.INT32, TensorProto.INT16, TensorProto.UINT8,
+                TensorProto.UINT16, TensorProto.UINT32, TensorProto.UINT64]
+    dtype_map = {k: TensorProto.INT64 for k in int_cast}
     dtype_map[TensorProto.DOUBLE] = TensorProto.FLOAT
 
     @classmethod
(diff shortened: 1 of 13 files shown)


tf2onnx/onnx_opset/tensor.py
```diff
@@ -2364,32 +2364,45 @@ def version_10(cls, ctx, node, **kwargs):
...
-                                                          to=dtypes[1])
-                ctx.set_dtype(cast_node.output[0], dtypes[1])
-                ctx.copy_shape(new_node.output[2], cast_node.output[0])
+        if len(node_outputs) == 3: # we need counts too (UniqueWithCounts)
+            ctx.replace_all_inputs(node_outputs[2], new_node.output[3])
+        if ctx.get_dtype(new_node.output[0]) != inp_dtype:
+            ctx.insert_new_node_on_output("Cast", new_node.output[0], to=inp_dtype,
+                                          name=utils.make_name(node.name) + "_cast")
+
+        # cast idx and counts if needed
```
fatcat-z, line 2400: Where do we cast 'counts'?

tf2onnx/onnx_opset/tensor.py
```diff
@@ -2364,32 +2364,45 @@ def version_10(cls, ctx, node, **kwargs):
...
+        out_dtype = node.get_attr_value('out_idx')
+        if out_dtype != TensorProto.INT64:
```
fatcat-z, line 2402: Since out_idx is optional, what will happen here if the output of Unique onnx op contains only 1 element? Line 2402 may still be true, but it is not what we expect, right?

I agree accessing these attributes by a name is more meaningful, but probably it is more safe to access them by output lenght and index. Thoughts?

f-salvetti: According to [TF API ](https://www.tensorflow.org/versions/r2.11/api_docs/python/tf/raw_ops/UniqueWithCounts) 'out_idx' should be either INT32 or INT64, and has a default value set to INT32, so I would expect that the attribute is always set to one of the two dtypes. And since onnx Unique op always provides "inverse_indices" and "counts" as INT64 they should be casted if 'out_idx' is different from INT64.

tf2onnx/onnx_opset/tensor.py
f-salvetti: In the the for loop (Line 2403). If the node has 2 outputs (TF "Unique" op) we cast onnx  output 2 ("inverse_indices"). If the node has 3 outputs (TF "UniqueWithCounts" op) we cast onnx output 2 ("inverse_indices") and output 3 ("counts"). Either case, we skip onnx output 1 ("indices") since it is not present  in the TF ops.

fatcat-z (approved):
LGTM, thanks!

PR #2203, opened 2023-07-10
Fix wrong shapes in loop body inputs if shape invariances are set in TF

### LINKED ISSUE #2202: While loop wrong input shape with TF shape invariance
**Describe the bug**
When converting a TF While node, if the the while loop has some shape invariance and the initial tensor has fully defined shape, the converter assumes the first shape as fixed, causing wrong inference behavior (infinite loop).

**To Reproduce**
```python
class Model(tf.Module):
    @tf.function(input_signature=[
        tf.TensorSpec(shape=(1,), dtype=tf.int32, name="x"),
    ])
    def call(self, x):
        const = tf.constant(np.array([2], dtype=np.int32))
        c = lambda x: tf.reduce_all(tf.shape(x) < 10)
        b = lambda x: tf.concat([x, const], 0)
        r = tf.while_loop(c, b, [x], shape_invariants=[tf.TensorShape([None])])
        return tf.identity(r, name="output")
        
    def __call__(self, *args, **kwargs):
        return self.call(*args, **kwargs)

model = Model()
tf.saved_model.save(model, export_dir="test_model", signatures=model.call.get_concrete_function())
```

When called with onnxruntime it enters and infinite loop and warns:

```python
2023-07-10 11:01:43.855159751 [W:onnxruntime:, execution_frame.cc:835 VerifyOutputSizes] Expected shape from model of {2} does not match actual shape of {3} for output while/concat:0
2023-07-10 11:01:43.855182551 [W:onnxruntime:, execution_frame.cc:835 VerifyOutputSizes] Expected shape from model of {2} does not match actual shape of {4} for output while/concat:0
```
and so on

**Screenshots**
image

ONNX assumes the input to be of fixed (1,) shape instead of (-1,)

f-salvetti:
Fix #2202

### DIFF
--- a/tests/test_loops.py
+++ b/tests/test_loops.py
@@ -286,15 +286,12 @@ class LoopTests(Tf2OnnxBackendTestBase):
         self.run_test_case(func, feed_dict, input_names_with_port, output_names_with_port, rtol=1e-5)
 
     @check_tf_min_version("1.9")
-    @check_tf_max_version("1.15")
     def test_simple_while_loop_var_shape(self):
         # test for while_loop with variant shape variables
-        # may not meet ONNX Loop spec
-        # Note: this is not working on tf2 itself.
         def func(i):
             const = tf.constant(np.array([2], dtype=np.int32))
             c = lambda i: tf.reduce_all(tf.shape(i) < 10)
-            b = lambda i: tf.concat([i, const], 0)
+            b = lambda i: [tf.concat([i, const], 0)]
             r = tf.while_loop(c, b, [i], shape_invariants=[tf.TensorShape([None])])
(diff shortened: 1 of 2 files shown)


fatcat-z:
Thanks, this is a great catch!

Could you please add the case you mentioned in related issue as a test into test_cond.py file as well?

f-salvetti:
@fatcat-z I enabled for TF2 a test already present (after fixing it for TF2), however it is failing for tflite due to an infinite loop. I am creating an issue to track this: https://github.com/onnx/tensorflow-onnx/issues/2210

fatcat-z (approved):
LGTM, thanks!

Issue #2154, opened 2023-04-02
TensorFlow 2.12.0 breaks nightly tests.


fatcat-z:
self = <test_backend.BackendTests testMethod=test_Conv2DBackpropInput>
tflite_path = '/tmp/tmp9j283eve/test_Conv2DBackpropInput/test_Conv2DBackpropInput.tflite'

    def tflite_has_supported_types(self, tflite_path):
        try:
            with open(tflite_path, 'rb') as f:
                buf = f.read()
            buf = bytearray(buf)
            model = Model.GetRootAsModel(buf, 0)
            tensor_cnt = model.Subgraphs(0).TensorsLength()
            interpreter = tf.lite.Interpreter(tflite_path)
            for i in range(tensor_cnt):
>               dtype = interpreter._get_tensor_details(i)['dtype']   # pylint: disable=protected-access
E               TypeError: Interpreter._get_tensor_details() missing 1 required positional argument: 'subgraph_index'

pranavladkat:
Is there an ETA for this issue?

q-ycong-p:
+1 on this issue. Any update on a fix?

q-ycong-p:
I put up a naive workaround #2204, just to start discussion towards a fix. Welcome more thoughtful fix of folks have deeper knowledge on tflite's change on subgraph at tf-2.12, and tf2onnx's corresponding path forward.

q-ycong-p:
Hi @fatcat-z, all tests passed after re-opening #2204 - which aims to address this issue. Would appreciate a feedback/review from maintainers. Thank you!

yan12125:
This issue can probably be closed as #2204 is merged?

PR #2204, opened 2023-07-12
Use TFLite Interpreter's public instead of private API to get tensor details

q-ycong-p:
TF-2.12.0 introduced API change that breaks tf2onnx UT tests on the tflite paths, due to the addition of compulsory subgraph arg to several function's input signature:
https://github.com/tensorflow/tensorflow/commit/55d84d7bba942fb4adb552881897c6bbd6e91eef

This commit changes tf2onnx functions that call to TFLite Interpreter class's private `_get_tensor_details`, to instead call its public API `get_tensor_details`. I think the code change is functionality-wise equivalent, from tf2onnx's usage perspective, and circumvent backward comp issue from using a private function.

#2154

### DIFF
--- a/tests/backend_test_base.py
+++ b/tests/backend_test_base.py
@@ -28,7 +28,6 @@ from tf2onnx.tf_loader import tf_reset_default_graph, tf_session, tf_placeholder
 from tf2onnx.tf_loader import tf_optimize, is_tf2, get_hash_table_info
 from tf2onnx.tf_utils import compress_graph_def
 from tf2onnx.graph import ExternalTensorStorage
-from tf2onnx.tflite.Model import Model
 
 
 if is_tf2():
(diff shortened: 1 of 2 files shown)


q-ycong-p:
Two naive thoughts I can think of:
- Would it more appropriate to fix upstream at tf so that their API change is backward-comp? For [example](https://github.com/tensorflow/tensorflow/commit/55d84d7bba942fb4adb552881897c6bbd6e91eef#r121433918).
- Looks like tflite Interpreter class's `_get_tensor_details` is a private func. Would it be better to add a public API that guarantees backward-comp in tflite?

q-ycong-p:
Some tests failed during environment setup independent to code diff. How can we rerun the tests without force push?

yan12125:
> How can we rerun the tests without force push?

Close and reopen this PR should work.

fatcat-z (approved):
This is great! Thanks for your contribution!

PR #2205, opened 2023-07-14
Remove the ml opset version that is saved by default without --extra_opset and --custom-ops

inzapp:
Remove ml version opset that is stored by default without using two flags(--extra_opset, --custom-ops)

This additional ml version opset can cause problems with converting for model serving for various edge devices

If additional ml version offset is required, add it via --extra_opset or --custom-ops


fatcat-z:
Directly removing it will cause other problem which has been covered by tests. Suggest to add a judgement around it so if extra_opset or custom-ops was specified, then skip this default value.

Issue #2211, opened 2023-07-25
Bidirectional LSTM, activation functions are not converted


quentin-baduel-numalis:
**Describe the bug**
Hello, thank you for developing this tool! 
When Converting a Keras model containing a Bidirectional LSTM layer, the activation function attribute (activations) is not set in the corresponding  lstm onnx operator. 
The bug does not seem to appear if we are not in the "Bidirectional" context.

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 20.04*):
- TensorFlow Version: 2.11
- tf2onnx: 1.14.0
- Python version: 3.10

**To Reproduce**

1: Creating the Keras network with the activation functions set to "relu": 
```
#! /usr/bin/env python3

from tensorflow.keras.models import Model
from tensorflow.keras.layers import Input, LSTM, Bidirectional
input = Input(shape=(1,1))
model = Bidirectional(LSTM(units=1, return_sequences=True, activation="relu", recurrent_activation="relu"))(input)
model = Model(input, model)
model.save(f"lstm_issue_activation_function.h5")
```

2: converting the model 
```

#! /usr/bin/env python3

import tensorflow as tf
import onnx
import sys

import tf2onnx

target_model = sys.argv[1]
model = tf.keras.models.load_model(target_model)
spec = (tf.TensorSpec((None,1, 1), tf.float32, name="input"),)
output_path = model.name + ".onnx"
model_proto, _ = tf2onnx.convert.from_keras(model, input_signature=spec, opset=15, output_path=output_path)
onnx.save(model_proto, output_path)
```
3: checking the converted model 
While inspecting the resulting onnx model, we can see that the attribute "activations" of the lstm node is not set and will default to "sigmoid/tanh" rather than the expected "relu" functions

**Screenshots**
![image](https://github.com/onnx/tensorflow-onnx/assets/80971384/276e4173-dd99-4fec-b4a5-d72670fc534e)

MaximeChurin:
Thanks for the lead, I had the same issue. I have a PR ongoing that should fix it: https://github.com/onnx/tensorflow-onnx/pull/2234

PR #2234, opened 2023-09-11
Keep activations in bidirectional LSTM (fixes: https://github.com/onnx/tensorflow-onnx/issues/2211)

MaximeChurin:
The forward and backward LSTMs have their correct activations but as soon as we rewrite it to a bidirectional this attribute is skipped and we get the default activations.

I am not sure how to write it in the `rewriter` file, you have something similar in the [bigru](https://github.com/onnx/tensorflow-onnx/blob/main/tf2onnx/rewriter/bigru_rewriter.py#L66) so let me know if you prefer the syntax with decode and comprehension list

### DIFF
--- a/tests/test_lstm.py
+++ b/tests/test_lstm.py
@@ -776,6 +776,31 @@ class LSTMTests(Tf2OnnxBackendTestBase):
         self.run_test_case(func, {"input:0": x_val}, [], ["output:0", "output1:0"], rtol=1e-05, atol=1e-06)
 
     @check_tf_min_version("2.0")
+    @skip_tf_versions("2.1", "Bug in TF 2.1")
+    def test_keras_bilstm_recurrent_activation_is_hard_sigmoid(self):
+        in_shape = [10, 3]
+        x_val = np.random.uniform(size=[2, 10, 3]).astype(np.float32)
+
+        model_in = tf.keras.layers.Input(tuple(in_shape), batch_size=2)
+        x = tf.keras.layers.Bidirectional(
+            tf.keras.layers.LSTM(
+                units=5,
+                return_sequences=True,
+                return_state=True,
+                kernel_initializer=tf.random_uniform_initializer(0.0, 1.0, seed=42),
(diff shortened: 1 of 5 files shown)


MaximeChurin:
@fatcat-z

fatcat-z:

Yes, please follow the way in bigru_rewriter.py to read activations from those LSTMs.

MaximeChurin:
thanks@fatcat-z
It should be ok now, let me know

fatcat-z:
> thanks@fatcat-z It should be ok now, let me know

The CI failures are caused by tf 1.15. For compatibility, we might need to check if they are None before we call decode() method.

MaximeChurin:
good catch @fatcat-z , I made the change and the tests passed but now it differs a bit from the bigru implem. If you have any suggestions please make it via GitHub code review suggestion

tests/test_lstm.py
```diff
@@ -775,6 +775,31 @@ def func(x):
             return tf.identity(y[0], name="output"), tf.identity(y[1], name="output1")
         self.run_test_case(func, {"input:0": x_val}, [], ["output:0", "output1:0"], rtol=1e-05, atol=1e-06)
 
+    @check_tf_min_version("2.0")
+    @skip_tf_versions("2.1", "Bug in TF 2.1")
```
fatcat-z, line 779: Which bug are we talking about here? Could you please put the url into the comment?

MaximeChurin: no clue I copied the structure of the bigru test that had this decorator, do you think I can remove it?

fatcat-z: > no clue I copied the structure of the bigru test that had this decorator, do you think I can remove it?

Yes please, let's have a try.

fatcat-z: BTW, TF 2.1 has been removed out of the CI pipeline, so this won't be run anyway.

MaximeChurin: done, waiting for the CI to complete

fatcat-z (approved):

Issue #2248, opened 2023-09-27
TensorRT only support "round_prefer_floor" nearest mode but tf2onnx uses "floor"


ShuaiShao93:
**Describe the bug**
When I convert a model that has Resize op, I got error `Assertion failed: (transformationMode != "tf_half_pixel_for_nn" || nearest_mode == "round_prefer_floor") && "This version of TensorRT only support round_prefer_floor nearest mode in tf_half_pixel_for_nn!"`

It turns out TRT8.6.1 only supports "round_prefer_floor" but tf2onnx uses "floor" [here](https://github.com/onnx/tensorflow-onnx/blob/b57fe915b3728fbf996b490170a374821dd8383a/tf2onnx/onnx_opset/nn.py#L1437C5-L1437C5)

ShuaiShao93:
Sent PR https://github.com/onnx/tensorflow-onnx/pull/2249

nchaly:
@ShuaiShao93 why PR https://github.com/onnx/tensorflow-onnx/pull/2249 was not merged?  
https://github.com/onnx/onnx-tensorrt/blob/0462dc31ae78f48744b6141ae376df1f96d3f459/builtin_op_importers.cpp#L3770-L3772  - that check is still there and I'm having a case where importing fails with message "[TRT] [E] ModelImporter.cpp:967: ERROR: onnxOpCheckers.cpp:613 In function checkResize:
[8] (transformationMode != "tf_half_pixel_for_nn" || nearest_mode == "round_prefer_floor") && "This version of TensorRT only support round_prefer_floor nearest mode in tf_half_pixel_for_nn!"

swdn133:
I just stumbled across the exact same issue.
I guess it would be possible to workaround by modifying the resulting onnx with graphsurgeon or anything like that, but this is kind of messy. 
It would be great to have this issue resolved.

nchaly:
@swdn133 Right, for me workaround is rather easy on model side - just try some interpolation which is not 'nearest' for `UpSampling2D` layer (bilinear in my case).

swdn133:
@nchaly 
I need to keep the 'nearest' mode for some reasons. Therefore simply switching the interpolation mode doesn't work for me. I need to keep consistency between multiple different models. The workaround for me right now is to walk over the resulting onnx graph and afterwards changing the attribute from 'floor' to 'round_prefer_floor'.
It's certainly not the cleanest solution, but at least it works.

nchaly:
[...]

PR #2249, opened 2023-09-27
Set correct Resize mode for TRT

ShuaiShao93:
https://github.com/onnx/onnx-tensorrt/blob/0462dc31ae78f48744b6141ae376df1f96d3f459/builtin_op_importers.cpp#L3770-L3772

Issue: https://github.com/onnx/tensorflow-onnx/issues/2248


bertsky:
This still seems necessary for tf-keras models with `UpSampling2D(... interpolation="nearest")`, which still produces the exact error message reported in #2248 under TensorRT execution provider for onnxruntime.

Could you re-open?

Issue #2183, opened 2023-05-31
Support for `LayerNormalization` from ONNX opset 17


pwuertz:
When converting a `tensorflow.keras.layers.LayerNormalization` layer to ONNX, `tf2onnx` currently decomposes layer normalizations into rather complex subgraphs with batch norms and more basic building blocks. Inference engines (like TensorRT in the following example) are hardly able to deduce the original layer norm op from the graph and will have to follow the instructions to the letter:
![image](https://user-images.githubusercontent.com/1819283/241226042-408e0524-1100-48f2-a898-ce615ec9bdf7.png)

Since ONNX version 17 however, `tensorflow.keras.layers.LayerNormalization` layers are directly convertible to ONNX [LayerNormalization](https://github.com/onnx/onnx/blob/main/docs/Operators.md#layernormalization) operators. This of course produces a much simpler and expressive ONNX graph and leaves more room for the inference engine for possible optimizations (ONNX-to-TensorRT example again):
![image](https://github.com/onnx/tensorflow-onnx/assets/1819283/a6026d65-ce42-4106-aa1e-3c20b62c692e)

The current version of `tf2onnx` doesn't seem to produce those new `LayerNormalization` operators yet, at least not with `tf2onnx.convert.from_keras(model, opset=17)`.

DayDayupupupup:
Same issue. Is there any plan to support rewrite LayerNormalization（opset 17）when use tf‘s savedmodel format？

charvey2718:
Same issue for me.

ShuaiShao93:
I have a tentative fix in https://github.com/onnx/tensorflow-onnx/pull/2250. Feel free to take it over

YouSenRong:
Same issue for me.
I am converting the frozen_graph (TF 1.15) to ONNX for employing TensorRT to do inference in FP16 precision.
However, I get the warning as:
"[W] [TRT] Running layernorm after self-attention in FP16 may cause overflow. Exporting the model to the latest available ONNX opset (later than opset 17) to use the INormalizationLayer, or forcing layernorm layers to run in FP32 precision can help with preserving accuracy.".
Converting the frozen_graph to ONNX with Opset 17 does not help either. Finally I found that it due form tf2onnx.
If there is any progress, please let me know.

PR #2250, opened 2023-09-27
Support LayerNormalization for opset 17

ShuaiShao93:
Issue: https://github.com/onnx/tensorflow-onnx/issues/2183


tf2onnx/onnx_opset/nn.py
```diff
@@ -959,6 +959,92 @@ def version_11(cls, ctx, node, **kwargs):
...
+
+        inp_shape = ctx.get_shape(node.input[0])
+        inp_rank = len(inp_shape) if inp_shape is not None else None
+        scale_shape = ctx.get_shape(node.input[1])
+        mean_shape = ctx.get_shape(node.input[3])
+        var_shape = ctx.get_shape(node.input[4])
+        val_type = utils.map_onnx_to_numpy_type(ctx.get_dtype(node.input[1]))
+        is_training = node.get_attr_value('is_training', True)
+
+        if is_training and node.get_attr_value('exponential_avg_factor', 1.0) == 1.0:
```
fatcat-z, line 1013: Thanks for your contributions!

Could you please create a common method where this loop could be refactored like [here](https://github.com/onnx/tensorflow-onnx/blob/b57fe915b3728fbf996b490170a374821dd8383a/tf2onnx/onnx_opset/nn.py#L835C9-L835C9)? So those different versions could call this common method and avoid duplicated code as much as possible.

ShuaiShao93: Ah I didn't know that. Done.

fatcat-z: Besides the pylint issue, looks like this change has broken some legacy tests, please also fix them. Thanks.

ShuaiShao93: Thanks! I mainly use ONNX for TRT, and don't know much about ONNX runtime. But it seems the ONNX layernorm doesn't support automatic broadcasting on scale tensor. So I need to do it manually. But I realized there is no broadcast op in ONNX either. Can you suggest how to broadcast an ONNX tensor?

fatcat-z: > Thanks! I mainly use ONNX for TRT, and don't know much about ONNX runtime. But it seems the ONNX layernorm doesn't support automatic broadcasting on scale tensor. So I need to do it manually. But I realized there is no broadcast op in ONNX either. Can you suggest how to broadcast an ONNX tensor?

Could you please follow the approach [here](https://github.com/onnx/tensorflow-onnx/blob/ca17b3c99eb37dc3f42c9be5a64620d247a24115/tf2onnx/onnx_opset/tensor.py#L4017C6-L4017C6) to try Expand op?

tf2onnx/onnx_opset/nn.py
```diff
@@ -1011,25 +1015,46 @@ def version_6(cls, ctx, node, **kwargs):
...
-            dims_const = ctx.make_const(utils.make_name("axes_const"), np.array(dims, dtype=np.int64)).output[0]
-            reduce_dims = ctx.make_node("Gather", [inp_shape, dims_const]).output[0]
-            dims_product = GraphBuilder(ctx).make_reduce_prod({"data": reduce_dims, "axes": [0], "keepdims": False})
-            cnt_float = ctx.make_node("Cast", [dims_product], attr={'to': ctx.get_dtype(node.input[0])})
-
-            pop_var_squeezed = ctx.make_node("Div", [var_squeezed, cnt_float.output[0]]).output[0]
-            ctx.replace_inputs(node, node.input[:3] + [avg_squeezed, pop_var_squeezed])
+            if ctx.opset >= 17:
+                node.type = "LayerNormalization"
+                first_axis_to_norm = 2
```
aboubezari, line 1020: Hey @ShuaiShao93, why did you add this change? Don't we have to extract the axis value from the axis attribute already in the TF op?

ShuaiShao93: I don't quite understand this very well either. I just saw line 1035 seems to assume axis=2 is first_axis_to_norm. And I remember I could verify the parity end to end.

ShuaiShao93: BTW a better alternative is probably making https://github.com/onnx/tensorflow-onnx/blob/main/tf2onnx/rewriter/layer_normalization_rewriter.py work, so that we can also get rid of a bunch of other ops added by TF

PR #2269, opened 2023-11-08
Support bool inputs for MatrixBandPart

### LINKED ISSUE #2268: MatrixBandPart op in tf2onnx does not support bool inputs
**Describe the bug**

For the op `MatrixBandPart`, inputs of type `bool` are not supported by `tf2onnx`. The corresponding op in TensorFlow does support bool types. The following error appears when converting `tf.linalg.band_part` with boolean input
```
onnxruntime.capi.onnxruntime_pybind11_state.InvalidGraph: [ONNXRuntimeError] : 10 : INVALID_GRAPH : Load model from model.onnx failed:This is an invalid model.
Type Error: Type 'tensor(bool)' of input parameter (Cast__43:0) of operator (Mul) in node (PartitionedCall/MatrixBandPart) is invalid.
```

**Urgency**

Appears in a couple of Transformer language models. We can provide a simple fix in our TF code in the mean time. Not so urgent.

**System information**

- OS Platform and Distribution (e.g., Linux Ubuntu 18.04*): Linux Ubuntu 18.04.6 LTS
- TensorFlow Version: 2.14.0, 2.13.1, 2.9.0 tested so far
- Python version: 3.9
- ONNX version (if applicable, e.g. 1.11*): 1.15.0
- ONNXRuntime version (if applicable, e.g. 1.11*): 1.16.1

**To Reproduce**
```python
import tensorflow as tf

class CustomModule(tf.Module):

    @tf.function(input_signature=[tf.TensorSpec([None, None], tf.bool)])
    def __call__(self, x):
        return tf.linalg.band_part(x, -1, 0)

bandpart = CustomModule()

x = tf.ones([3, 3], dtype=tf.bool)
print(bandpart(x))

tf.saved_model.save(bandpart, "saved-model-example")
```
then
```
python3 -m tf2onnx.convert --saved-model saved-model-example --output model.onnx
```
and then running
```python
import numpy as np
import onnxruntime as ort

ort_sess = ort.InferenceSession('model.onnx')
x = np.ones([3, 3], dtype=bool)

print(ort_sess.run(None, {'x': x}))

```

**Screenshots**

None

**Additional context**

We use this op with boolean inputs to create a attention mask.

One workaround would be
```python
def onnx_compat_matrix_band_part(x, num_lower, num_upper):
    if x.dtype == tf.bool:
        inputs = tf.cast(x, dtype=tf.float32)
    else:
        inputs = x                                                                                                                                                                                        
    res = tf.linalg.band_part(inputs, num_lower, num_upper)
    if x.dtype == tf.bool:
        return tf.cast(res, dtype=tf.bool)
    return res
```
See issue https://github.com/rwth-i6/returnn/issues/1455 for some notes.

Gerstenberger:
Fixes #2268 

Please let me know if this is something you want to support.
Should there also be a test case?

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -3926,7 +3926,6 @@ class BackendTests(Tf2OnnxBackendTestBase):
                 return tf.identity(res, name=_TFOUTPUT)
             self._run_test_case(func, [_OUTPUT], {_INPUT: input_val, _INPUT1: low_val, _INPUT2: high_val})
 
-    @check_opset_min_version(11, "CumSum")
     def test_matrix_band_part_bool(self):
         input_val = np.random.choice([False, True], size=(10, 15))
         def func(input_x):
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -3926,6 +3926,15 @@ class BackendTests(Tf2OnnxBackendTestBase):
                 return tf.identity(res, name=_TFOUTPUT)
             self._run_test_case(func, [_OUTPUT], {_INPUT: input_val, _INPUT1: low_val, _INPUT2: high_val})
 
+    @check_opset_min_version(11, "CumSum")
+    def test_matrix_band_part_bool(self):
+        input_val = np.random.choice([False, True], size=(10, 15))
+        def func(input_x):
+            res = tf.linalg.band_part(input_x, -1, 0)
+            res1 = tf.linalg.band_part(input_x, 0, -1)
+            return tf.identity(res, name=_TFOUTPUT), tf.identity(res1, name=_TFOUTPUT1)
+        self._run_test_case(func, [_OUTPUT, _OUTPUT1], {_INPUT: input_val})
+
     def test_floordiv(self):
         input_val_1 = np.random.random_sample(100).astype(np.int32)
         input_val_2 = (np.random.random_sample(100) + 1).astype(np.int32)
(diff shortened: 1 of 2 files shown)


Gerstenberger:
Added a test case

tests/test_backend.py
```diff
@@ -3926,6 +3926,15 @@ def func(input_x, low, high):
                 return tf.identity(res, name=_TFOUTPUT)
             self._run_test_case(func, [_OUTPUT], {_INPUT: input_val, _INPUT1: low_val, _INPUT2: high_val})
 
+    @check_opset_min_version(11, "CumSum")
```
fatcat-z, line 3929: Why do we need to add a limitation for CumSum op for this test?

Gerstenberger: i copied the function from above and then made edits (the other test don't need this too then?). I removed the limitation for the test set now.

fatcat-z: It depends on if ONNX op CumSum is required during the conversion of a tf op. If yes, then we need to limit min opset version to 11 because ONNX supports CumSum op since 11. By adding this, the test case will be ignored if the test suite is running with an ONNX opset lower than 11.

fatcat-z: MatrixBandPart is a tf op. It doesn't have a requirement for ONNX opset version. The ONNX opset version requirement depends on all ONNX ops we use during the conversion of a TF op.

fatcat-z (approved):

PR #2270, opened 2023-11-08
Fix unsupported ops TF 2.14.0: OnesLike

### LINKED ISSUE #2267: Unsupported ops TF 2.14.0: OnesLike not supported
# New Operator

### Describe the operator
With `tensorflow==2.14.0` the conversion of op `OnesLike` yields the error
```
Tensorflow op [output/rec/dec_0_self_att_att/ones_like: OnesLike] is not supported
```
which works fine using `tensorflow==2.13.1`.

### Do you know this operator be constructed using existing ONNX operators?
yes, similar to `ZerosLike` using `ConstantOfShape`

### Is this operator used by any model currently? Which one?
yes, in our SelfAttention (used eg. in our Transformer models) implementation for masking.

### Are you willing to contribute it? (Y/N)
yes

### Notes
For some notes see issue https://github.com/rwth-i6/returnn/issues/1455
The onnx package versions used
```
tf2onnx==1.15.1
onnx==1.15.0
onnxruntime==1.16.1
```

Gerstenberger:
Fixes #2267
Do you want a test case for this?

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -4105,10 +4105,7 @@ class BackendTests(Tf2OnnxBackendTestBase):
         self._run_test_case(func, [_OUTPUT], {_INPUT: input_x.astype(np.int32), _INPUT1: input_y}, as_session=True,
                             graph_validator=lambda g: check_op_count(g, "ConstantOfShape", 1, disabled=False))
 
-    @check_tf_max_version("2.13.1")
-    @check_opset_min_version(9, "ConstantOfShape")
-    @check_opset_max_version(9, "ConstantOfShape")
-    def test_ones_like_opset9(self):
+    def test_ones_like(self):
         input_x = np.random.random_sample([3, 16, 16]).astype(np.float32)
         input_y = np.array([16, 16, 3]).astype(np.int64)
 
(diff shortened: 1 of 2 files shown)


fatcat-z:

Yes, please add a test case as well.

Gerstenberger:
Added some test cases

tests/test_backend.py
```diff
@@ -4105,6 +4105,54 @@ def func(x, y):
         self._run_test_case(func, [_OUTPUT], {_INPUT: input_x.astype(np.int32), _INPUT1: input_y}, as_session=True,
                             graph_validator=lambda g: check_op_count(g, "ConstantOfShape", 1, disabled=False))
 
+    @check_tf_max_version("2.13.1")
```
fatcat-z, line 4108: This means this test case won't be run on any versions above 2.13.1. Could you please share why we can't do this? Or we can remove this limitation?

tests/test_backend.py
```diff
@@ -4105,6 +4105,54 @@ def func(x, y):
...
+    @check_opset_min_version(9, "ConstantOfShape")
+    @check_opset_max_version(9, "ConstantOfShape")
```
fatcat-z, line 4110: Since we already have 2 functions for different versions, we don't need to add an opset limitation for these test methods. And we don't need to create different test cases for different versions. 

You just need to create one test case for tf op OnesLike, and the test framework will help on testing different ops.

tests/test_backend.py
```diff
@@ -4105,6 +4105,54 @@ def func(x, y):
...
+        def func(x, y):
+            z = tf.reshape(x, y)
+            return tf.ones_like(z, name=_TFOUTPUT)
+
+        self._run_test_case(func, [_OUTPUT], {_INPUT: input_x, _INPUT1: input_y}, as_session=True,
+                            graph_validator=lambda g: check_op_count(g, "Expand", 1, disabled=False))
+        self._run_test_case(func, [_OUTPUT], {_INPUT: input_x.astype(np.int32), _INPUT1: input_y}, as_session=True,
+                            graph_validator=lambda g: check_op_count(g, "Expand", 1, disabled=False))
+
+    @check_tf_min_version("2.14.0")
```
fatcat-z, line 4139: Why do we need to require 2.14.0 version to run this test case?

Gerstenberger:
Hi, thanks for your comments.
I only kept one test case now without checking for specific ops (i think that more natural in this case).

fatcat-z (approved):

PR #2311, opened 2024-03-07
Add support for Erf and Selu in transpose_optimizer.py.

Zyrin:
I added support for Erf (for Gelu) and Selu to transpose_optimizer.py as our models use them and the resulting onnx models had far too many transpose operations.


fatcat-z (approved):
Great! Thanks for your contributions!

Issue #2328, opened 2024-04-24
Project requires an old version of Protobuf 


wingdagger:
The following setup.py file requires version ~=3.20 of protobuf.  This is a very old release.  Protobuf is currently on version 5.26.1.  This constraint of ~=3.20 is causing conflicts when I generate my python dependencies.  Can you please allow for newer versions of protobuf?  Thank you.

https://github.com/onnx/tensorflow-onnx/blob/ed022438576ca7a351c3858105439f4f25c35219/setup.py#L85C95-L85C111

fatcat-z:
Thanks for your comments..

At this moment, tf2onnx is still designed to support older TensorFlow versions as many as possible and some of them could not work on a version above 3.20.x. In addition, couple of dependencies of tf2onnx also required the protobuf version ~= 3.20.0, so we can't upgrade its version in setup.py now.

yan12125:
> Shouldn't tensorflow indicate the prereq version of protobuf that it requires. 

Apparently only [tensorflow > 2.10](https://github.com/tensorflow/tensorflow/commit/60d5bfbf0241e00267884b0dc1723bbdfee1806a) does that. That's probably why many tests fail with older tensorflow versions in #2327.

yan12125:
Here is an idea: Python 3.11 is supported [since tensorflow 2.12](https://github.com/tensorflow/tensorflow/releases/tag/v2.12.0). I think it's safe to say on Python >= 3.11, tensorflow already properly specifies compatible protobuf versions. How about something like this?
```python
install_requires=[
    # (other dependencies)
    'protobuf~=3.20; python_version < "3.11"',
    'protobuf; python_version >= "3.11"',
]
```

> In addition, couple of dependencies of tf2onnx also required the protobuf version ~= 3.20.0

I didn't check other dependencies. Such a trick may apply as well.

mjm522:
Could this be done on the `setup.py`. Currently the version `1.61.1` is not compatible with bigquery or any gcp package that requires protobuf > 4 Though we could fork and it and do it, just wanted know the possibility of updating it on the main? I understand from the comments that it does work with protobuf > 4

OvervCW:
Any news about this? We currently have to maintain a fork with a single line change because of this requirement.

dmal-msft:
> Any news about this? We currently have to maintain a fork with a single line change because of this requirement.
> 
> And when will this package officially support Python 3.11 and later?
Any updated on this?
[...]

PR #2336, opened 2024-05-30
Update `protobuf` Dependency to Support Versions 4 and 5

mldangelo:
Packages like MediaPipe (0.10.14) rely on a `protobuf` version greater than 4, meaning `tensorflow-onnx` and `mediapipe` cannot be installed together in package managers like poetry due to the version restriction. This PR updates the `protobuf` dependency to `<6.0.0`, we aim to ensure compatibility with these libraries and improve overall dependency resolution. This is my first contribution to the TensorFlow-ONNX project. If this direction is not appropriate or there are other considerations, please let me know, and I will address them accordingly. Thank you for maintaining such an excellent project!


fatcat-z:
There are some discussions about this topic in #2328 , please refer to there for more details.

BTW, what's the lowest TF version for your project?

PR #2344, opened 2024-06-25
Refactor CI jobs for adding required jobs in Actions.

fatcat-z:
Refactor CI workflow yaml files, so that we can add required CI jobs properly in Ruleset.


xiaowuhu (approved):

.github/actions/keras_application_test/action.yml
```diff
@@ -0,0 +1,75 @@
+name: Keras2onnx Application Test Run
```
xadupre, line 1: Do you need an extra folder?

fatcat-z: Such path is required by Github Workflow, so we don't need to customize them more.

PR #2361, opened 2024-10-19
Update end2end_tfkeras.py

MasoudKaviani:
Tensorflow Keras save model function change to export for `pb` export extension


fatcat-z (approved):

PR #2384, opened 2025-02-04
Troves for python 3.11 and 3.12

chayim:
Thanks for the great project!

When attempting to pull this in, under python 3.12, I noticed that the satsolver failed to resolve (I'm using [uv](https://docs.astral.sh/uv/)!) as there's no trove for python 3.12.

I'm positive there are changes other than this that are necessary, but I wanted to at least help this along, as python 3.12 is the default install on several Linux distributions.

How else can I help?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,5 +98,8 @@ setup(
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
+        'Programming Language :: Python :: 3.10',
+        'Programming Language :: Python :: 3.11'
+        'Programming Language :: Python :: 3.12']
         'Programming Language :: Python :: 3.10']
 )


xadupre (approved):

chayim:
@xadupre have a glace?

xadupre:
I started to look into this in #2385 but I realized it would take much more time than I anticipated since 2.18 seems to be very different from the previous versions. Some apis tensorflow-onnx rely on don't work anymore. But maybe we can do that step by step. Your PR seems closer to completion than mine.

andife:
I’ve published a release candidate:
https://pypi.org/project/tf2onnx/1.17.0rc1/

This release includes significant changes, so thorough testing would be greatly appreciated. Please try it out and report any issues or regressions you encounter.

Thanks in advance for your help!

Andreas

PR #2392, opened 2025-05-06
Replace np.cast (deprecated in numpy > 2) with np.asarray.

Satge96:
This should also work for older numpy versions.


Satge96:
Thanks for approving the change. What can I do to make the enforce-style successful?

Stonesjtu:
@xadupre can you take a look at this PR?

WangHHY19931001:
may be you should check numpy version and use diff way?

PR #2398, opened 2025-06-12
Support Keras 3 models in from_keras

nfirle:
### Summary

This PR adds support for Keras 3 in the `from_keras` conversion function. Keras 3 introduced several breaking changes in model serialization and internal APIs, which caused the existing conversion logic to fail. This update ensures compatibility with the new format.

### Changes

- Added a new function `from_keras3` that handles models created with Keras 3.
- Updated `from_keras` to detect the Keras version and delegate to `from_keras3` when necessary.
- The `from_keras3` function is designed to also work with older Keras versions, but for full backward compatibility, the original `from_keras` logic is preserved.
- This approach minimizes the risk of regressions and keeps the behavior consistent for users still relying on Keras 2.

### DIFF
--- a/tf2onnx/convert.py
+++ b/tf2onnx/convert.py
@@ -464,7 +464,7 @@ def from_keras3(model, input_signature=None, opset=None, custom_ops=None, custom
     
 
     valid_names = []
-    for out in model.output_names:
+    for out in [t.name for t in model.outputs]:
         if out in reverse_lookup:
             valid_names.append(reverse_lookup[out])
         else:
@@ -481,9 +481,6 @@ def from_keras3(model, input_signature=None, opset=None, custom_ops=None, custom
     with tf.device("/cpu:0"):
         frozen_graph, initialized_tables = \
             tf_loader.from_trackable(model, concrete_func, input_names, output_names, large_model)
-        
-        for node in frozen_graph.node:
-            print(node.name, node.op)
         model_proto, external_tensor_storage = _convert_common(
             frozen_graph,
             name=model.name,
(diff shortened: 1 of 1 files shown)


andife:
@xadupre could you have a look at it?

tf2onnx/convert.py
```diff
@@ -408,6 +408,106 @@ def _from_keras_tf1(model, opset=None, custom_ops=None, custom_op_handlers=None,
...
+            input_names=input_names,
+            output_names=output_names,
+            inputs_as_nchw=inputs_as_nchw,
+            outputs_as_nchw=outputs_as_nchw,
+            large_model=large_model,
+            tensors_to_rename=tensors_to_rename,
+            initialized_tables=initialized_tables,
+            output_path=output_path)
+        
+        #print(model_proto)
```
sharktide, line 508:
```suggestion
```

Might be better to remove the debug statement and extra newlines

nfirle:
Hello, I just wanted to ask whether the pull request is still under review or if there is no further interest in it. If it is still being considered, is there anything I should change to make it suitable for integration?

sharktide:
> Hello, I just wanted to ask whether the pull request is still under review or if there is no further interest in it. If it is still being considered, is there anything I should change to make it suitable for integration?

I am interested, and I think this would be a useful feature, however I feel that I am not the right reviewer for this PR

andife:
With the major repository changes that landed recently, this PR is now showing a merge conflicts. If you're up for it, feel free to retry and adapt the changes for the newer version — that would be much appreciated.

andife:
does the supported tensorflow version of tf2onnx has to be updated before?

PR #2401, opened 2025-07-05
fix md table (Update README.md)

andife:


sharktide:
This PR (#2402) contains both this fix and updated the python versions in the readme.
Will be superseded by that

PR #2414, opened 2025-10-01
Feat : support NHWC to NCHW in split node wo transpose

dongryeolle:
# Summary
- current tf2onnx doesn't support to convert without transpose NHWC to NCHW in split node. 
- and the most of functions are not consider when operator has multiple outputs.

# Changes
- add features that change axis of Split in transpose optimizer for NHWC to NCHW.
- add `_switch_transpose_and_node_with_multiple_outputs` functions for switching transpose node and split node has multiple-output.

# Test
- Unfortunately due to security issues, I attach before & after figure instead of my test model. 
image

- in unit_test,`python setup.py test`,  i got results as below. i think those failed issues are not related in my code.

image

### DIFF
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -312,7 +312,27 @@ class TransposeOptimizer(GraphOptimizerBase):
             self._g.set_shape(node.output[0], new_shape)
             self._g.set_shape(trans.output[0], shape)
         return True
-
+    # this is for the case where node has multiple outputs. e.g. split node.
+    def _switch_transpose_and_node_with_multiple_outputs(self, node, trans, update_shape=True):
+        input_index = self._get_input_index_for_trans(node, trans)
+        for idx,_output in enumerate(node.output):
+            shape = self._g.get_shape(_output)
+            nxt_nodes = self._g.find_output_consumers(_output)
+            if idx == 0:
+                transpose = trans
+                self._g.replace_input(node, node.input[input_index], transpose.input[0], input_index)
+                self._g.replace_input(trans, trans.input[0], _output, 0)
+            else:
+                transpose = self._g.make_node("Transpose", [_output], attr={"perm": trans.get_attr_value("perm")})
+            for nxt_node in nxt_nodes:
+                self._g.replace_input(nxt_node, _output, transpose.output[0])
+                
+            if update_shape and shape:
+                perm_inv = invert_perm(transpose.get_attr_value("perm"))
+                new_shape = [shape[i] for i in perm_inv]
+                self._g.set_shape(_output, new_shape)
+                self._g.set_shape(transpose.output[0], shape)
+        return True
     # if return value is True, then it means Transpose is handled as designed
     # otherwise, it means that we skip handling since it is not in our support set
     def _handle_nhwc_tranpose(self, trans):
@@ -694,6 +714,21 @@ class TransposeOptimizer(GraphOptimizerBase):
                 new_axes_np = np.array(split, dtype=np.int64)
                 new_axes_const = self._g.make_const(utils.make_name(node.inputs[1].name), new_axes_np)
                 self._g.replace_inputs(node, [node.input[0], new_axes_const.output[0]])
+            return True
+        # handling having branches
+        if len(node.output) > 1:
+            trans_rank = get_transpose_rank(trans)
+            axes = node.get_attr_value("axis", 0)
+            perm = trans.get_attr("perm").ints
+            axes = [axes + trans_rank if axes < 0 else axes]
+            if split:
+                new_axes_np = np.array(split, dtype=np.int64)
+                new_axes_const = self._g.make_const(utils.make_name(node.inputs[1].name), new_axes_np)
+            # [Transpose -> Split -> next_nodes] -> [Split -> Transpose -> next_nodes]
+            if not self._switch_transpose_and_node_with_multiple_outputs(node, trans, 1):
+                return False
+            new_axes = [perm[a] for a in axes]
+            node.set_attr("axes", new_axes)
             return True
         return False
 
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -312,10 +312,11 @@ class TransposeOptimizer(GraphOptimizerBase):
             self._g.set_shape(node.output[0], new_shape)
             self._g.set_shape(trans.output[0], shape)
         return True
+
     # this is for the case where node has multiple outputs. e.g. split node.
     def _switch_transpose_and_node_with_multiple_outputs(self, node, trans, update_shape=True):
         input_index = self._get_input_index_for_trans(node, trans)
-        for idx,_output in enumerate(node.output):
+        for idx, _output in enumerate(node.output):
             shape = self._g.get_shape(_output)
             nxt_nodes = self._g.find_output_consumers(_output)
             if idx == 0:
@@ -326,7 +327,7 @@ class TransposeOptimizer(GraphOptimizerBase):
                 transpose = self._g.make_node("Transpose", [_output], attr={"perm": trans.get_attr_value("perm")})
             for nxt_node in nxt_nodes:
                 self._g.replace_input(nxt_node, _output, transpose.output[0])
-                
+
             if update_shape and shape:
                 perm_inv = invert_perm(transpose.get_attr_value("perm"))
                 new_shape = [shape[i] for i in perm_inv]
@@ -715,20 +716,20 @@ class TransposeOptimizer(GraphOptimizerBase):
                 new_axes_const = self._g.make_const(utils.make_name(node.inputs[1].name), new_axes_np)
                 self._g.replace_inputs(node, [node.input[0], new_axes_const.output[0]])
             return True
-        # handling having branches
+        # Split with more than 1 output (the single-output case is handled above by
+        # _handle_node_having_branches, which bails out when len(node.output) != 1).
         if len(node.output) > 1:
+            perm = trans.get_attr_value("perm")
             trans_rank = get_transpose_rank(trans)
-            axes = node.get_attr_value("axis", 0)
-            perm = trans.get_attr("perm").ints
-            axes = [axes + trans_rank if axes < 0 else axes]
-            if split:
-                new_axes_np = np.array(split, dtype=np.int64)
-                new_axes_const = self._g.make_const(utils.make_name(node.inputs[1].name), new_axes_np)
+            axis = node.get_attr_value("axis", 0)
+            if axis < 0:
+                axis += trans_rank
             # [Transpose -> Split -> next_nodes] -> [Split -> Transpose -> next_nodes]
-            if not self._switch_transpose_and_node_with_multiple_outputs(node, trans, 1):
+            # The split sizes stay the same; only the axis is relabelled to the
+            # pre-transpose layout. ONNX Split's "axis" is a scalar int attribute.
+            if not self._switch_transpose_and_node_with_multiple_outputs(node, trans):
                 return False
-            new_axes = [perm[a] for a in axes]
-            node.set_attr("axes", new_axes)
+            node.set_attr("axis", perm[axis])
             return True
         return False
 
--- a/tf2onnx/optimizer/transpose_optimizer.py
+++ b/tf2onnx/optimizer/transpose_optimizer.py
@@ -315,6 +315,16 @@ class TransposeOptimizer(GraphOptimizerBase):
 
     # this is for the case where node has multiple outputs. e.g. split node.
     def _switch_transpose_and_node_with_multiple_outputs(self, node, trans, update_shape=True):
+        # Defensive guard, mirroring _handle_nhwc_tranpose's `trans.output[0] in outputs`
+        # check. Each output needs a compensating Transpose inserted between `node` and its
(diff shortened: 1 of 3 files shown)


andife:
The tests are failing. Do they have to be updated?

dongryeolle:
Hi @andife I'll look into failed issue.

dongryeolle:
Thanks for the review, @andife.

  I looked into the failing tests. They are all **collection errors caused by Keras 3**,
  not by this change:

    test_cudnn_compatible_gru / test_custom_rnncell / test_gru / test_grublock /
    test_lstm / test_lstmblock / test_seq2seq / test_stacked_lstm
    → AttributeError: `MultiRNNCell` / `BasicLSTMCell` / `LSTMCell` is not available with Keras 3

  These RNN cells were removed in Keras 3, so the test modules fail at import time
  regardless of this PR (they reproduce on `main` as well). They are unrelated to the
  Split / transpose-optimizer code touched here.

  I also pushed a fix for two real issues in the original commits, with a new test:
  - `onnx_opset/tensor.py`: removed the unconditional `axis 3 -> 1` rewrite in
    `Split.version_1`, which corrupted any 4D `tf.split` on the last axis when no layout
    change was involved. The layout-axis adjustment is handled by the transpose optimizer.
  - `optimizer/transpose_optimizer.py`: the multi-output Split branch set a list `axes`
    attribute, but ONNX Split uses the scalar `axis`; fixed that and removed dead code.
  - added `test_transpose_with_split_multiple_outputs` to cover the multi-output path.

  `tests/test_optimizers.py` and the Split backend tests pass locally (opset 9–18).

andife:
@dongryeolle could you fix dco?

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -695,6 +716,21 @@ def _split_handler(self, trans, node):
                 new_axes_const = self._g.make_const(utils.make_name(node.inputs[1].name), new_axes_np)
                 self._g.replace_inputs(node, [node.input[0], new_axes_const.output[0]])
```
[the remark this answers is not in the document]
dongryeolle: I agree. added in `2adc7bc`

andife:
@dongryeolle How about the copilot comments? Valid? Adressed?

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -695,6 +716,21 @@ def _split_handler(self, trans, node):
...
             return True
+        # Split with more than 1 output (the single-output case is handled above by
+        # _handle_node_having_branches, which bails out when len(node.output) != 1).
+        if len(node.output) > 1:
+            perm = trans.get_attr_value("perm")
+            trans_rank = get_transpose_rank(trans)
```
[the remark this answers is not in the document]
dongryeolle: I changed PR description.

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -695,6 +716,21 @@ def _split_handler(self, trans, node):
...
+            axis = node.get_attr_value("axis", 0)
+            if axis < 0:
+                axis += trans_rank
+            # [Transpose -> Split -> next_nodes] -> [Split -> Transpose -> next_nodes]
+            # The split sizes stay the same; only the axis is relabelled to the
+            # pre-transpose layout. ONNX Split's "axis" is a scalar int attribute.
+            if not self._switch_transpose_and_node_with_multiple_outputs(node, trans):
+                return False
+            node.set_attr("axis", perm[axis])
```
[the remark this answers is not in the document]
dongryeolle: As i know there is already exist merge_duplicated_tranpose method for this.

dongryeolle:
Hi @andife 

Yes! several of the Copilot comments were valid and I've addressed them. also i wrote msg in each comment of Copilot

dongryeolle:
I don't know why [this test](https://github.com/onnx/tensorflow-onnx/actions/runs/28496934219/job/84466272287?pr=2414) failed. could you re-run this test?

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -695,6 +715,27 @@ def _split_handler(self, trans, node):
                 new_axes_const = self._g.make_const(utils.make_name(node.inputs[1].name), new_axes_np)
                 self._g.replace_inputs(node, [node.input[0], new_axes_const.output[0]])
             return True
+        # Split with more than 1 output (the single-output case is handled above by
+        # _handle_node_having_branches, which bails out when len(node.output) != 1).
```
andife: The guard here only checks that `trans` has a single consumer — it does not check whether any of the split outputs are graph outputs or have zero node-consumers.

`_handle_nhwc_tranpose` already guards `trans.output[0]` from being a graph output (line 319 of the original file), but that protection does not extend to `node.output[i]`. If any `node.output[i]` is in `self._g.outputs` (a terminal graph output with no further node-consumers), `find_output_consumers(_output)` returns `[]`, so the consumer-redirect loop in `_switch_transpose_and_node_with_multiple_outputs` never runs. The split axis is relabelled to NCHW but no compensating transpose is inserted on that output edge, silently exposing an NCHW tensor at a graph output that was expected to be NHWC.

The same issue applies to disconnected/unused optional outputs: for `idx == 0` this leaves `trans.output[0]` dangling; for `idx > 0` an orphaned `Transpose` node is created.

Suggested fix: add a guard before entering this block, mirroring the check already present in `_handle_nhwc_tranpose`:
```python
if any(out in self._g.outputs for out in node.output):
    return False
```

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -680,7 +701,6 @@ def _concat_handler(self, trans, node):
         return False
 
     def _split_handler(self, trans, node):
-        # Todo: need handle cases where Split node has more than 1 outputs.
         split = None
```
andife: The `split` variable loaded here (the opset-13 split-sizes tensor) is only consumed inside the `if self._handle_node_having_branches(trans, node):` block above. `_handle_node_having_branches` returns `False` whenever `len(node.output) != 1`, so by the time the new `if len(node.output) > 1:` branch is reached, `split` is always unused — dead code in this path.

No correctness impact, but a future maintainer might assume the multi-output path also re-creates the split-sizes constant (as the single-output path does), which it does not. A comment or moving the variable inside the single-output branch would make the mutual exclusivity explicit.

tf2onnx/optimizer/transpose_optimizer.py
andife: posted by Claude

tf2onnx/optimizer/transpose_optimizer.py
andife: posted by Claude

dongryeolle: This one is subtler than it looks. Moving the split capture inside the                                   
  _handle_node_having_branches block (or after the call) breaks                                            
  test_transpose_with_split_opset13 with split tensor must be a vector.                                    
                                                      
  Reason: _handle_node_having_branches → _create_transpose_pairs_before_node wraps
  every input of the node in Reshape+Transpose pairs, including the opset-13 1-D                           
  split-sizes const (input[1]). So the split sizes must be snapshotted before that                         
  call and a clean const rebuilt after. The read only looks unused; it's a
  capture→mutate→repair sequence with real temporal coupling.                                              
                                                                                                           
  Rather than leave it implicit, I restructured _split_handler to dispatch on                              
  len(node.output):                                                                                        
  - multi-output → the transpose switch; opset-agnostic, never touches split sizes;                        
  - single-output → the _handle_node_having_branches push, with the opset-13                               
  capture/repair localized here and commented (opset < 13: split sizes are an attribute,                   
  untouched → split stays None).                                                                           
                                                                                                           
  I deliberately did not change _create_transpose_pairs_before_node to skip Split's                        
  control input — it's shared by ~8 handlers where every input is genuinely a data tensor,                 
  so a Split-specific carve-out there would be a leaky abstraction with broad blast radius.                
  The localized snapshot/restore is the lower-risk choice.                                                 
                                                                                                           
  Also folded in small nits: if split is not None: (empty list shouldn't read as                           
  "not captured"), seed the rebuilt const name from node.name instead of the now-mangled                   
  node.inputs[1].name, and renamed new_axes_* → split_sizes_* (they're split sizes,                        
  not axes).
  
  written by @dongryeolle and Claude

dongryeolle:
- test_transpose_with_split_outputs_are_graph_outputs — multi-output Split feeding graph                 
  outputs, numerics preserved (4D/5D).                                                                     
  - test_split_multi_output_graph_output_guard — white-box guard test (fails without the guard).           
  - test_transpose_with_split_multiple_outputs_split_input — opset≥13 multi-output Split                   
  with an explicit, uneven split-sizes input; verifies the sizes survive the switch.                       
                                                                                                           
  Full OptimizerTests suite: 236 passed, 0 failed (opset 9–18).

tf2onnx/optimizer/transpose_optimizer.py
dongryeolle: Thanks for your review.

Removed the multi-output Split graph-output guard — it's unreachable dead code.                          
                                                                                                           
  Graph.__init__ inserts an Identity in front of every graph output, so when _split_handler runs a Split's outputs are always raw names, never graph outputs — the guard condition can't be true. It isn't reachable
  in later passes either: by the time IdentityOptimizer strips the buffer, the Split has already been processed in an earlier pass, so no transpose feeds it anymore.
                                                                                                           
  Verified empirically: ran 7 dummy models ending in a Split (incl. attempts to trigger it via mid-graph Identity removal and inverse-transpose branches) through the full pipeline → guard fired 0/7, all results
  correct.                                                                                                 
                                                                                                           
  Keeping it would be dead code that fakes coverage and misleads future readers, and it could only be tested by hand-mutating g.outputs (an artificial, non-existent state). With the guard gone the helper has no
  failure path, so I also dropped the now-dead if not ...: return False at the call site.                  
                                                                                                           
  Replaced the earlier white-box test with an end-to-end Relu → Transpose → Split-at-tail test. Full OptimizerTests: 236 passed.
  
 reviewed dongryeolle, posted by Claude

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -680,22 +702,45 @@ def _concat_handler(self, trans, node):
...
+        # rebuild a clean const AFTER. On opset < 13 the sizes are an attribute and untouched,
+        # so `split` stays None and this repair is skipped.
+        split = None
+        if self._g.opset >= 13 and len(node.input) > 1 and node.inputs[1].is_const():
+            split = node.inputs[1].get_tensor_value(as_list=True)
+        if not self._handle_node_having_branches(trans, node):
+            return False
+        perm = trans.get_attr_value("perm")
+        node.set_attr("axis", perm[node.get_attr_value("axis", 0)])
+        if split is not None:
```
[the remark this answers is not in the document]
dongryeolle: Already correct as-is. `perm` here is the transpose's full permutation, so                              
  `len(perm) == rank`. For a negative `axis`, Python's `perm[axis]` is by                                 
  definition `perm[axis + len(perm)] == perm[axis + rank]`, which is exactly what                         
  the explicit `axis += rank` normalization in the multi-output path computes — so                        
  the single-output path already maps negative axes correctly, and the stored                             
  result is non-negative either way.                                                                      
                                                                                                          
  Verified empirically: a `Transpose → Split(axis=-1) → Transpose` model (both                            
  single- and multi-output) optimizes to the same rewritten Split axis with                               
  identical numerics and zero remaining transposes. The explicit normalization in                         
  the multi-output branch is redundant rather than a fix, so I left the                                   
  single-output path unchanged to avoid a no-op diff.

tf2onnx/optimizer/transpose_optimizer.py
```diff
@@ -313,6 +313,28 @@ def _switch_transpose_and_node(self, node, trans, update_shape=True):
             self._g.set_shape(trans.output[0], shape)
         return True
 
+    # this is for the case where node has multiple outputs. e.g. split node.
+    def _switch_transpose_and_node_with_multiple_outputs(self, node, trans, update_shape=True):
+        input_index = self._get_input_index_for_trans(node, trans)
```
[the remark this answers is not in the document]
dongryeolle: Agreed, addressed. Strictly the state isn't reachable now — `_split_handler`                          
  is only dispatched from `_handle_nhwc_tranpose` when the Split is already the                           
  transpose's sole consumer — but the helper was asymmetric with the single-output                        
  `_switch_transpose_and_node` (which self-guards) and unsafe for any future                              
  caller. I moved the `_nodes_has_single_consumer_node([trans])` guard into                               
  `_switch_transpose_and_node_with_multiple_outputs` itself (returns False before
  any rewiring) and made `_split_handler` honor that return value instead of                              
  pre-checking separately.                           
                                                                                                          
  Added `test_multi_output_switch_guards_shared_transpose`: it feeds a graph where
  one Transpose feeds both a multi-output Split and an independent Relu, calls the
  helper directly, and asserts it bails without touching the Split's axis/inputs or                       
  the sibling consumer. Confirmed the test fails if the guard is removed.

andife:
@dongryeolle ready to merge?

dongryeolle:
Hi @andife 

yes i think there is nothing to do for merge

andife (approved):

andife:
@dongryeolle the failing tests are unreleated to the change here. I would merge this now. should you have a look at the ci-failures?

Issue #1831, opened 2022-01-25
RunTimeWarning: tf2onnx.convert sysmodules 


DeveloperNo579212:
I get this below and tried to uninstall and then reinstall based on info on:
https://github.com/onnx/tensorflow-onnx

I get unexpected results in the conversion and would like to fix this, but why is it appearing when following install procedure above?  Seems that some precondition is not met (not listed?) or the install documentation needs to be reviewed?

**"AppData\Local\Programs\Python\Python37\lib\runpy.py:125: RuntimeWarning: 'tf2onnx.convert' found in sys.modules after import of package 'tf2onnx', but prior to execution of 'tf2onnx.convert'; this may result in unpredictable behaviour"**

I tried Python 3.9 and downgraded to Python 3.7.9 just incase.

DeveloperNo579212:
I confirmed my conversion is valid and comparable with the source in saved model format, in python.
Using the section "Verifying a Converted Model" here https://onnxruntime.ai/docs/tutorials/tf-get-started.html

fatcat-z:
Given the conversion is valid so closing this. Please feel free to file a new one if any further questions.

PR #2416, opened 2025-10-31
Fix RuntimeWarning: 'tf2onnx.convert' found in sys.modules after import of package 'tf2onnx', but prior to execution of 'tf2onnx.convert'

Adraub:
When running `python -m tf2onnx.convert` to convert a model i got this warning :
> \<frozen runpy\>:128: RuntimeWarning: 'tf2onnx.convert' found in sys.modules after import of package 'tf2onnx', but prior to execution of 'tf2onnx.convert'; this may result in unpredictable behaviour
>
I looked for the root reason, and it seems most module are eagerly imported directly in the \_\_init\_\_.py while nothing seems to justify it.
It also happened to others (see https://github.com/onnx/tensorflow-onnx/issues/1831) and can be confusing when your conversion is going wrong.
I think these lines should be removed, i ran the tests and found no regressions.

### DIFF
--- a/tf2onnx/__init__.py
+++ b/tf2onnx/__init__.py
@@ -5,6 +5,25 @@
 __all__ = ["utils", "graph_matcher", "graph", "graph_builder",
            "tfonnx", "shape_inference", "schemas", "tf_utils", "tf_loader", "convert"]
 
+import importlib
 import onnx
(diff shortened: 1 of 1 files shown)


adraub:
I added lazy loading of the submodules, it fixes the initial warning and the UT error

PR #2437, opened 2026-02-18
[Feature] Add support for tf.math.xlogy

jaytiwarihub:
**Description:**
This PR adds support for the `tf.math.xlogy(x, y)` operation, which computes `x * log(y)` but returns `0` if `x == 0`. This is a common operation in loss functions (like cross-entropy) and probabilistic models.

**Implementation Details:**
- Mapped `Xlogy` to a combination of ONNX nodes: `Log`, `Mul`, and `Where`.
- Implemented safe handling for the `x=0` case:
  - Standard `x * log(y)` results in `NaN` if `x=0` and `y<0`.
  - Used `Where(x==0, 0, x * log(y))` to ensure the result is correctly `0` (matching TensorFlow behavior) and to avoid propagating NaNs.

**Changes:**
- Added `Xlogy` handler to `tf2onnx/onnx_opset/math.py`.

**Verification:**
- Verified locally with a custom script comparing `tf.math.xlogy` output against the converted ONNX model.
- Confirmed correct graph generation for standard inputs and the `x=0` edge case.


jaytiwarihub:
@xadupre I've really enjoyed diving into the graph rewriting and adding missing ops over the last few weeks. I was wondering: is tensorflow-onnx planning to participate in the upcoming LFX Summer Mentorship term? If so, I would love to officially apply and take on a larger feature.

PR #2468, opened 2026-07-01
cover newer opsets

bertsky:
completes the `OPSET_TO_IR_VERSION` mapping as per https://github.com/onnx/onnx/blob/main/docs/Versioning.md#released-versions

(not sure this is enough at runtime, though – depending on `onnxruntime` version, it may still complain during `ValidateOpsetForDomain()` for most recent versions)

### DIFF
--- a/tf2onnx/constants.py
+++ b/tf2onnx/constants.py
@@ -56,4 +56,5 @@ ENV_TF2ONNX_CATCH_ERRORS = "TF2ONNX_CATCH_ERRORS"
 # Refer from https://github.com/onnx/onnx/blob/main/docs/Versioning.md#released-versions
 OPSET_TO_IR_VERSION = {
     1: 3, 2: 3, 3: 3, 4: 3, 5: 3, 6: 3, 7: 4, 8: 4, 9: 4, 10: 5, 11: 6, 12: 7, 13: 7, 14: 7, 15: 8, 16: 8, 17: 8, 18: 8
+    19: 9, 20: 9, 21: 10, 22: 10, 23: 11, 24: 12, 25: 13, 26: 13,
 }


dongryeolle:
Thanks for tackling this @bertsky — the `OPSET_TO_IR_VERSION` map being stuck at opset 18 is a real blocker for newer opsets, so this is a useful fix to drive.

  A couple of things I ran into while working on the same area:                                           
                                                                                                          
  **1. Small bug: missing comma.** The new line lands right after `... 18: 8` without a trailing comma on the previous entry, so the dict doesn't parse:
                                                                                                          
  ```python                                          
      ... 17: 8, 18: 8      # <- no comma here                                                            
      19: 9, 20: 9, ...     # -> SyntaxError: invalid syntax                                              
  ```                                                
 
  `import tf2onnx` fails as-is. (It slipped through because the lint/test CI is `action_required` on this PR and never ran.)
                                                     
  **2. Consider deriving the map instead of hand-maintaining it.** Since this dict has to be extended by hand every release (and the entries for opsets 24-26 here are educated guesses about future IR versions),
  it might be more robust to derive it from the release table ONNX already ships:                         
                                                                                                          
  ```python                                                                                               
  from onnx import helper                                                                                 
                                                                                                          
  # opset -> ir_version, from the onnx release table so new releases are picked up                        
  # automatically; naturally capped by the installed onnx's known opsets.                                 
  OPSET_TO_IR_VERSION = {opset: ir for _, ir, opset, *_ in helper.VERSION_TABLE}                          
  # opsets 2-4 have no table row, and 7/8 need IR4 for PlaceholderWithDefault:                            
  OPSET_TO_IR_VERSION.update({1: 3, 2: 3, 3: 3, 4: 3, 5: 3, 6: 3, 7: 4, 8: 4})                            
  ```                                                                                                     
                                                                                                          
  This removes the manual step entirely, can't drift, and avoids hardcoding IR versions for opsets that aren't released yet.
                                                                                                          
  **3. On your `ValidateOpsetForDomain()` concern** — you're right that it's real, but it's an *onnxruntime* ceiling, not an onnx one. I tested empirically: onnxruntime 1.16.3 loads up to opset 19, 1.19.2 up to
  opset 21, and opset 22 needs ort >= 1.20 (1.19.2 rejects it as "under development"). So enabling the map is necessary but a runnable model also needs an ort new enough for the target opset — worth
  clamping/documenting rather than relying on the map alone.                                              
                                                                                                          
  I have a branch that builds on the same idea (the derivation above) and adds CI matrix jobs that actually exercise opset 19-22 and onnx 1.21/1.22 end-to-end, plus a fix for `onnx.mapping` being removed in
  onnx 1.19. Happy to fold your change in / open it as a follow-up / co-author — whatever you prefer. Just didn't want to open a competing PR on the same file

bertsky:
Congratulations, @dongryeolle – you nailed it!

You are absolutely right in every respect, and your PR is vastly superior and very thorough. So I'll close here. Don't worry about attribution. Thank you very much!

Issue #2472, opened 2026-07-09
Keep tf2onnx in sync with modern ONNX releases (opsets, packages, deps)


dongryeolle:
## Goal                                                                                                                                                                                                        
  Track the work needed for tf2onnx to keep up with new ONNX releases — newer                                                                                                                                    
  opsets, newer `onnx` packages, and the transitive dependency churn (protobuf,                                                                                                                                  
  numpy, onnxruntime) that each release drags in. This is an umbrella for several                                                                                                                                
  small, independently reviewable PRs rather than one large change.                                                                                                                                              
                                                                                                                                                                                                                 
  Subsumes the recurring "integrate the ONNX release branch" requests                                                                                                                                            
  (#2446 onnx 1.21, #2462 onnx 1.22; historically #2310, #2358).                                                                                                                                                 
                                                                                                                                                                                                                 
  ## Approach                                                                                                                                                                                                    
  Land the CI *diagnostic* first, then let its results prioritise the fixes:                                                                                                                                     
  the advisory (`experimental`) matrix jobs added in #2473 run the suite against                                                                                                                                 
  onnx 1.21/1.22 and opset 19-22, so we can see *which* incompatibilities actually                                                                                                                               
  bite before speculatively fixing them.                                                                                                                                                                         
                                                                                                                                                                                                                 
  ## Status                                                                                                                                                                                                      
  Legend: ✅ done · 🔜 in progress · ⬜ planned                                                                                                                                                                  
                                                                                                                                                                                                                 
  ### Enablement & diagnostics — #2473 ✅ merged (2026-07-14)                                                                                                                                                    
  - ✅ Derive `OPSET_TO_IR_VERSION` from `onnx.helper.VERSION_TABLE` (new onnx                                                                                                                                   
    releases auto-enable their opsets; no hand-edited map).                                                                                                                                                      
  - ✅ Advisory CI matrix jobs: opset 19-22 (onnx 1.17) + onnx 1.21.0 / 1.22.0                                                                                                                                   
    package integration, non-blocking so they can't break unrelated PRs.                                                                                                                                         
  - ✅ Replace `onnx.mapping.TENSOR_TYPE_TO_NP_TYPE` (removed in onnx 1.19) with                                                                                                                                 
    `helper.tensor_dtype_to_np_dtype`.                                                                                                                                                                           
  - Related: #2468 also targets the opset→IR map (manual); this epic's PR #2473 uses the derived approach.                                                                                                       
[...]

PR #2474, opened 2026-07-09
fix: replace numpy 2.0-removed aliases (np.product/np.cast/np.string_)

dongryeolle:
## What                                            
                                                     
  numpy 2.0 removed several long-deprecated aliases. This replaces the remaining                          
  usages left in the tree with their supported equivalents. Every change is
  behaviour-preserving — the replacements are exact aliases or an equivalent                              
  expression, so no conversion output changes.                                                            
                                                                                                          
  | Location | Before | After |                                                                           
  |---|---|---|                                      
  | `tf2onnx/graph.py` | `np.product(...)` | `np.prod(...)` |                                             
  | `tf2onnx/tfjs_utils.py` | `np.product(...)` | `np.prod(...)` |
  | `tests/run_pretrained_models.py` | `np.product(...)` | `np.prod(...)` |
  | `tools/onnx-experiments.py` | `np.cast[t](*inputs)` | `np.asarray(*inputs).astype(t)` |
  | `tools/onnx-experiments.py` | `"Cast": np.cast` | `"Cast": not None` |
  | `tests/test_backend.py` | `dtype=np.string_` | `dtype=np.bytes_` |
                                                     
  Details:                                           
                                                                                                          
  - **`np.product` → `np.prod`** — `np.product` was removed in numpy 2.0; `np.prod`                       
    is the canonical name and has the identical signature (the `dtype=` kwarg in                          
    `tfjs_utils.py` is preserved).                                                                        
  - **`np.string_` → `np.bytes_`** — `np.string_` was removed in numpy 2.0;
    `np.bytes_` is the `S`-kind dtype it aliased.                                                         
  - **`np.cast`** — removed in numpy 2.0. In `onnx-experiments.py`, `Cast` is
    special-cased inside the constant-folding loop, so `func_map["Cast"]` is only                         
    read as a non-`None` sentinel (never called). It now matches the neighbouring
    `"Transpose"/"Unsqueeze"/"Slice": not None` entries, and the real cast is done
    with `np.asarray(x).astype(t)`.                                                                       
                                                     
  This keeps the existing numpy-alias cleanup going (`onnx.mapping` and earlier                           
  `np.*` fixes already landed); a repo-wide scan for numpy-2.0-removed symbols is                         
  clean after this change.      
                                                                          
## Scope
 
  This PR is **source-code only** — it swaps numpy-2.0-removed aliases for their
  supported equivalents so the library doesn't crash under numpy 2.x. It does
  **not** touch dependency pinning or CI.                                                                 
                                                     
  The packaging/CI side of numpy-2 support — dropping the redundant direct                                
  `protobuf` dependency and making the unit-test `numpy<2` pin configurable — is                          
  handled separately in #2475 so the two concerns can be reviewed independently.
  There is no file overlap between the two. Both are Part of #2472.
                                                                                                    
  ## Testing                                         
                                                                                                          
  - Repo-wide grep for numpy 2.0-removed aliases returns no hits.
  - Edited `tf2onnx` modules import cleanly; the edited tool/test files compile.
  - The affected `tests/test_backend.py::test_equal_string` passes.

Part of #2472

### DIFF
--- a/tests/run_pretrained_models.py
+++ b/tests/run_pretrained_models.py
@@ -151,7 +151,7 @@ def get_sentences(shape):
     def get_sentence():
         length = random.randint(2, 7)
         return ' '.join(random.choice(words) for _ in range(length))
-    return np.array([get_sentence() for _ in range(np.product(shape))]).reshape(shape)
+    return np.array([get_sentence() for _ in range(np.prod(shape))]).reshape(shape)
 
 
 _INPUT_FUNC_MAPPING = {
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1428,8 +1428,8 @@ class BackendTests(Tf2OnnxBackendTestBase):
         self._run_test_case(func, [_OUTPUT], {_INPUT: x_val1, _INPUT1: x_val2})
 
     def test_equal_string(self):
-        x_val1 = np.array(['1'], dtype=np.string_)
-        x_val2 = np.array(['2'], dtype=np.string_)
+        x_val1 = np.array(['1'], dtype=np.bytes_)
+        x_val2 = np.array(['2'], dtype=np.bytes_)
         def func(x1, x2):
             mi = tf.equal(x1, x2)
             return tf.identity(mi, name=_TFOUTPUT)
--- a/tf2onnx/graph.py
+++ b/tf2onnx/graph.py
@@ -102,7 +102,7 @@ class Node(object):
             return external_tensor_storage.node_to_modified_value_attr[self]
         if external_tensor_storage is None or a.type != AttributeProto.TENSOR:
             return a
-        if np.product(a.t.dims) > external_tensor_storage.external_tensor_size_threshold:
+        if np.prod(a.t.dims) > external_tensor_storage.external_tensor_size_threshold:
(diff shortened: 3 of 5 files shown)


andife:
@dongryeolle does the package still work with old numpy "<2" ?

dongryeolle:
Hi @andife 

Yes — fully backward-compatible with numpy "<2". All three replacements use the                         
  canonical spellings that exist in both numpy 1.x and 2.x; only the removed                              
  *aliases* were the numpy-2.0 casualties:                                                                
  - np.product -> np.prod   (np.prod has always existed; np.product was the alias)                        
  - np.string_ -> np.bytes_ (np.bytes_ is the canonical scalar type in 1.x too)                           
  - np.cast[t](x) -> np.asarray(x).astype(t)  (standard API, version-independent)                         
  Verified on numpy 1.24.3: np.prod([2,3,4])==24, np.bytes_ resolves, and
  np.asarray([1.5,2.5]).astype(np.int64) works; and test_equal_string (which                              
  exercises the np.bytes_ path) passes under numpy 1.x. So the PR makes the code
  run on numpy 2.x without dropping numpy <2 support.

andife (approved):

PR #2476, opened 2026-07-15
fix: handle protobuf runtime_version import across versions

### LINKED ISSUE #2418: ImportError: cannot import name 'runtime_version' from 'google.protobuf'
Running the following command fails:

```bash
python -m venv venv
venv/bin/pip install git+https://github.com/onnx/tensorflow-onnx
venv/bin/python -m tf2onnx.convert --tflite x.tflite --output x.onnx
```

Produces the following error:
```
  File "<frozen runpy>", line 189, in _run_module_as_main
  File "<frozen runpy>", line 112, in _get_module_details
  File "/home/abique/develop/bitwig/fixes1/float/transient-detector/build/venv/lib/python3.13/site-packages/tf2onnx/__init__.py", line 8, in <module>
    import onnx
  File "/home/abique/develop/bitwig/fixes1/float/transient-detector/build/venv/lib/python3.13/site-packages/onnx/__init__.py", line 78, in <module>
    from onnx.external_data_helper import (
    ...<3 lines>...
    )
  File "/home/abique/develop/bitwig/fixes1/float/transient-detector/build/venv/lib/python3.13/site-packages/onnx/external_data_helper.py", line 14, in <module>
    from onnx.onnx_pb import (
    ...<5 lines>...
    )
  File "/home/abique/develop/bitwig/fixes1/float/transient-detector/build/venv/lib/python3.13/site-packages/onnx/onnx_pb.py", line 4, in <module>
    from .onnx_ml_pb2 import *  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/abique/develop/bitwig/fixes1/float/transient-detector/build/venv/lib/python3.13/site-packages/onnx/onnx_ml_pb2.py", line 9, in <module>
    from google.protobuf import runtime_version as _runtime_version
ImportError: cannot import name 'runtime_version' from 'google.protobuf' (/home/abique/develop/bitwig/fixes1/float/transient-detector/build/venv/lib/python3.13/site-packages/google/protobuf/__init__.py)
```

pctablet505:
Fixes #2418.

Installing from source and running `python -m tf2onnx.convert --tflite x.tflite --output x.onnx` can fail at import time with `ImportError: cannot import name 'runtime_version' from 'google.protobuf'`. The traceback points through `tf2onnx/__init__.py` → `onnx` → `onnx_ml_pb2.py`, which tries to import `google.protobuf.runtime_version`. That module was only introduced in protobuf 5.27, so ONNX 1.16+ generated files fail to load on older protobuf runtimes before `tf2onnx.convert` even starts.

A contributor checked this against current `main` (`6537fcb2`) in a clean Python 3.13 environment and could no longer reproduce the failure, because the current resolver now pulls in ONNX 1.22.0 and protobuf 7.35.1, both of which expose `runtime_version`. That change is partly driven by #2419, which bumped protobuf to patch CVEs. Still, the mismatch remains for anyone pinned to an older protobuf, and failing at import time is a bad experience regardless of what the latest resolver happens to choose.

Rather than force a protobuf upgrade, this adds a small compatibility shim. `tf2onnx/protobuf_compat.py` checks whether `google.protobuf.runtime_version` exists; if it does not, it injects a minimal stub module containing the constants ONNX's generated code expects (`MAJOR`, `MINOR`, `PATCH`, `SUFFIX`, `OSS_VERSION`) and a no-op `ValidateProtobufRuntimeVersion`. `tf2onnx/__init__.py` imports the shim before `import onnx`, so the stub is in place early enough. When the real module is already present, the shim is a no-op.

I added `tests/test_protobuf_compat.py` to cover both the stub path and the passthrough path. Running those tests together with the convert and const-fold suites passes; the unrelated `test_api.py` failures are the pre-existing Keras 3 / TF 2.16 issues.

### DIFF
--- a/tests/test_protobuf_compat.py
+++ b/tests/test_protobuf_compat.py
@@ -1,0 +1,50 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for protobuf runtime_version compatibility shim."""
+
+import sys
+import types
+
+import pytest
+
+
+def test_runtime_version_shim_provides_validator():
+    """When google.protobuf.runtime_version is missing, the shim injects one."""
+    # Simulate an older protobuf runtime by removing runtime_version from sys.modules.
+    real_module = sys.modules.pop("google.protobuf.runtime_version", None)
+    protobuf_pkg = sys.modules.get("google.protobuf")
+    original_has_runtime_version = hasattr(protobuf_pkg, "runtime_version") if protobuf_pkg else False
+    if protobuf_pkg and original_has_runtime_version:
+        delattr(protobuf_pkg, "runtime_version")
+
+    try:
+        # Clear any cached tf2onnx.protobuf_compat module so the shim path runs again.
+        sys.modules.pop("tf2onnx.protobuf_compat", None)
+        import tf2onnx.protobuf_compat  # noqa: F401  # side-effect under test
+
+        shim = sys.modules["google.protobuf.runtime_version"]
+        assert isinstance(shim, types.ModuleType)
+        assert hasattr(shim, "ValidateProtobufRuntimeVersion")
+        # The stubbed validator must be callable and not raise.
+        shim.ValidateProtobufRuntimeVersion(1, 2, 3, "", "dummy")
+    finally:
+        # Restore original state.
+        if real_module is not None:
+            sys.modules["google.protobuf.runtime_version"] = real_module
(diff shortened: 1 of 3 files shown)


pctablet505:
Closing this draft — the issue thread indicates the ImportError no longer reproduces on current main with updated dependencies, so there is no remaining failing case to fix.

PR #2480, opened 2026-07-30
fix: avoid IndexError for TFLite ops without outputs

### LINKED ISSUE #2380: bug: tflite conversion: IndexError: list index out of range on output_names[0]
**Describe the bug**

When converting a tflite model to onnx format, I get the following error:

```
$ python -m tf2onnx.convert --opset 200 --tflite okay_nabu.tflite --output okay_nabu.onnx


<frozen runpy>:128: RuntimeWarning: 'tf2onnx.convert' found in sys.modules after import of package 'tf2onnx', but prior to execution of 'tf2onnx.convert'; this may result in unpredictable behaviour
2025-01-12 01:55:41,581 - WARNING - ***IMPORTANT*** Installed protobuf is not cpp accelerated. Conversion will be extremely slow. See https://github.com/onnx/tensorflow-onnx/issues/1557
2025-01-12 01:55:41,582 - INFO - Using tensorflow=2.18.0, onnx=1.17.0, tf2onnx=1.16.1/e1042c
2025-01-12 01:55:41,582 - INFO - Using opset <onnx, 200>
2025-01-12 01:55:41,582 - WARNING - Currently installed onnx package 1.17.0 is too low to support opset 200, please upgrade onnx package to avoid potential conversion issue.
INFO: Created TensorFlow Lite XNNPACK delegate for CPU.
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "./venv/lib/python3.12/site-packages/tf2onnx/convert.py", line 714, in <module>
    main()
  File "./venv/lib/python3.12/site-packages/tf2onnx/convert.py", line 273, in main
    model_proto, _ = _convert_common(
                     ^^^^^^^^^^^^^^^^
  File "./venv/lib/python3.12/site-packages/tf2onnx/convert.py", line 168, in _convert_common
    g = process_tf_graph(tf_graph, const_node_values=const_node_values,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./venv/lib/python3.12/site-packages/tf2onnx/tfonnx.py", line 453, in process_tf_graph
    main_g, subgraphs = graphs_from_tflite(tflite_path, input_names, output_names)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./venv/lib/python3.12/site-packages/tf2onnx/tflite_utils.py", line 153, in graphs_from_tflite
    parse_tflite_graph(tfl_graph, opcodes, model, prefix, tensor_shapes_from_interpreter)
  File "./venv/lib/python3.12/site-packages/tf2onnx/tflite_utils.py", line 485, in parse_tflite_graph
    onnx_node = utils.make_onnx_node_with_attr(optype, input_names, output_names, name=output_names[0], **attr)
                                                                                       ~~~~~~~~~~~~^^^
IndexError: list index out of range
```

**System information**
- OS Platform and Distribution (e.g., Linux Ubuntu 18.04*): macOS arm64
- TensorFlow Version: 2.18
- Python version: 3.12.8
- ONNX version (if applicable, e.g. 1.11*):
- ONNXRuntime version (if applicable, e.g. 1.11*): 1.20.1

**To Reproduce**

```bash
curl -L -o okay_nabu.tflite https://github.com/esphome/micro-wake-word-models/raw/refs/heads/main/models/okay_nabu.tflite
python -m tf2onnx.convert --opset 200 --tflite okay_nabu.tflite --output okay_nabu.onnx
```

davidpavlovschi:
Fixes #2380

## Problem

Converting a TFLite model that contains an op with zero outputs (`ASSIGN_VARIABLE`, `CALL_ONCE`, ...) crashes:

```
File "tf2onnx/tflite_utils.py", line 532, in parse_tflite_graph
    onnx_node = utils.make_onnx_node_with_attr(optype, input_names, output_names, name=output_names[0], **attr)
IndexError: list index out of range
```

## Cause

`parse_tflite_graph` had three consecutive lines on `main`:

```python
onnx_node = utils.make_onnx_node_with_attr(optype, input_names, output_names, name=output_names[0], **attr)
node_name = output_names[0] if output_names else utils.make_name(f"{optype}_Output")
onnx_node = utils.make_onnx_node_with_attr(optype, input_names, output_names, name=node_name, **attr)
```

The guarded `node_name` fallback and the node creation that uses it are already there, but they are unreachable: the first, unguarded call runs before them and raises for any op with an empty `output_names`. Its result is also immediately overwritten, so the line is dead apart from the crash.

## Fix

Remove the stale unguarded call so the existing guarded path runs. `utils.make_name` appends a global counter, so the generated fallback names are unique; this is the only `output_names[0]` use left in the file.

## Verification

Reproduced with the model from the issue (macOS arm64, Python 3.12, TF 2.21, onnx 1.22):

```bash
curl -L -o okay_nabu.tflite https://github.com/esphome/micro-wake-word-models/raw/refs/heads/main/models/okay_nabu.tflite
python -m tf2onnx.convert --tflite okay_nabu.tflite --opset 18 --output okay_nabu.onnx
```

- Before: `IndexError: list index out of range` at `tflite_utils.py:532` (the model has 22 `ASSIGN_VARIABLE` ops and one `CALL_ONCE`, all with zero outputs).
- After: conversion finishes and writes a 220-node ONNX model; every node has a name. `tf2onnx` still logs `Unsupported ops: TFL_VAR_HANDLE, TFL_READ_VARIABLE` for that model — that is a separate missing-op gap and out of scope here; this change only removes the crash.

## Test

Added `tests/test_tflite_utils.py::TFListUtilsTests::test_parse_tflite_graph_op_without_outputs`: it builds a small `tf.Module` that assigns to a `tf.Variable`, converts it with `experimental_enable_resource_variables`, which produces `ASSIGN_VARIABLE`/`CALL_ONCE` ops with no outputs, and asserts `parse_tflite_graph` succeeds with non-empty, unique node names. It fails with the reported `IndexError` without the fix. CPU-only, no downloaded model.

`pytest tests/test_tflite_utils.py tests/test_tflite_postprocess.py` → 9 passed. `ruff check tf2onnx tests/*.py tools` → clean.

### DIFF
--- a/tests/test_tflite_utils.py
+++ b/tests/test_tflite_utils.py
@@ -72,3 +72,44 @@ class TFListUtilsTests(Tf2OnnxBackendTestBase):
             self.assertEqual(dtype, dtypes[name])
 
         self.assertTrue(len(onnx_nodes) >= 4)
+
+    @check_tf_min_version("2.0")
+    def test_parse_tflite_graph_op_without_outputs(self):
+        # Ops such as ASSIGN_VARIABLE have no outputs at all, so their node name
+        # cannot be derived from the first output tensor.
+
+        class VariableModule(tf.Module):
+            def __init__(self):
+                super().__init__()
+                self.v = tf.Variable([0.0, 0.0], dtype=tf.float32)
+
+            @tf.function(input_signature=[tf.TensorSpec([2], tf.float32, name="input")])
+            def __call__(self, x):
+                self.v.assign(x)
+                return tf.identity(self.v + 1.0, name="output")
+
+        module = VariableModule()
+        concrete_func = module.__call__.get_concrete_function()
+        converter = tf.lite.TFLiteConverter.from_concrete_functions([concrete_func], module)
+        converter.experimental_enable_resource_variables = True
+        tflite_model = converter.convert()
+
+        tflite_path = os.path.join(self.test_data_directory, self._testMethodName + ".tflite")
+        os.makedirs(os.path.dirname(tflite_path), exist_ok=True)
+        with open(tflite_path, 'wb') as f:
+            f.write(tflite_model)
+
+        tflite_graphs, opcodes_map, model, tensor_shapes = read_tflite_model(tflite_path)
+        nodes_without_outputs = []
+        for tflite_graph in tflite_graphs:
+            onnx_nodes = parse_tflite_graph(tflite_graph, opcodes_map, model,
+                                            tensor_shapes_override=tensor_shapes)[0]
+            names = [node.name for node in onnx_nodes]
+            self.assertEqual(len(names), len(set(names)))
+            for node in onnx_nodes:
+                self.assertTrue(node.name)
+                if len(node.output) == 0:
+                    nodes_without_outputs.append(node)
+
+        # The model must exercise the op-without-outputs path for this test to mean anything.
+        self.assertTrue(len(nodes_without_outputs) > 0)
--- a/tf2onnx/tflite_utils.py
+++ b/tf2onnx/tflite_utils.py
@@ -529,7 +529,6 @@ def parse_tflite_graph(tflite_g, opcodes_map, model, input_prefix='', tensor_sha
                         output_shapes[out] = None
         if has_prequantized_output:
             output_names = [get_prequant(out) for out in output_names]
-        onnx_node = utils.make_onnx_node_with_attr(optype, input_names, output_names, name=output_names[0], **attr)
         node_name = output_names[0] if output_names else utils.make_name(f"{optype}_Output")
         onnx_node = utils.make_onnx_node_with_attr(optype, input_names, output_names, name=node_name, **attr)
         onnx_nodes.append(onnx_node)


ryol8888:
Thanks for tracking this one down — the fix is right, and the history backs it up. #2278 added the guarded `node_name` fallback plus a second `make_node` call but left the original unguarded call in place, and #2279 then rewrote both to `make_onnx_node_with_attr`. So the line you delete has been dead-but-crashing ever since: its result is overwritten on the very next line.

Verified locally (python 3.11 / tf 2.15 / onnx 1.17):

- On `main`, a resource-variable model raises `IndexError` at `tflite_utils.py:532`; with this branch `parse_tflite_graph` completes and the zero-output ops come through as `TFL_ASSIGN_VARIABLE` / `TFL_CALL_ONCE`.
- No regression risk for models that convert today: `make_onnx_node_with_attr` is pure (it does not mutate `kwargs` or touch global counters), and the `utils.make_name` fallback only runs on the empty-outputs path that used to crash. Converting one fixed `.tflite` with `main` and with this branch gives structurally identical graphs — same op types, node names, inputs/outputs, attributes and initializers. (Raw bytes differ, but they also differ between two runs of `main`: only the `const_fold_opt__NN` counters move. Pre-existing, unrelated to this PR.)
- `tests/test_tflite_utils.py` passes, and the 6 red CI lanes are the pre-existing experimental ones that #2485 is fixing, not something this PR introduced.

Two comments.

**1. The new test reads as if resource-variable models now convert.** Asserting that `parse_tflite_graph` succeeds with unique, non-empty names is the right regression guard for the crash, but end-to-end those models are still not convertible, and nothing in the test says so. Could you make the scope explicit — a name like `test_parse_tflite_graph_op_without_outputs_does_not_crash`, or an assertion that pins the current limitation? Otherwise a later reader can easily take it as coverage for `ASSIGN_VARIABLE` / `CALL_ONCE` support.

Minor, related: the test depends on the TFLite converter still emitting those ops, so if a future TF stops doing that, `assertTrue(len(nodes_without_outputs) > 0)` fails for an unrelated reason. If that ever happens, a `check_tf_max_version` bound would keep the intent honest rather than the test being weakened.

**2. `Fixes #2380` overstates what lands here.** With this branch the model from that issue converts to a 220-node ONNX file, but 11 `TFL_VAR_HANDLE` + 11 `TFL_READ_VARIABLE` nodes stay unmapped, `onnx.checker` rejects the result (`No Op registered for TFL_VAR_HANDLE`), onnxruntime refuses to load it (`INVALID_GRAPH`), and the 22 `ASSIGN_VARIABLE` + 1 `CALL_ONCE` nodes are pruned because nothing consumes them. From the reporter's point of view the crash became a silently unusable file, so I would suggest `Part of #2380` and leaving the issue open.

Worth recording for whoever picks up that follow-up: these ops cannot be covered by adding handlers. TFLite keeps them precisely because the model writes to variables at inference time — with resource variables disabled the converter refuses outright (`Variable constant folding is failed. Please consider enabling experimental_enable_resource_variables`) — and the model really is stateful: feeding the same input five times gives `40, 40, 47, 54, 63`. ONNX has no persistent state between runs, so real support means lifting the 11 state buffers (7.8 KB total) into graph inputs/outputs, which changes the model signature. Partial support would be worse than none: mapping only `VAR_HANDLE` / `READ_VARIABLE` produces a model that reads state and never updates it — and unlike today's output, that one would pass the checker.

None of that is a request for this PR. The current end state (an error log plus an invalid model) is `tfonnx.py`'s global behaviour for unmapped ops, not something this change should touch.

andife (approved):

